PR #8, opened 2025-05-08
Fix Prefix-Store Hashing

vMaroon:


pkg/prefixstore/lru-store.go
```diff
@@ -106,22 +107,29 @@ func (c *LRUTokenStore) AddTokenization(modelName string, text string, tokens []
...
+	digest := xxhash.New()
+
 	// Chunk the text into blocks and populate the cache
-	for start := 0; start < len(text); start += c.blockSize {
+	for start := 0; start < len(promptBytes); start += c.blockSize {
 		end := start + c.blockSize
-		if end > len(text) {
-			end = len(text)
+		if end > len(promptBytes) {
+			end = len(promptBytes)
```
kfirtoledo: here you don't need a break?

PR #20, opened 2025-05-19
[build]: Add pre-commit target

kfirtoledo:
Add pre-commit target that includes copyright auto-generation.


.licenserc.yaml
```diff
@@ -0,0 +1,61 @@
+header:
+  license:
+    spdx-id: Apache-2.0
+    copyright-owner: Apache Software Foundation
+    software-name: llm-d
+    content: |
+      Copyright 2025 The llm-d Authors.
```
vMaroon, line 7: Is this a common practice? Can the year be automatic?

PR #38, opened 2025-06-06
docs: Add a setup documentation about examples/kv-cache-index

buraksekili:
This PR re-adds the setup docs related to examples/kv-cache-index/main.go, with minor updates:

- explains environment variables needed for examples/kv-cache-index/main.go
- parse those environment variables in the code,
- update context cancellation just in case


vMaroon:
The simulator does not simulate KV-cache events yet. But in all cases, the helm-chart covers the vLLM deployments with LMCache and the Redis instance.

Then the [kv-cache-index example](https://github.com/llm-d/llm-d-kv-cache-manager/blob/main/examples/kv-cache-index/main.go) can be used, but the model name must be aligned with that configured in the helm-chart. 
It would be more interesting to have a simple router that utilizes the scorer example instead, but that can follow up (should be tracked in an issue). Does this make sense to you?

docs/deployment/setup.md
```diff
@@ -0,0 +1,176 @@
+# KVCache Manager Setup Guide
```
vMaroon: s/KVCache/KV-Cache/

docs/deployment/setup.md
```diff
@@ -0,0 +1,176 @@
...
+Key configuration parameters:
+
+- `vllm.model.name`: The HuggingFace model to use (default: `meta-llama/Llama-3.1-8B-Instruct`)
+- `vllm.replicaCount`: Number of vLLM replicas (default: 4)
+- `vllm.poolLabelValue`: Label value for the inference pool (used by scheduler)
+- `redis.enabled`: Whether to deploy Redis for KV cache indexing (default: true)
+- `persistence.enabled`: Enable persistent storage for model cache (default: true)
+- `secret.create`: Create HuggingFace token secret (default: true)
+
+## Deploying the Inference Scheduler
```
vMaroon: I think this section should be emitted and addressed in another PR. This PR should focus on deploying the vLLM pods that feed the Redis index properly.

docs/deployment/setup.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Prerequisites
+
+- vLLM with LMCache and Redis deployed using the Helm chart above
+- llm-d-inference-scheduler repository
+
+<!-- TODO -->
+4. Configure the Gateway API resources to use the endpoint-picker. Refer to the [llm-d-inference-scheduler documentation](https://github.com/llm-d/llm-d-inference-scheduler) for detailed instructions.
+
+## Manual Configuration (Alternative to Helm)
```
vMaroon: This too should be emitted because this no longer works with the current code-base, due to Redis schema changes.

docs/deployment/setup.md
```diff
@@ -0,0 +1,176 @@
...
+
+Ensure you have a running deployment with vLLM and Redis as described above.
+
+### Running the Example
+
+The vLLM node can be tested with the prompt found in `examples/kv-cache-index/main.go`.
+
+```shell
+export HF_TOKEN=<token>
+export REDIS_HOST=<redis-host>  # optional, defaults to localhost:6379
```
vMaroon: This can include the password, see this recent work: https://github.com/llm-d/llm-d-inference-scheduler/issues/154 (this can be most useful: https://github.com/llm-d/llm-d-inference-scheduler/pull/131). The parsed address can look like: `redis://<user>:<pass>@localhost:6379/<db>`.

docs/deployment/setup.md
```diff
@@ -0,0 +1,176 @@
...
+```
+
+Environment variables:
+
+- `HF_TOKEN` (required): HuggingFace access token
+- `REDIS_HOST` (optional): Redis address; defaults to localhost:6379
+- `REDIS_PASSWORD` (optional): Redis password
+- `MODEL_NAME` (optional): The model name used in vLLM deployment; defaults to meta-llama/Llama-3.1-8B-Instruct
+
+### Model Name Alignment
```
vMaroon: I think this can be made simpler if you set the `MODEL_NAME` env variable to be used in both the helm-chart and here. Then this section can be condensed to a description around the env var's new location.

docs/deployment/setup.md
```diff
@@ -0,0 +1,176 @@
...
+
+```bash
+export HF_TOKEN=<your-huggingface-token>
+export NAMESPACE=<your-namespace>
+```
+
+2. Deploy using Helm:
+
+```bash
+helm upgrade --install vllm-p2p ./vllm-setup-helm \
```
vMaroon: s/vllm-p2p/llm-d-kv-cache-manager/ or any other name you like. At the time p2p referred to p2p sharing, which is not relevant here.

docs/deployment/setup.md
```diff
@@ -0,0 +1,176 @@
+# KVCache Manager Setup Guide
+
+This guide provides instructions for deploying the llm-d-kv-cache-manager with vLLM, LMCache, and Redis using Helm.
```
vMaroon: I think this opener should better introduce the scope and agenda of this doc. It should describe what is going to be ran in short, including the example and its content.

docs/deployment/setup.md
buraksekili: thanks for the context! updated according to your suggestion; `llm-d-kv-cache-manager`

docs/deployment/setup.md
```diff
@@ -0,0 +1,117 @@
+# KV-Cache Manager Setup Guide
+
+This guide provides a complete walkthrough for setting up and testing the example llm-d-kv-cache-manager system. You will deploy a vLLM with LMCache and Redis, then run an example application that demonstrates KV cache indexing capabilities.
+
+By following this guide, you will:
+
+1. **Deploy the Infrastructure**: Use Helm to set up a Kubernetes cluster running:
```
vMaroon: Technically the Kubernetes cluster is a pre-requisite and is not being set up here. This is well reflected below, perhaps the postfix sentence can just be dropped.

docs/deployment/setup.md
```diff
@@ -0,0 +1,117 @@
...
+   - vLLM servers (default: 4 replicas) serving Llama 3.1 8B Instruct model,
```
vMaroon: I think commas should be removed from list items.

docs/deployment/setup.md
```diff
@@ -0,0 +1,117 @@
...
+2. **Test with Example Application**: Run a Go application that:
+   - Connects to your deployed vLLM and Redis infrastructure,
+   - Demonstrates KV cache indexing by processing a sample prompt,
+   - Shows how to retrieve pod scores, 
+
+The end result is a working distributed LLM system with intelligent KV cache management that can route requests to pods with relevant cached computations.
+
+## vLLM Deployment
+
+The llm-d-kv-cache-manager repository includes a Helm chart for deploying vLLM with LMCache and Redis. This section describes how to use this Helm chart for a complete deployment.
```
vMaroon: s/vLLM with LMCache and Redis/vLLM with CPU offloading (LMCache) and KV-events indexing (Redis)/.

examples/kv-cache-index/main.go
```diff
@@ -37,40 +39,78 @@ In Redis:
...
+	if err != nil {
+		return err
 	}
 
 	logger.Info("created Indexer")
 
 	go kvCacheIndexer.Run(ctx)
-	logger.Info("started Indexer")
+	modelName := getModelName()
+	logger.Info("started Indexer", "model", modelName)
```
vMaroon: Not introduced in this PR, but `Info` messages should be capitalized.

docs/deployment/setup.md
```diff
@@ -0,0 +1,117 @@
...
+2. Deploy using Helm:
+
+```bash
+helm upgrade --install vllm-stack ./vllm-setup-helm \
+  --namespace $NAMESPACE \
+  --create-namespace \
+  --set secret.create=true \
+  --set secret.hfTokenValue=$HF_TOKEN \
+  --set vllm.model.name=$MODEL_NAME \
+  --set vllm.poolLabelValue="vllm-model-pool" \
```
vMaroon: Think it would be clearer if this is also an environment variable.

vMaroon (changes requested):
Thank you for the update, apologies for the delay in reviewing. Minor changes left then ready to go!

buraksekili:
Thank you @vMaroon for the help! I have updated the code accordingly. Please have a look

docs/deployment/setup.md
```diff
@@ -0,0 +1,118 @@
+# KV-Cache Manager Setup Guide
+
+This guide provides a complete walkthrough for setting up and testing the example llm-d-kv-cache-manager system. You will deploy a vLLM with LMCache and Redis, then run an example application that demonstrates KV cache indexing capabilities.
+
+By following this guide, you will:
+
+1. **Deploy the Infrastructure**: Use Helm to set up:
+   - vLLM servers (default: 4 replicas) serving Llama 3.1 8B Instruct model
```
vMaroon:
```suggestion
   - vLLM nodes with LMCache CPU offloading (4 replicas) serving Llama 3.1 8B Instruct model
```

Apologies for missing earlier, LMCache being a standalone item makes it sound like it will run separately. The LMCache engine is initialized within the vLLM processes.

docs/deployment/setup.md
```diff
@@ -0,0 +1,118 @@
...
+   - Redis server
+   - LMCache
```
vMaroon:
```suggestion
```

docs/deployment/setup.md
```diff
@@ -0,0 +1,118 @@
...
+2. **Test with Example Application**: Run a Go application that:
+   - Connects to your deployed vLLM and Redis infrastructure,
+   - Demonstrates KV cache indexing by processing a sample prompt
+   - Shows how to retrieve pod scores
```
vMaroon:
```suggestion
   - Demonstrates KV cache indexing by processing a sample prompt
```

The concept of scoring here can be confusing.

docs/deployment/setup.md
```diff
@@ -0,0 +1,118 @@
...
+
+The end result is a working distributed LLM system with intelligent KV cache management that can route requests to pods with relevant cached computations.
```
vMaroon:
```suggestion
The demonstrated KV-cache indexer is utilized for AI-aware routing to accelerate inference across the system through minimizing redundant computation.
```

This line would be true for a demo with a router. This setup guide only runs the indexer.

docs/deployment/setup.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Prerequisites
+
+Ensure you have a running deployment with vLLM and Redis as described above.
+
+### Running the Example
+
+The vLLM node can be tested with the prompt found in `examples/kv-cache-index/main.go`.
+
+First, ensure that the tokenizer engine is available locally for our example:
```
vMaroon:
```suggestion
First, download the tokenizer bindings required by the `kvcache.Indexer` for prompt tokenization:
```

docs/deployment/setup.md
```diff
@@ -0,0 +1,118 @@
...
+
+```bash
+make download-tokenizer
+```
+
+This will download the tokenizer engine.
```
vMaroon:
```suggestion
Then, set the required environment variables and run example:
```

vMaroon (changes requested):
Thanks, apologies for the long process, final set of suggestions then merging.

docs/deployment/setup.md
buraksekili: Thank you for pointing out this! 🙏 i have updated the code

Issue #36, opened 2025-05-26
vLLM GPU KV-Cache Based Indexing


vMaroon:
## Summary 

At the moment, the llm-d KV-cache awareness is limited to information indexed by components outside vLLM, to which KVs are offloaded through the Connector API. A part of the grander llm-d vision is to index the entire state of KVs across all memory-devices, inside or outside of vLLM.

This following work addresses the indexing gap inside vLLM and is currently in a state that we can use
- https://github.com/vllm-project/vllm/issues/16669

## Completion Criteria
- llm-d KV-events index to contain GPU events

PR #40, opened 2025-06-15
[KV-Events] Introduce KV-Block Indexing Backends - Part 1 of 3

vMaroon:
## Summary
This PR is the first in this work:
- #36 

This PR does not add or remove functionality, it merely restructures the code to allow for more sensible relations prepare for introducing an in-memory event-based index.

In short:
- `KVBlockIndexer` interface was split into an `Indexer` struct and a new `IndexBackend` interface + Redis implementation
- The tokenization prefix-cache implementation was moved under the `tokenization` package to avoid confusion


pkg/kv-cache/kv-block/indexer.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// IndexerConfig holds the configuration for the Indexer.
+type IndexerConfig struct {
+	*RedisIndexConfig
```
kfirtoledo, line 27: You will also need type: Redis or in-memory to know which NewXXXIndexBackend to create

pkg/kv-cache/kv-block/indexer.go
```diff
@@ -0,0 +1,67 @@
...
+// 1. A slice of strings representing the keys.
+// 2. A map where the keys are those in (1) and the values are pod names.
+// 3. An error if any occurred during the operation.
+//
+//nolint:gocritic // no need named return values here
+func (idxr *Indexer) GetPodsForKeys(ctx context.Context,
+	keys []Key, podIdentifierSet sets.Set[string],
+) ([]string, map[string][]string, error) {
+	return idxr.index.Lookup(ctx, keys, podIdentifierSet)
+}
```
kfirtoledo, line 67: You don't want to create one  to add and evict and not use the internal backend methods

kfirtoledo (changes requested):

pkg/kv-cache/kv-block/indexer.go
vMaroon: This will be in the follow-up.

pkg/kv-cache/kv-block/indexer.go
vMaroon: This will be in the follow-up. In this PR the scope is purely restructuring the existing code, without extending anything new/relevant for in-memory indexing. The latter is in-progress and will be pushed after.

kfirtoledo (approved):

PR #44, opened 2025-07-04
[KV-Events] KV-Events Processing - Part 3 of 3

vMaroon:
## Summary

This is the final PR in this series:
- #36

In this change, we refactor the codebase further (building on PRs #40 and #41) and introduce a new `kvevents` package to receive and process KVEvents over ZMQ.

The new package provides:

- A ZMQ subscriber with a pool of KV-Events processing workers, following vLLM’s formatting and prefix-cache bookkeeping pipeline. For now, all vLLM instances publish events to a single SUB socket; future work will add vLLM discovery and per-instance subscriptions.

Also included in this PR:

- **Token-processing updates for KV-Block key generation**  
  - Aligned with vLLM’s prefix-cache block hashing (`sha256_cbor`)
  - **Breaking change:** LMCache’s Redis indexing format is no longer compatible

- **KV-Events examples**  
  - Offline demo with a dummy publisher 
  - Online demo against a live vLLM deployment

- **Helm chart improvements**  
  - Enhanced configurability 
  - KVEvents support in vLLM
  - Deployment manifest for the online KV-Events demo - critical for syncing parameters across the vLLM/Indexer stack:  
    - vLLM KV-block hashing: `sha256_cbor`  
    - `PYTHONHASHSEED` environment variable  
    - vLLM prefix-cache block size

> This PR was tested against the [vMaroon/vllm](https://github.com/vMaroon/vllm) repository using the (new) online KV-Events example.

## Merge blockers (TODO)

- Update the documentation  
- Run stress tests and benchmarks  
- vLLM work:  
  - API standardization RFC 
  - Implement new hashing algorithms (`sha256_cbor` and `xxhash_cbor`), both using CBOR for serialization instead of pickle  
  - Fix KVEvents serialization bug for 256-bit hashes (currently only `builtin` algorithm works due to msgpack limits)

## Follow-up tasks

- Integrate LMCache via a new cache-controller component
- Add end-to-end and unit tests


vMaroon:
Online KV-Events example test run:

1. Request to `score` endpoint: (empty vLLM caches, empty indexer tables)
```
➜  vMaroon curl -X POST "http://localhost:8080/score" \        
  -H "Content-Type: application/json" \
  -d '{"prompt":"'"${TEXT}"'"}'                                  
null
```
2. Same request to vLLM:
```
➜  vMaroon curl -X POST "http://localhost:8000/v1/completions" \
  -H "Content-Type: application/json" \
  -d '{"prompt":"'"${TEXT}"'","max_tokens":50,"temperature":0.7}'
{"id":"cmpl-f0bbba3f2b49408eaac45b1ddb3650c0","object":"text_completion","created":1751584042,"model":"meta-llama/Llama-3.1-8B-Instruct","choices":[{"index":0,"text":" \nLorem ipsum ...","logprobs":null,"finish_reason":"length","stop_reason":null,"prompt_logprobs":null}],"usage":{"prompt_tokens":672,"total_tokens":722,"completion_tokens":50,"prompt_tokens_details":null},"kv_transfer_params":null}%  
```
3. Step (1) again after getting KV-Events
```
➜  vMaroon curl -X POST "http://localhost:8080/score" \
  -H "Content-Type: application/json" \
  -d '{"prompt":"'"${TEXT}"'"}'
{"10.128.5.220":9}
```

Dockerfile
```diff
@@ -44,15 +48,20 @@ RUN ranlib lib/*.a
 # the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
 # by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
 
-RUN CGO_ENABLED=1 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -ldflags="-extldflags '-L$(pwd)/lib'" -a -o bin/kv-cache-manager cmd/cmd.go
+RUN CGO_ENABLED=1 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH:-amd64} go build -ldflags="-extldflags '-L$(pwd)/lib'" -a -o bin/kv-cache-manager examples/kv-events/online/main.go
```
elevran: curiousity: why only amd64?

examples/kv-cache-index/main.go
```diff
@@ -84,22 +76,30 @@ func main() {
 	ctx := context.Background()
 	logger := klog.FromContext(ctx)
 
-	if err := kvCacheIndexer(ctx, logger); err != nil {
-		logger.Error(err, "failed to run kv-cache-indexer")
+	kvCacheIndexer, err := setupKVCacheIndexer(ctx)
+	if err != nil {
+		logger.Error(err, "failed to setup KVCacheIndexer")
```
elevran: nit: set up?

Suggest reference: setup vs set-up vs set up

examples/kv-events/README.md
```diff
@@ -0,0 +1,78 @@
...
+### Running the Example
+```
+    go run -ldflags="-extldflags '-L$(pwd)/lib'" examples/kv-events/offline/main.go examples/kv-events/offline/publisher.go
+```
+
+The example will start the KV-Cache Manager (indexer) and a dummy publisher that simulates KV-Events. 
+The demo will progress through:
+1. Initializing the KV-Cache Manager and the dummy publisher
+2. Querying the KV-Cache Manager for pod scores (initially empty)
+3. Simulateting KV-Events by the publisher for a dummy prmopt
```
elevran:
```suggestion
2. Querying the KV-Cache Manager for Pod scores (initially empty)
3. Simulating KV-Events by the publisher for a dummy prompt
```

examples/kv-events/offline/main.go
```diff
@@ -0,0 +1,282 @@
...
...m arcu. Aliquam consequat. Curabitur augue lorem, dapibus quis, laoreet et, pretium ac, nisi. Aenean magna nisl, mollis quis, molestie eu, feugiat in, orci. In hac habitasse platea dictumst. sunt in culpa qui officia deserunt mollit anim id est laborum. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Curabitur pretium tincidunt lacus. Nulla gravida orci a odio. Nullam varius, turpis et commodo pharetra, est eros bibendum elit, nec luctus magna felis sollicitudin mauris. Integer in mauris eu nibh euismod gravida. Duis ac tellus et risus vulputate vehicula. Donec lobortis risus a elit. Etiam tempor. Ut ullamcorper, ligula eu tempor congue, eros est euismod turpis, id tincidunt sapien risus a quam. Maecenas fermentum consequat mi. Donec fermentum. Pellentesque malesuada nulla a mi. Duis sapien sem, aliquet nec, commodo eget, consequat quis, neque. Aliquam faucibus, elit ut dictum aliquet, felis nisl adipiscing sapien, sed malesuada diam lacus eget erat. Cras mollis scelerisque nunc. Nullam arcu. Aliquam consequat. Curabitur augue lorem, dapibus quis, laoreet et, pretium ac, nisi. Aenean magna nisl, mollis quis, molestie eu, feugiat in, orci. In hac habitasse platea dictumst.`
```
elevran: this was already defined in another file?
consider using a separate file (not even `.go`) and using `go:embed` directives to add where needed

examples/kv-events/offline/main.go
```diff
@@ -0,0 +1,282 @@
...
+	blockHash1       = 7116574867160041607
+	blockHash2       = 9824230139929555794
+	blockHash3       = 3874749374659795295
+	defaultModelName = "bert-base-uncased"
+
+	envHFToken         = "HF_TOKEN"
+	envModelName       = "MODEL_NAME"
+	envZMQEndpoint     = "ZMQ_ENDPOINT"
+	envZMQTopic        = "ZMQ_TOPIC"
+	envPoolConcurrency = "POOL_CONCURRENCY"
```
elevran: which of these actually need to user configurable and which can be hardcoded?

examples/kv-events/offline/main.go
```diff
@@ -0,0 +1,282 @@
...
+
+	logger.Info("ZMQ publisher created successfully")
+	return publisher, nil
+}
+
+func runEventsDemo(ctx context.Context, kvCacheIndexer *kvcache.Indexer, publisher *Publisher) error {
+	logger := klog.FromContext(ctx)
+
+	modelName := getModelName()
+	logger.Info("@@@ Starting KV Events Demo", "model", modelName)
```
elevran: nit: why the `@@@` style in logging?

examples/kv-events/offline/publisher.go
```diff
@@ -0,0 +1,86 @@
...
+type Publisher struct {
+	socket   *zmq.Socket
+	endpoint string
+	seqNum   uint64
+}
+
+// NewPublisher creates a new ZMQ publisher.
+// endpoint is the ZMQ address to bind to (e.g., "tcp://*:5557").
+func NewPublisher(endpoint string) (*Publisher, error) {
+	socket, err := zmq.NewSocket(zmq.PUB)
```
elevran: from extensibility perspective, you may want to wrap the actual channel behind a wrapper interface.
Optional implementations; ZMQ, HTTP SSE, NATS, test mocks, ...
This could let you reuse more of the code in testing (eg `NewPublished(ch kv-events.Channel`)

examples/kv-cache-index/README.md
elevran, line 1: using dash (`-`) in go directory names is usually not a good idea (e.g., it is not a valid package path, IIRC).
It does not matter here as you're using it for examples which won't be imported, but consider replacing with all lower case name (kvcacheindex or kvindex) as a matter of being consistent

examples/kv-events/online/main.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/llm-d/llm-d-kv-cache-manager/pkg/kvcache/kvblock"
+	"github.com/llm-d/llm-d-kv-cache-manager/pkg/kvcache/kvevents"
+)
+
+const (
+	envHFToken     = "HF_TOKEN"
+	envModelName   = "MODEL_NAME"
+	envZMQEndpoint = "ZMQ_ENDPOINT"
+	envZMQTopic    = "ZMQ_TOPIC"
+
```
elevran: same question as before - which are really needed by a user? Esp. given this is a controlled sample...

pkg/kvcache/kvblock/token_processor.go
```diff
@@ -14,55 +14,44 @@ See the License for the specific language governing permissions and
...
 )
 
 const defaultChunkSize = 256
 
 // TokenProcessorConfig holds the configuration for the token processor.
 type TokenProcessorConfig struct {
 	ChunkSize int
-	*TemporaryTokenProcessorConfig
+	// HashSeed is used to prefix initial hash chunks, similarly to vLLM's NONE_HASH.
+	// This should be aligned with vLLM's `PYTHONHASHSEED` environment variable.
```
elevran, line 36: nit: consider indicating how to make it aligned?

pkg/kvcache/kvblock/token_processor.go
```diff
@@ -84,27 +73,60 @@ func NewChunkedTokenDatabase(config *TokenProcessorConfig) TokenProcessor {
...
-	// write each token to the buffer as binary data (using 64-bit big-endian format)
-	for _, token := range tokens {
-		// convert token to int64 for binary consistency
-		// LittleEndian is important to match the Python code
-		if err := binary.Write(buf, binary.LittleEndian, token); err != nil {
-			klog.FromContext(context.Background()).Error(err, "failed to write token to buffer")
-		}
+// hash computes a uint64 hash (lower 64 bits of SHA256).
+// The format, serialization and hashing is aligned with that of vLLM.
+func (db *ChunkedTokenDatabase) hash(parent uint64, tokens []uint32, extra interface{}) uint64 {
```
elevran, line 104: Q: is there really nothing that can be said about `extra`? 
Is it a byte array, does it implement Stringer, ....?
Seems to at least be cbor.Marshal enable?

pkg/kvcache/kvevents/events.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"fmt"
+
+	"github.com/vmihailenco/msgpack/v5"
+)
+
+// Event is a marker interface for KV cache events.
+type Event interface {
+	isEvent()
```
elevran, line 11: `Event` interface is package private (unexported `isEvent` method). 
Consider if you might be able to make the interface private to begin with? 
```go
type event interface{ ... }
```

pkg/kvcache/kvevents/pool.go
```diff
@@ -0,0 +1,260 @@
...
+	Concurrency int
+}
+
+// DefaultConfig returns a default configuration for the event processing pool.
+func DefaultConfig() *Config {
+	return &Config{
+		ZMQEndpoint: "tcp://indexer:5557",
+		TopicFilter: "kv@",
+		Concurrency: 4,
+	}
```
elevran: nit: consider making these constants?
They seem to be used in the examples as well, so you may want to import from here.

pkg/kvcache/kvevents/pool.go
```diff
@@ -0,0 +1,260 @@
...
+
+// Message represents a message that is read from a ZMQ topic.
+type Message struct {
+	Topic   string
+	Payload []byte
+	// Sequence number of the message
+	Seq uint64
+	// PodIdentifier is the identifier of the pod that sent the event.
+	// This will be extracted from the ZMQ topic.
+	PodIdentifier string
```
elevran, line 44: consider making `PodIdentifier` a type?

elevran (approved):
a few minor suggestions.

Dockerfile
vMaroon: This sets a default value, does not overwrite if set from outside (which is the case if going through the makefile). Chose `amd64` as default since that's the prominent architecture we're building for.

examples/kv-events/offline/main.go
vMaroon: Good point, minimizing to `HF_TOKEN`.

examples/kv-events/offline/main.go
vMaroon: Made it easier to track, especially when enabling debug/trace logging.

examples/kv-events/online/main.go
vMaroon: This one runs an actual indexer endpoint that is hooked with running vLLMs, so configuration here is useful and is used in the helm chart.

pkg/kvcache/kvblock/token_processor.go
vMaroon: Delegating this to follow-up work as originally intended. In general this field will contain information such as LoRA model-id and multi-modal data. Opening issue.

examples/kv-events/offline/publisher.go
vMaroon: Delegating to a followup. Opened issue.

pkg/kvcache/kvevents/pool.go
vMaroon: For now I prefer not to, it may make more sense when addressing the TODO on pod-identifier/model-name extraction per publisher.

pkg/kvcache/kvblock/index.go
```diff
@@ -97,12 +97,12 @@ type Index interface {
 // Key struct represents a unique identifier for a KV-cache block.
 type Key struct {
 	ModelName string // TODO: eject after aligning LMCache
-	ChunkHash string
+	ChunkHash int64
```
lorelei-sakai: I'm confused why this type change is described as "general refactoring".

From the examples, it seems the original hashes were 256 bits, but the new type is only 64 bits long. Is that a long enough hash for identifying cache blocks? Anyway, that seems like the type of change that should be called out separately and justified with at least some evidence.

vMaroon: This should be changed to bytes once that is done in vLLM's representation of hashes. Right now vLLM uses `int` for hashes and through `Msgpack`'s serialization it is limited to 64bits.

This was justified with the working example and reproduction of the hashes.

elevran (dismissed):

elevran (approved):

PR #48, opened 2025-07-09
Initial distributed tracing instrumentation

sallyom:
Initial distributed tracing instrumentation
Also, fix nil pointer dereference in KV Cache scorer initialization.


pkg/kvcache/indexer.go
```diff
@@ -113,34 +115,81 @@ func (k *Indexer) KVBlockIndex() kvblock.Index {
...
 	// 1. get available tokens of longest prefix
+	ctx, findTokensSpan := tracing.StartSpan(ctx, "kv-cache-manager.FindTokens", tracing.OperationFindTokens)
 	tokens := k.tokensIndexer.FindLongestContainedTokens(prompt, modelName)
 	if len(tokens) == 0 {
+		span.SetAttributes(
+			attribute.Int(tracing.AttrKVCacheTokenCount, 0),
+			attribute.String(tracing.AttrOperationOutcome, tracing.OutcomeSuccess),
+		)
+		tracing.SetSpanError(findTokensSpan, nil)
+		findTokensSpan.End()
```
frzifus: Does it make sense to wrap the `FindLongestContainedTokens` operation to set the attributes move it to the actual implementation? Missing the `End()` in the future can be easy.

pkg/kvcache/indexer.go
```diff
@@ -113,34 +115,81 @@ func (k *Indexer) KVBlockIndex() kvblock.Index {
...
+	// Calculate hit ratio for observability
+	totalPods := len(podIdentifiers)
+	if totalPods == 0 {
+		// If no specific pods requested, use all pods with scores
+		totalPods = len(podScores)
+	}
+
+	var hitRatio float64
+	if totalPods > 0 {
+		hitRatio = float64(len(podScores)) / float64(totalPods)
```
frzifus: What if `podScores == 0`?

pkg/kvcache/indexer.go
sallyom: updated to a more minimal solution - no longer relevant

pkg/kvcache/indexer.go
```diff
@@ -131,6 +135,20 @@ func (k *Indexer) KVBlockIndex() kvblock.Index {
 func (k *Indexer) GetPodScores(ctx context.Context, renderReq *preprocessing.ApplyChatTemplateRequest, prompt, modelName string,
 	podIdentifiers []string,
 ) (map[string]float64, error) {
+	// Start tracing span for main operation
```
vMaroon, line 136: I think all traces code should be encapsulated and contained, and enabled by configuration and not always. It would also be helpful to get some overhead data.

pkg/telemetry/tracing.go
```diff
@@ -0,0 +1,151 @@
...
+// Configuration is done via environment variables:
+// - OTEL_SERVICE_NAME: Service name (default: llm-d-kv-cache)
+// - OTEL_EXPORTER_OTLP_ENDPOINT: OTLP collector endpoint (default: http://localhost:4317)
+// - OTEL_TRACES_SAMPLER: Sampling strategy (default: parentbased_traceidratio)
+// - OTEL_TRACES_SAMPLER_ARG: Sampling ratio (default: 0.1 for 10%)
+func InitTracing(ctx context.Context) (func(context.Context) error, error) {
+	logger := klog.FromContext(ctx)
+
+	// Get service name from environment
+	serviceName := os.Getenv("OTEL_SERVICE_NAME")
```
vMaroon, line 70: Unless absolutely senseless, I think the configuration here should follow that of the rest of the components, by exposing a `Config` struct, and plugging it as part of the library's main config, or keeping it separate and leaving initialization responsibility to higher level users such as the scheduler.

pkg/kvcache/indexer.go
sallyom: We had a similar discussion in GAIE: https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2057#discussion_r2668935804 
The OTel spec provides Enabled() & IsRecording() API for conditional instrumentation when you have "computationally expensive operations when creating Spans" but that doesn't apply to these spans, we're passing already computed values as attributes [1].  I think it makes sense to keep the instrumentation unconditional because: 
* We want complete context propagation across `gaie -> epp -> kv-cache -> pd-proxy -> vllm` for a continuous trace. Making instrumentation conditional would break the context propagation. Trace initialization is conditional, though.
* When tracing is disabled, otel.Tracer() returns a no-op implementation [2]. The overhead is span allocation, which is negligible compared to the actual work (lookups, tokenization, scoring). It's not zero overhead, but guessing the no-op overhead would be <1% of a typical GetPodScores call. We should benchmark to confirm. 

 InitTracing() is conditional:
* When bundled in llm-d-inference-scheduler, the scheduler initializes the global provider, InitTracing isn't called from llm-d-kv-cache.
* When running standalone and tracing enabled: InitTracing() sets up the provider.
* Neither of above: no-op tracer with context propagation and negligible overhead
  
 [1] https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#enabled 
 [2] https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#behavior-of-the-api-in-the-absence-of-an-installed-sdk

pkg/telemetry/tracing.go
sallyom: For the library use case, when bundled with llm-d-inference-scheduler, the initialization is delegated to the scheduler. Scheduler controls all tracing configuration, the library has no initialization responsibility.

For standalone services/examples only, pkg/telemetry.InitTracing() uses standard OTel environment variables per the spec, we could add a config but I think it would duplicate Otel's standard mechanism:
  - https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/
  - https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/

pkg/kvcache/indexer.go
vMaroon: I see, that's ok. Do you think we can contain these operations within the different components? E.g., for metrics collection, we have an "InstrumentedIndex" which wraps the `kvblock.Index` implementations. Some of the calculations done here do appear there / in the `kvblock.Scorer` implementation.

sallyom: We could but the parent span in GetPodScores can't be pushed into a wrapper since it is the top-level method. We could keep the parent span in GetPodScores (it has to be there), but move the child spans into traced wrappers for Index.Lookup and KVBlockScorer.Score, similar to InstrumentedIndex. This would remove lookupWithSpan and scoreWithSpan from indexer.go. I'd rather do this in a follow-up since as/is, the tracing is more contained for this initial merge with all the spans inline.

pkg/kvcache/indexer.go
```diff
@@ -137,30 +159,138 @@ func (k *Indexer) GetPodScores(ctx context.Context, renderReq *types.RenderChatR
 	blockKeys := k.tokenProcessor.TokensToKVBlockKeys(kvblock.EmptyBlockHash, tokens, modelName)
 	if len(blockKeys) == 0 {
 		traceLogger.Info("no block keys found, returning empty scores")
+		span.SetAttributes(attribute.Int("llm_d.kv_cache.block_keys.count", 0))
 		//nolint:nilnil // no need to return an error
 		return nil, nil
 	}
 
+	span.SetAttributes(attribute.Int("llm_d.kv_cache.block_keys.count", len(blockKeys)))
```
liu-cong: nit: Move this up right after `TokensToKVBlockKeys` call and you can avoid the branching logic

pkg/kvcache/indexer.go
```diff
@@ -137,30 +159,138 @@ func (k *Indexer) GetPodScores(ctx context.Context, renderReq *types.RenderChatR
...
 		"pods", podsPerKeyPrintHelper(keyToPods))
 
-	// 5. score pods
-	podScores, err := k.kvBlockScorer.Score(blockKeys, keyToPods)
+	// Calculate total blocks available
+	totalBlocksAvailable := 0
+	for _, pods := range keyToPods {
+		totalBlocksAvailable += len(pods)
+	}
+	span.SetAttributes(attribute.Int("llm_d.kv_cache.total_blocks_available", totalBlocksAvailable))
```
liu-cong: Is there a doc listing all the attributes and their meaning? If not we should add one somewhere

pkg/kvcache/indexer.go
```diff
@@ -137,30 +159,138 @@ func (k *Indexer) GetPodScores(ctx context.Context, renderReq *types.RenderChatR
...
+		if score > 0 {
+			podsWithHits++
+		}
+	}
+	hitRatio := 0.0
+	if len(podIdentifiers) > 0 {
+		hitRatio = float64(podsWithHits) / float64(len(podIdentifiers))
+	}
+	span.SetAttributes(
+		attribute.Float64("llm_d.kv_cache.hit_ratio", hitRatio),
```
liu-cong: The definition of hit ratio here can be confusing, I would expect this to be the ratio of the hit blocks vs. total blocks. I don't think the ratio of hit pods vs. total pods is very useful.

pkg/kvcache/indexer.go
liu-cong: +1 to get tracing encapsulated. Having some sparse inline tracing code is fine, but the bulk should be encapsulated. I would argue for keeping the inline bits, and wrap the indexer and scorer with tracing either in this PR or in a follow up. It's much cleaner that way without needing another refacor

pkg/kvcache/indexer.go
sallyom: got it - I've updated this to `blocksFound / blockKeys` Also renamed the attributes to `block_hit_ratio` and `blocks_found` - Also updated the tracing proposal to match.

pkg/kvcache/indexer.go
sallyom: done

pkg/kvcache/indexer.go
sallyom: it's in the tracing proposal PR that will merge when this does. https://github.com/llm-d/llm-d/pull/119

pkg/kvcache/indexer.go
```diff
@@ -135,32 +157,131 @@ func (k *Indexer) GetPodScores(ctx context.Context, renderReq *types.RenderChatR
...
+		span.SetStatus(codes.Error, err.Error())
 		return nil, fmt.Errorf("failed to query kvblock scorer: %w", err)
 	}
 	traceLogger.Info("found pod scores", "pod-scores", podScores)
 
 	return podScores, nil
 }
 
+// lookupWithSpan wraps kvBlockIndex.Lookup with a tracing span.
+func (k *Indexer) lookupWithSpan(ctx context.Context, blockKeys []kvblock.BlockHash,
```
liu-cong: can we encapsulate the tracing logic into `kvBlockIndexc.Lookup`, or a new `kvBlockIndex.InstrumentedLookup` like @vMaroon suggested? Same comment for the `Score` method below.

pkg/kvcache/kvblock/traced_index.go
```diff
@@ -0,0 +1,88 @@
...
+	return t.next.Evict(ctx, engineKey, entries)
+}
+
+func (t *tracedIndex) Lookup(
+	ctx context.Context,
+	requestKeys []BlockHash,
+	podIdentifierSet sets.Set[string],
+) (map[BlockHash][]PodEntry, error) {
+	tracer := otel.Tracer(telemetry.InstrumentationName)
+	ctx, span := tracer.Start(ctx, "llm_d.kv_cache.storage.lookup",
```
liu-cong:
```suggestion
	ctx, span := tracer.Start(ctx, "llm_d.kv_cache.index",
```

pkg/kvcache/kvblock/traced_index.go
```diff
@@ -0,0 +1,88 @@
...
+		trace.WithSpanKind(trace.SpanKindInternal),
+	)
+	defer span.End()
+
+	span.SetAttributes(
+		attribute.Int("llm_d.kv_cache.lookup.block_count", len(requestKeys)),
```
liu-cong:
```suggestion
		attribute.Int("llm_d.kv_cache.index.lookup.block_count", len(requestKeys)),
```

liu-cong: I suggest adding `index` prefix to the attributs.

PR #52, opened 2025-07-14
Matched Default TokenProcessorConfig.BlockSize with vLLM's

vMaroon:
## Summary
- updated vLLM deployment image in helm-chart: should still be replaced to `llm-d:v0.2.0` once cut
- updated default config `TokenProcessorConfig.BlockSize` to 16 to match vLLM's default

### DIFF
--- a/pkg/kvcache/kvblock/token_processor.go
+++ b/pkg/kvcache/kvblock/token_processor.go
@@ -27,6 +27,8 @@ import (
 	"github.com/llm-d/llm-d-kv-cache-manager/pkg/utils"
 )
 
+// defaultBlockSize is the default number of tokens per block.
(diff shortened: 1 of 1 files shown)


pkg/kvcache/kvblock/token_processor.go
```diff
@@ -27,11 +27,11 @@ import (
 	"github.com/llm-d/llm-d-kv-cache-manager/pkg/utils"
 )
 
-const defaultChunkSize = 256
+const defaultBlockSize = 16
```
kfirtoledo, line 32: Can you add a comment on why you chose this value

dmitripikus (approved):
Changes are fine

Issue #54, opened 2025-07-16
kv cache aware pod scores are coming as null


pallavijaini0525:
Enabled the KV Cache aware scorer by setting ENABLE_KVCACHE_AWARE_SCORER to True and passing the HF_TOKEN , KVCACHE_INDEXER_REDIS_ADDR.

<img width="1426" height="333" alt="Image" src="https://github.com/user-attachments/assets/d1407efa-1754-4b8f-94db-e57f88e1c4d6" />

<img width="1475" height="330" alt="Image" src="https://github.com/user-attachments/assets/157e5251-a445-4a82-ad6b-d827e9d2da5b" />

after putting some debug statements realized that **E0714** memory allocation failures are coming from the https://github.com/llm-d/llm-d-kv-cache-manager/blob/v0.1.0/pkg/tokenization/tokinezer.go#L89. Tried by increasing the EPP container memory, no luck. 

please provide some suggestions. Thanks

vMaroon:
Thanks for sharing that, see PR above. We do suggest switching to v0.2 once it's out.

pallavijaini0525:
PR 63 is to update the kv-cache-manager to use the latest tokenizers to avoid the above issue.

PR #63, opened 2025-07-21
Update Tokenizer Release Version

vMaroon:
## Summary

Following the discussion in #54, the tokenizer version fixes an inconsistent bug that can occur on tokenizer load up.

## Fixes
- #54


Makefile
```diff
@@ -27,7 +27,7 @@ help: ## Print help
 LDFLAGS ?= -extldflags '-L$(shell pwd)/lib'
 CGO_ENABLED=1
 TOKENIZER_LIB = lib/libtokenizers.a
-TOKENIZER_RELEASE = v1.20.2
+TOKENIZER_RELEASE = v1.22.1
```
elevran: nit: consider if you want to grep for the same release in the Dockerfile as well to ensure alignment?

vMaroon: Done.

kfirtoledo (approved):

PR #64, opened 2025-07-22
Remove Default StorageClass:  "ocs-storagecluster-cephfs"

dumb0002:
This PR removes setting a default storage (i.e., `ocs-storagecluster-cephfs`) so that PVC for the vLLM pod can use a default storageClass available in any k8s cluster. Consequently, this will allow the deployment of the vLLM pod on environment independently of their default storageClass.


vMaroon (approved):
LGTM

dumb0002:
> @dumb0002 thanks for fixing - can you verify your signature?

@vMaroon done, thanks!

PR #75, opened 2025-07-31
Fix README Diagram

vMaroon:
## Summary

Minor README architecture diagram fixes - previous version was getting interpreted as a list that cannot be rendered on an arrow.


nilig (approved):
LGTM

PR #80, opened 2025-08-08
fix: Redis kvblock parsing bugs and add basic unit tests

yankay:
## Summary
This PR fixes 2 critical bugs in the Redis index implementation and adds unit tests for both Redis and in-memory kvblock index backends.

### 🐛 Bug Fixes
- **​​Fixed Redis pod identifier parsing​​**: Modified the string split delimiter from `:` to `@` to align with the expected PodEntry tostring format (podID@deviceTier). This change ensures accurate parsing when handling IPv6 addresses (which contain `:`) in PodEntry IP fields.
But it's a break change when upgrading.
- **Fixed Redis Lookup return value**: Corrected slice boundary calculation from `keys[:highestHitIdx]` to `keys[:highestHitIdx+1]` to include the last hit key

### ✅ Testing Improvements
- **Added basic unit tests** for both Redis and in-memory index implementations
- **Added miniredis dependency** for lightweight Redis testing without external dependencies
ref: #72

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/llm-d/llm-d-kv-cache-manager
 go 1.24.1
 
 require (
+	github.com/alicebob/miniredis/v2 v2.35.0
 	github.com/cespare/xxhash/v2 v2.3.0
 	github.com/daulet/tokenizers v1.22.1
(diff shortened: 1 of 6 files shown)


vMaroon (approved):
LGTM, thanks @yankay.

PR #92, opened 2025-08-20
Chat-Completions Enhancements: Updated Examples + Code Improvements

guygir:
## Summary

- Updated the KVEvents online example with chat completions functionality
- Enhanced implementation


vMaroon:
cc @elevran @nirrozenbaum for upcoming dependencies/build changes.

examples/kv_events/README.md
```diff
@@ -54,25 +52,68 @@ and the vLLM service is port-forwarded to `localhost:8000`, run:
...
 ```
+
+#### /v1/chat_completions
+
+1. Send a long prompt to the manager (expect no pod scores):
+```bash
+curl -X POST "http://localhost:8080/score_chat_completions" \
+  -H "Content-Type: application/json" \
+  -d '{
+    "model": "meta-llama/Llama-3.1-8B-Instruct",
```
nirrozenbaum: I suggest to keep consistency across the examples file:
```suggestion
    "model":"'"${MODEL}"'"
```

examples/kv_events/README.md
```diff
@@ -54,25 +52,68 @@ and the vLLM service is port-forwarded to `localhost:8000`, run:
...
+    ]
+  }'
+```
+
+1. Send an inference request to the vLLM endpoint (`v1/chat_completions`):
+```bash
+curl -X POST "http://localhost:8080/v1/chat/completions" \
+  -H "Content-Type: application/json" \
+  -d '{
+    "model": "meta-llama/Llama-3.1-8B-Instruct",
```
nirrozenbaum: ditto
```suggestion
    "model":"'"${MODEL}"'"
```

examples/kv_events/README.md
```diff
@@ -54,25 +52,68 @@ and the vLLM service is port-forwarded to `localhost:8000`, run:
...
+    ],
+  }'
+```
+
+1. Query the manager again with the same prompt:
+```bash
+curl -X POST "http://localhost:8080/score_chat_completions" \
+  -H "Content-Type: application/json" \
+  -d '{
+    "model": "meta-llama/Llama-3.1-8B-Instruct",
```
nirrozenbaum:
```suggestion
    "model":"'"${MODEL}"'"
```

examples/kv_events/online/main.go
```diff
@@ -52,6 +52,107 @@ const (
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	logger := klog.FromContext(ctx)
+
+	if err := run(ctx); err != nil {
+		logger.Error(err, "Failed to run unified KV-cache service")
+		return
+	}
+}
```
nirrozenbaum: This is duplicating the context and cancel func creation. also duplicated the error logging in error cases.
seems like those are not needed in main function but rather in run function:
 ```suggestion
	if err := run(); err != nil { // pay attention that ctx was removed 
		    os.Exit(1) // differentiate between error exit code and return in success path
	}
}
```

File: examples/kv_events/online/main.go [hunk 5]
```diff
@@ -52,6 +52,107 @@ const (
...
+
+func run(ctx context.Context) error {
+	logger := klog.FromContext(ctx)
+
+	// Create a cancellable context for the service
+	serviceCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
```
nirrozenbaum: and then just create the context here:
```suggestion
    ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	logger := klog.FromContext(ctx)
```

examples/kv_events/online/main.go
```diff
@@ -52,6 +52,107 @@ const (
...
+	logger.Info("Events pool started and listening for ZMQ messages")
+
+	// Setup graceful shutdown
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		<-sigChan
+		logger.Info("Received shutdown signal")
+		cancel()
+	}()
```
nirrozenbaum: it's probably better and more aligned with k8s to import controller-runtime and use here:
`ctrl.SetupSignalHandler()` instead of this code (encapsulates the SIGTERM & SIGINT signals).
the call `ctrl.SetupSignalHandler()` returns a context. if the only usage of `cancel()` is when a termination signal is received, it would be better to switch using that.

when reading the code again now, I'm updating to my previous comments - 
in main it would be better to do run(ctrl.SetupSignalHandler()), and then you don't need to create a context inside with cancellation (that context already listerns to cancellation).

nirrozenbaum: and if you already import controller-runtime, it is also more aligned with other parts to import `log` from there instead of `klog`, so using `log.FromContext` is my recommendation.

Dockerfile
```diff
@@ -21,9 +21,10 @@ WORKDIR /workspace
 
 USER root
 # Install EPEL repository directly and then ZeroMQ, as epel-release is not in default repos.
+# Install all necessary dependencies including Python 3.9 for chat-completions templating.
```
nirrozenbaum: I really think you should revisit the way you're handling dependencies.
This is not maintainable over time and not reproducible in a local environment (I mean for the deps that are installed in github workflows).

I really think you should switch to how we do it in GIE, where depedencies are downloaded to a local bin directory, and both CI, makefile and local testing are all referencing to that dir.

nirrozenbaum: atm on every dependency update you should change code in multiple places. that's not great.

examples/kv_events/online/main.go
```diff
@@ -158,37 +261,85 @@ func main() {
...
+
+	// Start HTTP server
+	httpPort := os.Getenv(envHTTPPort)
+	if httpPort == "" {
+		httpPort = defaultHTTPPort
+	}
+
+	go func() {
+		logger.Info("HTTP endpoint /score_completions exposed", "port", httpPort)
+		logger.Info("HTTP endpoint /score_chat_completions exposed", "port", httpPort)
```
nirrozenbaum: you have log lines for the fact that the endpoint is exposed multiple times. 
in general there are log duplication here.
suggest to remove the dups, it doesn't look good on the log and will make it harder for debug when needed.

Dockerfile
vMaroon: There is a difference between the two. The dependencies here are for linking against. It can be refactored if sensible when moving to the GIE / in llm-d later, but out of scope now.

nirrozenbaum: yes. it was a non blocking comment. just something to keep in mind.
might be worth opening an issue to track it

nilig (approved):
LGTM

nilig (approved):
LGTM

nilig (approved):
LGTM

PR #95, opened 2025-08-25
feat: Add Synchronous Tokenization Support to Tokenization Pool

### LINKED ISSUE #76: Synchronous Tokenization
In the current design, tokenization is never done synchronously. This was driven by wanting to avoid duplicated tokenization on the active serving path, but recently this is seeming like less of a problem.

Handling this issue would mean investigating a change in which tokenization is done synchronously as part of the active path, but still cached. The handling should ideally remain by the workers pool.

Profiling should be ideally provided.

sagearc:
**Overview**
This PR extends the tokenization pool with synchronous tokenization support. Callers can now block until a prompt has been tokenized and get the results back directly, instead of relying only on async/background workers.

**Key Changes**

Synchronous Tokenization API
- Added a `Tokenize()` method that blocks until tokenization is finished and returns tokens + offsets.
- Uses an internal result channel for communication.
- The async path still exists (`EnqueueTokenization`), so both options are available.

Indexer Integration
- `GetPodScores` now **always runs tokenization**, even if the prefix store hasn’t seen the prompt before.
- This makes lookups more reliable and ensures KV-cache scoring works for “cold” prompts too.

Test Performance Improvements
- Removed all `time.Sleep()` calls from E2E tests.
- **E2E Tests now run ~90% faster** (42s → 4s on my machine).
- No more race conditions: synchronous tokenization makes test results deterministic.

**Migration Notes**
- This isn’t 100% backwards compatible:
- The old AddTask() method has been renamed to EnqueueTokenization().
- Any code using AddTask will need a one-line update.
- Aside from that, existing async usage still works the same, and new code can opt into synchronous tokenization where needed.

**Related**
Closes #76, profiling work will be handled in a separate PR

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -42,7 +42,7 @@ sequenceDiagram
 
     Router->>Indexer: GetPodScores(prompt, model, pods[])
     
-    Note over Indexer: Asynchronously tokenizes prompt if not in cache
+    Note over Indexer: Synchronously tokenizes prompt using pool if not in prefix cache
     Indexer->>PrefixStore: FindLongestContainedTokens(prompt, model)
     PrefixStore-->>Indexer: cachedTokens[]
     
@@ -60,14 +60,13 @@ sequenceDiagram
 
 **Key Steps:**
 
-1.  **Token Retrieval**: The `Indexer` first checks the `PrefixStore` for the longest token sequence it has for the prompt's prefix. If the prompt isn't known, it's queued for background tokenization.
+1.  **Token Retrieval**: The `Indexer` first checks the `PrefixStore` for the longest token sequence it has for the prompt's prefix. If the prompt isn't cached or coverage is insufficient, it performs synchronous tokenization using the worker pool.
 2.  **Key Generation**: The retrieved tokens are sent to the `TokenProcessor`, which chunks and hashes them into a sequence of deterministic KV-block keys that match vLLM's logic.
 3.  **Index Lookup**: With the keys, the `Indexer` queries the `kvblock.Index` to see which pods have them. The lookup is optimized to find the longest *consecutive* chain of hits from the start.
 4.  **Scoring**: The `Scorer` takes the hit data and scores each pod based on its number of consecutive matching blocks.
 5.  **Response**: A final map of pod scores is sent back to the router.
 
-Note: step (1) means that the first time a prompt is scored, it may return an empty result while the tokenization happens in the background.
-It is assumed that this cache will be populated with common prompts, so the first scoring request is an edge case.
+Note: The tokenization pool now supports both asynchronous (fire-and-forget) and synchronous modes, ensuring scoring requests can always return complete results.
 
 ### Write Path: Processing Cache Events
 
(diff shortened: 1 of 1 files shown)


pkg/tokenization/pool.go
```diff
@@ -79,16 +87,29 @@ func NewTokenizationPool(config *Config, store prefixstore.Indexer) (*Pool, erro
...
+func (pool *Pool) EnqueueTokenization(prompt, modelName string) {
 	task := Task{
 		Prompt:    prompt,
 		ModelName: modelName,
 	}
 	pool.queue.Add(task)
 }
 
+// Tokenize queues a task and blocks until the final result is available.
+func (pool *Pool) Tokenize(prompt, modelName string) TokenizationResponse {
```
vMaroon, line 101: Since the offsets use is completely internal - I think this function can just return `[]uint32` (token IDs), and the `TokenizationResponse` type can be hidden.

vMaroon: Only the `tokenIDs` subfield of the returned value is ever used from these calls. I believe it would be clearer if they are returned directly. Then the `TokenizationResponse` type can be hidden and would not require exporting.

pkg/tokenization/pool_test.go
```diff
@@ -116,7 +116,7 @@ func TestPool_RunIntegration(t *testing.T) {
 	defer cancel()
 
 	for _, prompt := range prompts {
-		pool.AddTask(prompt, testModelName)
+		pool.EnqueueTokenization(prompt, testModelName)
```
vMaroon, line 143: Since this function is only used here - I think we can just remove it. Instead `Tokenize` can be tested.

sagearc: Keeping this test for now, as async tokenization will become
configurable soon.

pkg/kvcache/indexer.go
```diff
@@ -118,15 +118,9 @@ func (k *Indexer) GetPodScores(ctx context.Context, prompt, modelName string,
...
-	k.tokenizersPool.AddTask(prompt, modelName)
-
-	// 1. get available tokens of longest prefix
-	tokens := k.tokensIndexer.FindLongestContainedTokens(prompt, modelName)
-	if len(tokens) == 0 {
-		//nolint:nilnil // no need to return an error
-		return nil, nil
-	}
+
+	// 1. submit tokenization request
```
vMaroon: nit: "submit" might be now confusing.

pkg/tokenization/pool.go
```diff
@@ -43,10 +44,17 @@ func DefaultConfig() *Config {
 	}
 }
 
+// tokenizationResponse is the single (final) message sent back to a caller.
```
vMaroon: The structs description should be decoupled from its use.

pkg/tokenization/pool.go
```diff
@@ -79,16 +87,34 @@ func NewTokenizationPool(config *Config, store prefixstore.Indexer) (*Pool, erro
 	}, nil
 }
 
-// AddTask enqueues a new tokenization task.
+func (pool *Pool) GetTokenizer() Tokenizer {
```
vMaroon: Exported funcs should have documentation.

pkg/kvcache/indexer.go
```diff
@@ -118,15 +118,9 @@ func (k *Indexer) GetPodScores(ctx context.Context, prompt, modelName string,
 	podIdentifiers []string,
 ) (map[string]int, error) {
 	traceLogger := klog.FromContext(ctx).V(logging.TRACE).WithName("kvcache.GetPodScores")
-	// 0. add to tokenizers pool
-	k.tokenizersPool.AddTask(prompt, modelName)
-
-	// 1. get available tokens of longest prefix
-	tokens := k.tokensIndexer.FindLongestContainedTokens(prompt, modelName)
```
vMaroon, line 125: The lookup should remain - think it was accidentally ejected.

vMaroon (changes requested):
Very minor comments then good to go.

pkg/tokenization/pool.go
sagearc: Unused, deleted.

vMaroon (approved):
Great work @sagiahrac!

Commits are not verified, pls amend.

PR #102, opened 2025-09-03
[CI]: added some index-related test cases while refactoring the test code to be more concise.

yankay:
# Summary

Resolved a concurrency issue in the in-memory index, and added some index-related test cases while refactoring the test code to be more concise.

# Changes Made

Renamed `common_test.go` to `index_test.go` and added new test cases:  
- Duplicate pod handling behavior  
- Filtered lookup with pod identifier sets  
- Basic eviction functionality


yankay:
HI @vMaroon 
Would you please help to review it.

vMaroon:
The `github.com/hashicorp/golang-lru/v2` package is thread-safe, therefore an additional mutex was not added in the implementation.

I think I documented that somewhere but this PR shows that clarity is missing. I think this PR can pivot to a documentation enhancement on the matter, what do you think? @yankay

yankay:
> The `github.com/hashicorp/golang-lru/v2` package is thread-safe, therefore an additional mutex was not added in the implementation.
> 
> I think I documented that somewhere but this PR shows that clarity is missing. I think this PR can pivot to documentation/test enhancements, what do you think? @yankay

Thanks @vMaroon

I've modified this PR to only provide the testcase. Additionally, since I added a new `testConcurrentOperations`, I discovered a concurrency issue with `in_memory` and submitted another issue to describe it: #104

Issue #89, opened 2025-08-19
Improve InMemoryIndex to Use a Cost-Based Cache for Predictable Memory Usage


sagar0x0:
## Problem:
The current InMemoryIndex in [pkg/kvcache/kvblock/in_memory.go](https://github.com/llm-d/llm-d-kv-cache-manager/blob/main/pkg/kvcache/kvblock/in_memory.go) uses an item-count-based LRU cache. This leads to unpredictable memory usage because the size of cached values (the PodSet for each key) can vary significantly. A key associated with 100 pods(10 in our case) consumes far more memory than one with 2 pods, yet both are treated as a single item. This creates a risk of exceeding memory limits and causing system instability. This issue directly addresses the TODO comment in the code regarding memory-based configuration.

## Proposed Solution:
Replace the hashicorp/golang-lru cache with a high-performance, cost-based cache like [Ristretto.](https://github.com/hypermodeinc/ristretto)

This change will allow us to configure the cache with a fixed memory budget (e.g., 2 GB). Ristretto will manage this budget by evicting items based on their access frequency and their actual memory **cost**, which we will calculate based on the number of pods each key tracks. This also simplifies our architecture by replacing the two-level LRU with a single, more powerful cache.

## Key Benefits:

Predictable Memory Footprint: Guarantees the cache operates within a stable, predefined memory limit, preventing OOM errors.

Improved Performance & Concurrency: Ristretto is designed for high-throughput systems and uses techniques like lock sharding to minimize contention, resolving the existing TODO about batching operations.

Smarter Eviction Policy: Evicts items based on their memory cost and usage (LFU), leading to a more efficient use of cache memory and potentially a higher hit ratio.

PR #103, opened 2025-09-03
[feat] add cost-aware memory index implementation with Ristretto cache

yankay:
## Overview
This PR introduces a new cost-aware memory index implementation using the Ristretto cache library to provide efficient memory management for KV cache blocks.

TODO: Improve to be lock-free

## What's Changed
- **New CostAwareMemoryIndex**: Implements the Index interface with cost-aware memory management using Ristretto cache
- **Documentation Updates**: Updated configuration and architecture documentation to reflect the new index type
- **Dependencies**: Added Ristretto v2 dependency for high-performance caching

# Discussion points with reviewer:

Should we implement a new cost-aware index as like this PR, or improve the In-Memory index **in-place** by replacing `golang-lru` with `Ristretto`?

fix: #89
pending by: #102


pkg/kvcache/kvblock/cost_aware_memory.go
```diff
@@ -0,0 +1,252 @@
...
+	c.cache.Range(func(key, value interface{}) bool {
+		count++
+		return true
+	})
+	return count
+}
+
+// CalculateByteSize estimates memory usage for ristretto cost calculation.
+// This is an approximation used for cache eviction decisions.
+func (c *CostPodCache) CalculateByteSize(keyStr string) int64 {
```
vMaroon, line 111: Since the cost function can be anything (not necessarily accurate-size related) - I think this can be simplified to the number of pod entries. That would be an efficient reduction of the size (and arguably importance) of an entry.

What do you think?

pkg/kvcache/kvblock/cost_aware_memory.go
```diff
@@ -0,0 +1,252 @@
...
+
+const (
+	defaultNumCounters = 1e8                    // 100M keys
+	defaultIndexSize   = 2 * 1024 * 1024 * 1024 // 2 GB in bytes
+	defaultBufferItems = 64                     // default buffer size for ristretto
+)
+
+// CostAwareMemoryIndexConfig holds the configuration for the CostAwareMemoryIndex.
+type CostAwareMemoryIndexConfig struct {
+	Size int64 `json:"size"`
```
vMaroon: For clarity, this can benefit from a package like **https://github.com/dustin/go-humanize** to allow setting human readable sizes, example use (from their godocs):
```
ParseBytes("42 MB") -> 42000000, nil 
ParseBytes("42 mib") -> 44040192, nil
```

If you have other library preferences then also ok.

pkg/kvcache/kvblock/cost_aware_memory.go
yankay: HI @vMaroon 

If we modify CalculateByteSize to return the number of pod entries instead of the byte size, we would break Ristretto's memory management capability. ref:  https://github.com/llm-d/llm-d-kv-cache-manager/pull/103/files#diff-b508336e00a53c98f59abd0ac1bf6cebbac7dd3f0f3a96f799273f9251d64b5bR56

1. Cost Mismatch: The MaxCost is configured as 2GB (bytes), but each cache entry would only cost 1-10 (entry count)
2. Ineffective Eviction: Ristretto would think it can store 2G entries before hitting the limit
3. Lost Functionality: We'd lose the byte-based cache size limiting feature

How do you think about it?

pkg/kvcache/kvblock/cost_aware_memory.go
yankay: Thanks @vMaroon 


It has been changed, and a test case https://github.com/llm-d/llm-d-kv-cache-manager/pull/103/files#diff-cb164f5602d32204943a549b76165fce3a3992e7701c2f64c046852531346e21R84 has been added to it :-)

And another solution is to use the Kubernetes quantity. https://github.com/kubernetes/kubernetes/blob/v1.34.0/staging/src/k8s.io/apimachinery/pkg/api/resource/quantity.go#L277

But in that case, it can't handle "42 MB", only "42M". That's reasonable for k8s, but seems a bit too strict for configuration files.

pkg/kvcache/kvblock/cost_aware_memory.go
vMaroon: My bad, their example confused me - thought the two concepts are decoupled. Upon reading their introduction blog it was clearer.

Sounds good.

vMaroon (approved):
Looks great - thank you @yankay.

/lgtm

PR #107, opened 2025-09-04
Minimize high-concurrency `kvblock.Index` PodCache maintenance race-condition window

### LINKED ISSUE #104: Race condition in InMemoryIndex concurrent operations causing test failures
## Problem Description

The `InMemoryIndex` implementation has a race condition when performing concurrent operations (Add, Lookup, Evict) on the same key. This manifests as test failures in the `TestInMemoryIndexBehavior/ConcurrentOperations` test.

## Test Failure Details

ref to: #102

```

--- FAIL: TestInMemoryIndexBehavior (0.02s)
    --- FAIL: TestInMemoryIndexBehavior/ConcurrentOperations (0.02s)
        /root/oss/llm-d-kv-cache-manager/pkg/kvcache/kvblock/index_test.go:219: 
            	Error Trace:	/root/oss/llm-d-kv-cache-manager/pkg/kvcache/kvblock/index_test.go:219
            	            				/usr/local/go/src/runtime/asm_amd64.s:1700
            	Error:      	[]string{"pod-2-0", "pod-55-0", "pod-39-0", "pod-8-0", "pod-9-0", "pod-12-0", "pod-30-0"} does not contain "pod-3-0"
            	Test:       	TestInMemoryIndexBehavior/ConcurrentOperations
FAIL
FAIL	github.com/llm-d/llm-d-kv-cache-manager/pkg/kvcache/kvblock	0.038s
FAIL

```

## Root Cause Analysis

The issue stems from multiple race conditions in the `InMemoryIndex` implementation:

###  Race Condition in Evict Method
In the `Evict` method (lines 193-194 in `in_memory.go`), there's a critical race condition:

``` go
if podCache.cache.Len() == 0 {
    m.data.Remove(key)  // Race condition here!
}
```

The Problem: This check-then-act operation is not atomic. Between checking if the pod cache is empty and removing the key, another goroutine could add new pods to the cache .

vMaroon:
## Summary

In #104 @yankay pointed out that there are data race-conditions within the `kvblock.InMemoryIndex` implementation. While the index is thread-safe, the race-condition windows come at the benefit of fewer lock contentions for faster KVEvent processing, at the cost of risking temporary "instability" (misses) at irregular workloads. The race-condition windows are when multiple threads simultaneously create a `PodCache` for the same key (several pods admit the same new KVBlocks at the exact same time), or when a key is simultaneously evicted and admitted from two sources.

The changes in this PR tighten the open windows while not fully resolving them, as an acceptable tradeoff of likelihood vs. stable benefits.

## Related Issues
- Fixes #104

### DIFF
--- a/pkg/kvcache/kvblock/in_memory.go
+++ b/pkg/kvcache/kvblock/in_memory.go
@@ -172,7 +172,7 @@ func (m *InMemoryIndex) Add(ctx context.Context, keys []Key, entries []PodEntry)
 			contains, _ := m.data.ContainsOrAdd(key, newPodCache)
 			if contains {
 				podCache, found = m.data.Get(key)
-				if !found {
+				if !found { // Extremely irregular workload pattern - key evicted
 					m.data.Add(key, newPodCache)
 					podCache = newPodCache
 				}
--- a/pkg/kvcache/kvblock/in_memory.go
+++ b/pkg/kvcache/kvblock/in_memory.go
@@ -19,6 +19,7 @@ package kvblock
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	lru "github.com/hashicorp/golang-lru/v2"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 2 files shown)


nilig (approved):
LGTM

Issue #14, opened 2025-05-13
KVCache Indexer as a Service


vMaroon:
So far the `kvcache.Indexer` has been used as a Go module to implement kvcache-aware routing logic. Such use carries all the build dependencies such as linking tokenizer bindings and module dependencies.

The `kvcache.Indexer` is envisioned to evolve into a manager-service that provides more than API on indexing information. A step in that direction is the support of deploying the indexer as a gRPC service.

zhengkezhou1:
@vMaroon just want ask, can i pick up this issue?

vMaroon:
@liu-cong we'll see that once we have it. To be clear, it's an additional deployment manner, and not a replacement.

A benefit of this method can be duplicating the index separately for HA.

PR #109, opened 2025-09-08
feat: make kvcache.Indexer as gRPC Service

zhengkezhou1:
fix: #14

## Main Changes

- Added a gRPC API specification  to expose the core functionality of `kvcache.Indexer` as a service.
- Implemented a minimal server and client example under kv_cache_index_service to demonstrate basic usage and integration.


zhengkezhou1:
https://github.com/vMaroon please take a look.

zhengkezhou1:
@vMaroon please take a look.

vMaroon:
Thanks @zhengkezhou1 - will review as soon as possible.

examples/testdata/data.go
```diff
@@ -1,3 +1,17 @@
+// Copyright 2025 The llm-d Authors.
```
vMaroon, line 1: Can you extract this and others into a separate PR? To minimize changes in this one.

.dockerignore
```diff
@@ -1,3 +1,17 @@
+# Copyright 2025 The llm-d Authors.
```
vMaroon: No need for boilerplate here.

examples/testdata/data.go
zhengkezhou1: Ok, I ran `make precommit` before committing the changes.

.dockerignore
zhengkezhou1: maybe we need update Makefile.

examples/kv_cache_index_service/server/server.go
```diff
@@ -0,0 +1,106 @@
...
+			Score: int32(score),
+		})
+	}
+
+	return &indexerpb.GetPodScoresResponse{
+		Scores: scores,
+	}, nil
+}
+
+// TODO: Implement LookupBlocks when needed
```
vMaroon: This should be ejected as it is not supported by the indexer - same for the function that follows.

examples/kv_cache_index_service/server/main.go
```diff
@@ -0,0 +1,96 @@
...
+
+func setupIndexerService(ctx context.Context, config *kvcache.Config) (*IndexerService, error) {
+	indexer, err := setupKVCacheIndexer(ctx, config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create indexer: %w", err)
+	}
+
+	indexerSvc := NewIndexerService(indexer)
+
+	// Start the indexer
```
vMaroon, line 90: I think it would be useful to also setup the `kvevents.Pool` in this example - what do you think?

zhengkezhou1: Yes, i don't have deep understand, but maybe we should do the same thing here. https://github.com/llm-d/llm-d-kv-cache-manager/blob/main/examples/kv_events/online/main.go#L104-L113

	kvCacheIndexer, err := setupKVCacheIndexer(ctx)
	if err != nil {
		logger.Error(err, "failed to setup KVCacheIndexer")
		return err
	}


	// Setup events pool
	eventsPool := setupEventsPool(ctx, kvCacheIndexer.KVBlockIndex())
	eventsPool.Start(ctx)
	logger.Info("Events pool started and listening for ZMQ messages")

zhengkezhou1:
@vMaroon please take a look.

examples/kv_cache_index_service/server/server.go
```diff
@@ -0,0 +1,94 @@
...
+
+	indexerpb "github.com/llm-d/llm-d-kv-cache-manager/api"
+	"github.com/llm-d/llm-d-kv-cache-manager/examples/testdata"
+	"github.com/llm-d/llm-d-kv-cache-manager/pkg/kvcache"
+	"github.com/llm-d/llm-d-kv-cache-manager/pkg/kvcache/kvblock"
+	"github.com/llm-d/llm-d-kv-cache-manager/pkg/utils"
+)
+
+// IndexerService implements the IndexerServiceServer interface.
+type IndexerService struct {
```
vMaroon, line 30: I think it would be useful to have this contain a standalone service that has both the write and read paths of this project. The `IndexerService` can contain the `kvevents.Pool` as well and set it up properly. Then it would be usable as-is.

examples/kv_cache_index_service/server/main.go
```diff
@@ -0,0 +1,146 @@
...
+	// Wait for shutdown signal
+	<-ctx.Done()
+	logger.Info("Shutting down...")
+
+	// Graceful shutdown of events pool
+	shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer shutdownCancel()
+	eventsPool.Shutdown(shutdownCtx)
+
+	log.Printf("gRPC server listening on %s", *addr)
```
vMaroon: nit: use `Info` for consistency.

examples/kv_cache_index_service/server/main.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	defer publisher.Close()
+
+	indexerSvc, err := setupIndexerService(ctx)
+	if err != nil {
+		logger.Error(err, "failed to create indexer service")
+	}
+
+	// Initial query - should be empty since no events have been published
+	pods, err := indexerSvc.indexer.GetPodScores(ctx, testdata.Prompt, testdata.ModelName, nil)
```
zhengkezhou1, line 59: This step is mandatory; otherwise, the client will be unable to query the data. I do not understand why this is happening. As shown in the image, blockKey-4 should not exist.

Image

vMaroon:
@zhengkezhou1 thank you for the contribution, and the patience! Been a rough few weeks full of holidays. I think this looks good now - can you rebase and confirm that all updated examples work? Then would be good to go.

Thanks again!

zhengkezhou1:
The problem described in https://github.com/llm-d/llm-d-kv-cache-manager/issues/140 prevents us from verifying whether the current changes are correct.

vMaroon:
> The problem described in #140 prevents us from verifying whether the current changes are correct.

Since this was fixed - can you confirm that all is good?

zhengkezhou1:
sure, it's good.

client
image

server
image

vMaroon:
/lgtm
/approve

Thanks @zhengkezhou1 - great work.

PR #110, opened 2025-09-08
fix: Modified the download url of libtokenizers.darwin-x86_64.tar.gz

WillardHu:
The download url of libtokenizers for Darwin platforms is libtokenizers.darwin-x86_64.tar.gz not libtokenizers.darwin-amd64.tar.gz. Reference: https://github.com/daulet/tokenizers/releases


vMaroon:
Thanks @WillardHu - can you verify your commit signature?

PR #118, opened 2025-09-10
Fix indexer behavior when no kvblock-keys are generated

vMaroon:
## Summary 
`kvcache.Indexer` to return empty-scores if `kvblock.TokenProcessor` does not generate keys.


vMaroon:
I have tested that this works with the inference-scheduler as-is. Moving in for v0.3.1 due to urgency.

/lgtm
/approve

PR #119, opened 2025-09-12
feat: Add a SliceMapE function for handle errors and add unit tests

WillardHu:
### Overview
Sometimes, errors can occur when processing data transformations. The SliceMap function is relatively weak in error handling. We've add a SliceMapE function to handle these scenarios and added unit tests for both functions.


vMaroon:
Hi @WillardHu - I think abstractions are justified by use mostly. While this sounds reasonable, do you intend to propose using it instead of `SliceMap` in some instances?

PR #123, opened 2025-09-12
Implement auto-assign for reviewers without write permissions

vMaroon:
The problem:
- CODEOWNERS supports granular ownership and auto assignment, but requires write permissions (cannot effectively have reviewers who cannot merge requests)
- OWNERS with the current PROW bot does not have auto-assign / granular permissions

Until we get a custom or reasonable PROW bot, this is a good enough solution to auto-assign reviewers that do not have write permissions.


yankay:
Thanks @vMaroon

PR #137, opened 2025-09-28
[WIP] feat: Add UDS-based external tokenizer service

delavet:
This PR introduces a new Unix Domain Socket (UDS) based tokenizer service that can be used as an external tokenizer for the KV Cache Manager, related to https://github.com/llm-d/llm-d-kv-cache-manager/issues/126.  The changes include:

1. Added a new tokenizer mode in the KV Cache Manager, which communicate with an external tokenizer service via http over UDS.

2. Added an example external tokenizer service with:
   - Server implementation which can do both tokenization and chat-templating
   - Tokenizer implementation with HF (transformers) code
   - Dockerfile for containerization
   - Gunicorn configuration for production deployment
   - alongside documentation and tests

3. Updated the Helm chart to deploy the external tokenizer as a sidecar container alongside the KV Cache Manager

4. Added configuration options to enable the external tokenizer in the kv events online example.


services/uds_tokenizer/models/README.md
```diff
@@ -0,0 +1,129 @@
+# Model Caching in Tokenizer Service
```
vMaroon, line 1: I think that the `uds_tokenizer` package would be better housed in a new `services` directory.

services/uds_tokenizer/tokenizer_service/tokenizer.py
```diff
@@ -0,0 +1,270 @@
...
+    
+    def tokenize_and_process(self, prompt: str) -> BatchEncoding:
+        """
+        Tokenize the prompt with proper handling of special tokens.
+        
+        When using chat templates, we need to be careful not to add BOS tokens twice:
+        1. Once from the chat template itself (which may include BOS token)
+        2. Once from the add_special_tokens parameter
+        
+        vLLM handles this by setting add_special_tokens=False when using chat templates.
```
vMaroon, line 233: Do you think it would make sense to extract and reuse vLLM preprocessing code as-is? Serving as a lightweight vLLM sub-component for disaggregated tokenization. Its maintenance would be syncing up versions and dependencies.

Not a blocker, for this open PR, but I think we should aim towards this path. What do you think?

delavet: Agreed. I think we all agree that if we can adopt the same logic as vLLM in a lightweight manner, it would be the ideal form for this service. We will take this as the future development direction of this service and start to investigate on this matter.

pkg/tokenization/uds_tokenizer.go
```diff
@@ -0,0 +1,162 @@
...
+	client := &http.Client{
+		Transport: transport,
+		Timeout:   60 * time.Second,
+	}
+
+	return &UdsTokenizer{
+		httpClient: client,
+		baseURL:    baseURL,
+	}, nil
+}
```
guygir, line 100: HTTP over UDS has per-call marshaling and request overhead compared to our internal pre-processing. Would be interesting to benchmark and see the overhead cost of this approach, so we can measure the trade-off between internal and external tokenization appropriately.

pkg/tokenization/uds_tokenizer.go
```diff
@@ -0,0 +1,162 @@
...
+		return nil, nil, fmt.Errorf("server returned status %d: %s", resp.StatusCode, string(body))
+	}
+
+	var tokenized TokenizedInput
+	if err := json.Unmarshal(body, &tokenized); err != nil {
+		return nil, nil, fmt.Errorf("failed to unmarshal response: %w", err)
+	}
+
+	return tokenized.InputIDs, tokenized.OffsetMapping, nil
+}
```
guygir, line 135: (For a follow-up later on) Might be worth considering request batching to reduce the per-request overhead, if it turns out to be significant.

pkg/tokenization/uds_tokenizer.go
```diff
@@ -0,0 +1,162 @@
...
+	}, nil
+}
+
+// Encode tokenizes the input string and returns the token IDs and offsets.
+func (u *UdsTokenizer) Encode(input, modelName string) ([]uint32, []tokenizers.Offset, error) {
+	req, err := http.NewRequestWithContext(
+		context.Background(),
+		http.MethodPost,
+		u.baseURL+"/tokenize",
+		bytes.NewBuffer([]byte(input)),
```
guygir: (For a follow-up later on) Might be worth to consider per-request timeouts and retry logic, since tokenization is usually fast and the current global timeouts might be too generous for failure detection.

guygir: (For a follow-up later on) Might be worth trying to use strings.NewReader(input) to avoid the byte slicing and save on 1 allocation, if possible?

pkg/tokenization/uds_tokenizer.go
```diff
@@ -0,0 +1,162 @@
...
+	messagesBytes, err := json.Marshal(messages)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal messages: %w", err)
+	}
+
+	req, err := http.NewRequestWithContext(
+		context.Background(),
+		http.MethodPost,
+		u.baseURL+"/chat-template",
+		bytes.NewBuffer(messagesBytes),
```
guygir, line 148: (For a follow-up later on) Same as previous buffer.

services/uds_tokenizer/server.py
```diff
@@ -0,0 +1,360 @@
...
+                    raise
+            else:
+                logging.info("Tokenizer already initialized by another worker")
+                
+        finally:
+            # Release the lock
+            fcntl.flock(lock_file.fileno(), fcntl.LOCK_UN)
+            lock_file.close()
+    
+    return create_app()
```
guygir, line 353: I think there's a possible race condition here?
Both workers might pass the first check and end up calling initialize_tokenizer() twice, which would load the model twice and use 2x memory. Result should still be correct though. Maybe should use a process-global lock?

guygir (commented):
Looks great @delavet, Thank you very much!
Added some minor comments mostly for what I think are possible follow-up optimizations.

pkg/tokenization/uds_tokenizer.go
delavet: Thanks for the suggestion. The current global 60-second timeout is indeed too generous. I have added a more appropriate timeout and retry mechanism.

pkg/tokenization/uds_tokenizer.go
delavet: This is a good suggestion which requires further evaluation. As a follow-up, we plan to first conduct a performance benchmark of the tokenizers to determine the benefits of this approach.

pkg/tokenization/uds_tokenizer.go
delavet: use strings.NewReader(input) now

pkg/tokenization/uds_tokenizer.go
delavet: this `messagesBytes` is a byte slice, so I keep `bytes.NewBuffer` here.

services/uds_tokenizer/server.py
delavet: Here, file locks are used to enforce locking among different workers. Meanwhile, after a worker acquires the lock, it will perform a secondary confirmation on the `tokenizer_service`. This should help prevent the race condition mentioned above, IMO.

PR #138, opened 2025-10-01
Align with recent vLLM kv-block hashing changes

### LINKED ISSUE #127: Sync up with hashing/KVEvents changes in vLLM
vLLM PR https://github.com/vllm-project/vllm/pull/23673

vMaroon:
## Summary

In a recent [PR](https://github.com/vllm-project/vllm/pull/23673), vLLM's internal kv-block hash management underwent _unintended_ breaking changes that are not backwards compatible. For the coming release, we will sync with the changes, and sooner than later break the hashing dependency in indexing.

## Related Issues
- Fixes #127

### DIFF
--- a/examples/kv_events/offline/main.go
+++ b/examples/kv_events/offline/main.go
@@ -25,6 +25,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/llm-d/llm-d-kv-cache-manager/pkg/utils"
 	"github.com/vmihailenco/msgpack/v5"
 	"k8s.io/klog/v2"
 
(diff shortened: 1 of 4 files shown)


vMaroon:
I think I'll just break the dependency by two-way hashing, that is chunking the tokens carried by the KVEvents, hashing them independently and mapping the KVEvents hashes to the local ones. Today we sync on a few parameters to make use of the exact same hashes and avoid carrying the tokens too - but the latter is not done yet.

For testing, I ran the `73-capacity` benchmark I've been using recently (one in the blog), here are the results:
ttft_tpot_throughput_tripanel

The same order-of-magnitude implies that the indexing and hits are working properly. I also have this metric:
```
I1001 11:35:34.512602       1 collector.go:121] "metrics beat" logger="metrics" admissions=295064 evictions=294966 lookups=13245 hits=0 latency_count=13245 latency_sum=5.127642542999978 latency_avg=0.00038713797984144796
```

The latency is more or less the same as before (lost earlier logs, but they used to sit at 0.0004s on avg.
The hits counter broke in a PR earlier last month - thought it was fixed after but that got missed. Will fix for v0.4.

PR #142, opened 2025-10-07
Add support for local tokenizer files

pierDipi:
This adds the ability to load tokenizers from local files instead of
always downloading from HuggingFace. This is useful for air-gapped
environments and reduces first request latency time when tokenizers
are pre-loaded.

The implementation uses a composite tokenizer by default that tries
local files first and falls back to HuggingFace if needed and enabled.

LocalTokenizerConfig:
- Maps model names to local tokenizer.json file paths
- for the default behavior:
  - Auto-discovers tokenizers by walking LOCAL_TOKENIZER_DIR
  - Supports HF directory structure or (as fallback) directory nesting (e.g., Qwen/Qwen3)
  - Defaults to /mnt/models if LOCAL_TOKENIZER_DIR is not set
  - Example directory structure:
    ```
    /mnt/models/
      llama-7b/tokenizer.json          -> "llama-7b"
      Qwen/Qwen3/tokenizer.json        -> "Qwen/Qwen3"
      a/b/c/model/tokenizer.json       -> "a/b/c/model"
    ```

CompositeTokenizer:
- Tries each tokenizer in order until one succeeds
- Returns combined errors if all fail
- Enables graceful fallback from local to HF

Tests include unit tests for local tokenizer functionality, composite
tokenizer fallback behavior, and e2e tests with the KV cache indexer.

Signed-off-by: Pierangelo Di Pilato <pierdipi@redhat.com>

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -60,9 +60,10 @@ Here's a complete configuration example with all options:
   "tokenizersPoolConfig": {
     "workersCount": 8,
     "minPrefixOverlapRatio": 0.85,
-    "enabled": true,
-    "huggingFaceToken": "your_hf_token_here",
-    "tokenizersCacheDir": "/tmp/tokenizers",
+    "hf": {
+      "huggingFaceToken": "your_hf_token_here",
+      "tokenizersCacheDir": "/tmp/tokenizers"
+    },
     "local": {
       "autoDiscoveryDir": "/mnt/models",
       "autoDiscoveryTokenizerFileName": "tokenizer.json"
@@ -206,18 +207,27 @@ Configures the tokenization worker pool and cache utilization strategy.
 {
   "workersCount": 5,
   "minPrefixOverlapRatio": 0.8,
-  "huggingFaceToken": "",
-  "tokenizersCacheDir": ""
+  "hf": {
+    "enabled": true,
+    "huggingFaceToken": "",
+    "tokenizersCacheDir": ""
+  },
+  "local": {
+    "autoDiscoveryDir": "/mnt/models",
+    "autoDiscoveryTokenizerFileName": "tokenizer.json",
+    "modelTokenizerMap": {
+      "my-model": "/path/to/custom-model/tokenizer.json"
+    }
+  }
 }
 ```
 
-| Field | Type | Description | Default |
-|-------|------|-------------|--------|
-| `workersCount` | `integer` | Number of tokenization worker goroutines | `5` |
-| `minPrefixOverlapRatio` | `float64` | Minimum overlap ratio to use cached prefix tokens (0.0-1.0) | `0.8` |
-| `huggingFaceToken` | `string` | HuggingFace authentication token | `""` |
-| `tokenizersCacheDir` | `string` | Directory for caching tokenizers | `""` |
-
+| Field                   | Type                   | Description                                                 | Default |
+|-------------------------|------------------------|-------------------------------------------------------------|---------|
+| `workersCount`          | `integer`              | Number of tokenization worker goroutines                    | `5`     |
+| `minPrefixOverlapRatio` | `float64`              | Minimum overlap ratio to use cached prefix tokens (0.0-1.0) | `0.8`   |
+| `hf`                    | `HFTokenizerConfig`    | HuggingFace tokenizer config                                |         |
+| `local`                 | `LocalTokenizerConfig` | Local tokenizer config                                      |         |
 
 ### Local Tokenizer Configuration (`LocalTokenizerConfig`)
 
@@ -228,7 +238,7 @@ Configures loading tokenizers from local files. Useful for air-gapped environmen
   "autoDiscoveryDir": "/mnt/models",
   "autoDiscoveryTokenizerFileName": "tokenizer.json",
   "modelTokenizerMap": {
-    "custom-model": "/path/to/custom-model/tokenizer.json"
+    "my-model": "/path/to/custom-model/tokenizer.json"
   }
 }
 ```


pkg/tokenization/tokenizer.go
```diff
@@ -38,6 +40,7 @@ type Tokenizer interface {
 
 // HFTokenizerConfig holds the configuration for the HuggingFace tokenizer.
 type HFTokenizerConfig struct {
+	Enabled            bool   `json:"enabled"`
```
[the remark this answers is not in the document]
pierDipi: It is set in `DefaultHFTokenizerConfig()`

pkg/tokenization/pool.go
```diff
@@ -82,19 +89,36 @@ type Pool struct {
...
+
+	if config.LocalTokenizerConfig != nil && config.LocalTokenizerConfig.IsEnabled() {
+		localTokenizer, err := NewCachedLocalTokenizer(*config.LocalTokenizerConfig)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create local tokenizer: %w", err)
+		}
+		tokenizers = append(tokenizers, localTokenizer)
+	}
+
+	if config.HFTokenizerConfig != nil && config.HFTokenizerConfig.Enabled {
```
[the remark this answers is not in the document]
pierDipi: It's not a pointer

pkg/tokenization/tokenizer.go
```diff
@@ -46,29 +49,143 @@ type HFTokenizerConfig struct {
...
+
+// LocalTokenizerConfig provides a mapping from model names to local tokenizer.json file paths.
+// This allows the system to use pre-downloaded tokenizer files instead of fetching them from HuggingFace,
+// which is useful for air-gapped environments or when models are preloaded on disk.
+type LocalTokenizerConfig struct {
+	// Mapping is a map from model name to the absolute path of its tokenizer.json file.
+	// The model name (key) is typically the directory name containing the tokenizer.json file.
+	//
+	// Example mapping: {"model-a": "/mnt/models/model-a/tokenizer.json", ...}
+	Mapping map[string]string `json:"mapping,omitempty"`
```
pierDipi: if we envision needed multiple files, I think we can make the value a struct

pkg/tokenization/pool.go
```diff
@@ -39,15 +39,22 @@ type Config struct {
 	// Minimum overlap ratio to skip full tokenization and use cached prefix tokens.
 	MinPrefixOverlapRatio float64 `json:"minPrefixOverlapRatio"`
 	*HFTokenizerConfig
```
pierDipi: I didn't want to make a breaking change but ideally this `HFTokenizerConfig` is moved to a JSON key `hf` (or similar) so that the config doesn't have `enabled: false` (top-level) and then `local...: {} (local-level configs which turns the config to be enabled)
```yaml
# ...

minPrefixOverlapRatio: ...
enabled: false
local:
  ...
```

Let me know what's your preference since we could afford config-level breaking changes at this stage

pkg/tokenization/tokenizer.go
```diff
@@ -46,29 +49,143 @@ type HFTokenizerConfig struct {
...
+	localTokenizerDir      = os.Getenv("LOCAL_TOKENIZER_DIR")
+	localTokenizerFileName = os.Getenv("LOCAL_TOKENIZER_FILENAME")
+)
+
+// defaultLocalTokenizerDir is the default directory to search for local tokenizer files.
+// This is typically used in containerized environments where models are mounted at /mnt/models.
+//
+//nolint:gosec // These are default paths, not credentials
+const (
+	defaultLocalTokenizerDir      = "/mnt/models"
```
yankay, line 82: HI @pierDipi 

A truly great feature.  

Could we also include a script and documentation for downloading the Hugging Face tokenizer.json? This way, users will know how to obtain the tokenizer.json file in a non-air-gap environment.
```
from huggingface_hub import hf_hub_download

repo_id = "Qwen/Qwen3-0.6B"
filename = "tokenizer.json"

file_path = hf_hub_download(repo_id=repo_id, filename=filename)
#file_path  /root/.cache/huggingface/hub/models--Qwen--Qwen3-0.6B/snapshots/c1899de289a04d12100db370d81485cdf75e47ca/tokenizer.json
```

pkg/tokenization/tokenizer.go
```diff
@@ -46,29 +49,143 @@ type HFTokenizerConfig struct {
...
+//
+//	Single-level:
+//	  /mnt/models/llama-7b/tokenizer.json       -> "llama-7b"
+//
+//	Two-level (HuggingFace org/model format):
+//	  /mnt/models/Qwen/Qwen3/tokenizer.json     -> "Qwen/Qwen3"
+//
+//	Arbitrary nesting:
+//	  /mnt/models/a/b/c/tokenizer.json          -> "a/b/c"
+//
```
yankay, line 146: Another Idea is：

If using hf_hub_download, the directory structure is very consistent, such as:

```
/root/.cache/huggingface/hub/models--Qwen--Qwen3-0.6B/snapshots/c1899de289a04d12100db370d81485cdf75e47ca/tokenizer.json
/root/.cache/huggingface/hub/models--meta-llama--Llama-2-7b-chat-hf/snapshots/f5db02db724555f92da89c216ac04704f23d4590/tokenizer.json
/root/.cache/huggingface/hub/models--meta-llama--Llama-3.2-1B-Instruct/snapshots/9213176726f574b556790deb65791e0c5aa438b6/tokenizer.json
/root/.cache/huggingface/hub/models--meta-llama--Llama-3.1-8B-Instruct/snapshots/0e9e39f249a16976918f6564b8830bc894c89659/tokenizer.json
/root/.cache/huggingface/hub/models--facebook--opt-125m/.no_exist/27dcfa74d334bc871f3234de431e71c6eeba5dd6/tokenizer.json
/root/.cache/huggingface/hub/models--jason9693--Qwen2.5-1.5B-apeach/snapshots/615164d42ef211cb774910cafe5fbc63872366d6/tokenizer.json
/root/.cache/huggingface/hub/models--gpt2/snapshots/607a30d783dfa663caf39e06633721c8d4cfcd7e/tokenizer.json
/root/.cache/huggingface/hub/models--meta-llama--Llama-3.2-3B-Instruct/snapshots/0cb88a4f764b7a12671c53f0838cd831a0843b95/tokenizer.json
/root/.cache/huggingface/hub/models--Qwen--Qwen2.5-0.5B-Instruct/snapshots/7ae557604adf67be50417f59c2c2f167def9a775/tokenizer.json
/root/.cache/huggingface/hub/models--meta-llama--Meta-Llama-3.1-8B-Instruct/snapshots/0e9e39f249a16976918f6564b8830bc894c89659/tokenizer.json
```

Therefore, if we walk through the Hugging Face cache directory, we can automatically retrieve all model names and tokenizer.json files.
Using Hugging Face’s directory structure might be better than a custom directory structure.

pkg/tokenization/tokenizer.go
```diff
@@ -46,29 +49,143 @@ type HFTokenizerConfig struct {
...
+// LocalTokenizerConfig provides a mapping from model names to local tokenizer.json file paths.
+// This allows the system to use pre-downloaded tokenizer files instead of fetching them from HuggingFace,
+// which is useful for air-gapped environments or when models are preloaded on disk.
+type LocalTokenizerConfig struct {
+	// Mapping is a map from model name to the absolute path of its tokenizer.json file.
+	// The model name (key) is typically the directory name containing the tokenizer.json file.
+	//
+	// Example mapping: {"model-a": "/mnt/models/model-a/tokenizer.json", ...}
+	Mapping map[string]string `json:"mapping,omitempty"`
+}
```
yankay, line 111: The variable name **"Mapping"** seems somewhat inappropriate. It might be better to use a more descriptive name like **"modelTokenizerMap"** or  **"localPaths"** or something similar.

pkg/tokenization/tokenizer.go
yankay: I'm not entirely sure if I'm correct — if I'm wrong, please feel free to ignore this.  

Here, the **Map** (or directory-based approach) doesn't seem to require any configuration, because the tokenizer files can be automatically discovered by walking through the directory.  

On the other hand,  
```
localTokenizerDir = os.Getenv("LOCAL_TOKENIZER_DIR")
```
does require configuration.

pkg/tokenization/tokenizer.go
```diff
@@ -46,29 +49,143 @@ type HFTokenizerConfig struct {
...
 		TokenizersCacheDir: getTokenizerCacheDir(),
 	}
 }
 
-// CachedHFTokenizer is implements the Tokenizer interface using
-// bindings to HuggingFace's rust tokenizer.
+// localTokenizerDir is the base directory for local tokenizer files.
+// It can be set via the LOCAL_TOKENIZER_DIR environment variable.
+// If not set, it defaults to defaultLocalTokenizerDir.
+var (
```
vMaroon, line 72: I think it would be better to keep all parameters contained in the relevant configuration structs - and higher level entities (such as the scorer) would be responsible for initializing them properly.

E.g., today the huggingface token is a configuration field, but the `precise-prefix-cache-scorer` in the inference-scheduler allows for setting it through the plugins ConfigMap or through an environment variable.

pkg/tokenization/tokenizer.go
```diff
@@ -128,3 +271,76 @@ func getTokenizerCacheDir() string {
...
+// Returns an error if the model name is not found in the mapping.
+func (p *localTokenizerProvider) get(modelName string) (*tokenizers.Tokenizer, error) {
+	path, ok := p.cfg.Mapping[modelName]
+	if !ok {
+		return nil, fmt.Errorf("tokenizer for model %q not found", modelName)
+	}
+	return tokenizers.FromFile(path)
+}
+
+// CompositeTokenizer implements the Tokenizer interface with a fallback mechanism.
```
vMaroon, line 404: This and all the new configurations should be reflected in `docs/architecture.md` and `docs/configuration.md`. I think we can set the example below as the recommended default.

pkg/tokenization/tokenizer.go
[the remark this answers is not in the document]
liu-cong: what's the motivation behind adding this flag vs. checking the config is empty/null?

pkg/tokenization/tokenizer.go
liu-cong: Instead of picking between auto discovery from a path vs. manual mapping, would it be more user friendly to combine both, something like:

```
type LocalTokenizerConfig struct {
   AuthDiscoveryPath string
   TokenizerMappingOverride map[string]string
}
```

This allows both approaches to work together, so users won't get confused when they configure both the path and the mapping.

pierDipi: I enabled the use of both at the same time and through plugin configuration, left env variables as default fallback mechanism when unspecified

pkg/tokenization/tokenizer.go
pierDipi: I've added documentation

pkg/tokenization/tokenizer.go
```diff
@@ -46,29 +50,227 @@ type HFTokenizerConfig struct {
...
+	}
+	if localTokenizerFileName == "" {
+		localTokenizerFileName = defaultLocalTokenizerFileName
+	}
+}
+
+// LocalTokenizerConfig provides a mapping from model names to local tokenizer.json file paths.
+// This allows the system to use pre-downloaded tokenizer files instead of fetching them from HuggingFace,
+// which is useful for air-gapped environments or when models are preloaded on disk.
+type LocalTokenizerConfig struct {
```
liu-cong, line 98: Can we just adopt the HF file name format as the only auto discovery format for now (unless we have use cases that demand something else)? So basically we will default the AutoDiscoveryDir to `~/.cache/huggingface/hub/`.

docs/configuration.md
```diff
@@ -192,22 +197,74 @@ Configures the tokenization worker pool and cache utilization strategy.
...
+    "custom-model": "/path/to/custom-model/tokenizer.json"
+  }
+}
+```
+
+| Field | Type | Description | Default |
+|-------|------|-------------|---------|
+| `autoDiscoveryDir` | `string` | Directory to recursively scan for tokenizer files. Can be set via `LOCAL_TOKENIZER_DIR` environment variable. | `"/mnt/models"` |
+| `autoDiscoveryTokenizerFileName` | `string` | Filename to search for during auto-discovery. Can be set via `LOCAL_TOKENIZER_FILENAME` environment variable. | `"tokenizer.json"` |
+| `modelTokenizerMap` | `map[string]string` | Manual mapping from model name to tokenizer file path. Takes precedence over auto-discovery. | `{}` |
```
liu-cong: Does it replace the auto discovery map entirely, or does it just override on a per-model basis?

pkg/tokenization/tokenizer.go
pierDipi: We have use cases to support non-HF models (like S3 or compatible storage, OCI, etc), not all models come from Hugging Face or follow its specific prefix and snapshot/hash structure. This fallback gives users the ability to use models from other sources (e.g., private object storage) with their own folder structures, without forcing them to reorganize everything to mimic the HF layout.

docs/configuration.md
pierDipi: overrides, I'll update the description to improve clarity

pkg/tokenization/pool.go
liu-cong: +1 

Sorry I missed this comment before, and now I understand why you had an `enabled` here. However that makes it really confusing, because the `enabled` is at the root level. If I have `enabled:false` but it reads like I disabled the tokenizer completely, whereas in fact the local tokenizer is enabled if I provide the `local` config.

My preference is to make this change. We haven't set any rules on making such changes yet I think. And at this moment the llm-d well-lit paths are probably the only consumer? So I think it's relatively safe to do so.
@vMaroon do you see any risks of making this breaking change?

liu-cong:
> The core issue, which I noted in https://github.com/llm-d/llm-d-kv-cache-manager/pull/142#issuecomment-3453339650, is that we have integration use cases where the default discovery path itself needs to be configurable at a system level.

@pierDipi Thanks for explaining the use case. This makes sense to me

pkg/tokenization/pool.go
vMaroon: No risks, please go ahead and do it @pierDipi - should be the final piece before merging. Sorry for the delays in reviewing, had a lot of traveling in the past few weeks.

liu-cong: Same here, this is the last bit. Will lgtm after this. Thanks for your patience!

pierDipi: Latest commit:
- [Use 'hf' key to configure HuggingFace tokenizer](https://github.com/llm-d/llm-d-kv-cache-manager/pull/142/commits/8d3211c54d6621f4efcce34bfb97e521672685bd)

docs/configuration.md
```diff
@@ -206,18 +207,27 @@ Configures the tokenization worker pool and cache utilization strategy.
 {
   "workersCount": 5,
   "minPrefixOverlapRatio": 0.8,
-  "huggingFaceToken": "",
-  "tokenizersCacheDir": ""
+  "hf": {
+    "enabled": true,
```
liu-cong, line 211: we don't need `enabled` anymore, right?

pierDipi: Currently, the scheduler sets the user config as "overlay" of the default [ref](https://github.com/llm-d/llm-d-inference-scheduler/blob/091f07e7809160bc062be2199500f2675e9cf734/pkg/plugins/scorer/precise_prefix_cache.go#L42-L56), which makes setting `hf` to `null` impossible without the flag :(

pierDipi: do you see a different way / path here?

pierDipi: we could use the `hf.tokenizersCacheDir = ""` as a proxy for `enabled = false` but it's not very intuitive IMO

liu-cong: We should change  the `PrecisePrefixCachePluginFactory` method, so it only applies the default when local tokenizer config is not supplied.

liu-cong: @vMaroon Can you advise the release procedure between the kv-cache-manager and inference-scheduler repos if we make this change?

vMaroon: We usually sync up and update the inference-scheduler after cutting releases here. So "breaking changes" are ok now.

Issue #144, opened 2025-10-08
[Minor Bug] `hits` metric does not work


vMaroon:
The `hits` metric today is not getting updated at all - a PR removed its updates and no followup was made.

kyanokashi:
Has this been resolved?

PR #146, opened 2025-10-10
Fix LookupHits metrics not work

Frapschen:
issue: #144

### DIFF
--- a/pkg/kvcache/kvblock/instrumented_index.go
+++ b/pkg/kvcache/kvblock/instrumented_index.go
@@ -64,15 +64,11 @@ func recordHitMetrics(keyToPods map[Key][]string) {
 	podCount := make(map[string]int)
 	for _, pods := range keyToPods {
 		for _, p := range pods {
-			if _, ok := podCount[p]; ok {
-				podCount[p]++
-			} else {
-				// First time seeing this pod in current lookup window.
-				// Reset to 1 because counts are local to this call (not cumulative over time).
-				// This ensures compatibility with sliding window attention (SWA) and cache eviction,
-				// where only recent hits within the active window are considered.
-				podCount[p] = 1
-			}
+			// First time seeing this pod in current lookup window.
+			// set to 1 because counts are local to this call (not cumulative over time).
+			// This ensures compatibility with sliding window attention (SWA) and cache eviction,
+			// where only recent hits within the active window are considered.
+			podCount[p]++
 		}
 	}
 


pkg/kvcache/kvblock/instrumented_index.go
```diff
@@ -55,6 +55,7 @@ func (m *instrumentedIndex) Lookup(
 	metrics.LookupRequests.Inc()
 
 	pods, err := m.next.Lookup(ctx, keys, podIdentifierSet)
+	metrics.LookupHits.Add(float64(len(pods)))
```
vMaroon: The hits refer to the largest number of KV-blocks hit on a pod (though what you added could be interesting as a new metric!).

`pods` is a map of key -> pod-identifiers where the prefix trait is not necessarily present (it is checked at a later stage of the higher call, when scoring pods). The key to solving this issue is to try to avoid the duplication if possible.

pkg/kvcache/kvblock/instrumented_index.go
```diff
@@ -55,6 +55,29 @@ func (m *instrumentedIndex) Lookup(
 	metrics.LookupRequests.Inc()
 
 	pods, err := m.next.Lookup(ctx, keys, podIdentifierSet)
+	go recordHitMetrics(pods)
 
 	return pods, err
 }
+
+func recordHitMetrics(pods map[Key][]string) {
```
vMaroon: not introduced here, but better naming can make the code clearer:

pods -> keyToPods
pod -> pods (... range pods)

pkg/kvcache/kvblock/instrumented_index.go
```diff
@@ -55,6 +55,29 @@ func (m *instrumentedIndex) Lookup(
...
+	podCount := make(map[string]int)
+	for _, pod := range pods {
+		for _, p := range pod {
+			if _, ok := podCount[p]; ok {
+				podCount[p]++
+			} else {
+				podCount[p] = 1
```
vMaroon: A reset here makes it so that counts are not necessarily from the beginning. I think this is ok though, since the inference engine should make sure suffixes are evicted first, plus this works for SWA (sliding window attention). Could be helpful to document this here.

vMaroon (commented):
Added two minor comments then good to go, thanks!

pkg/kvcache/kvblock/instrumented_index.go
Frapschen: > not introduced here
Any advice I should put the code?

pierDipi:
@Frapschen FYI, linter is failing
```
S1036: unnecessary guard around map access (staticcheck)
CI - PR Checks / lint-and-test
```

Frapschen:
not relate error:
```
=== RUN   TestInMemoryIndexBehavior/ConcurrentOperations
    index_test.go:220: 
        	Error Trace:	/home/runner/work/llm-d-kv-cache-manager/llm-d-kv-cache-manager/pkg/kvcache/kvblock/index_test.go:220
        	            				/opt/hostedtoolcache/go/1.24.1/x64/src/runtime/asm_amd64.s:1700
        	Error:      	map[kvblock.Key][]string{} does not contain kvblock.Key{ModelName:"test-model", ChunkHash:0x3e8}
        	Test:       	TestInMemoryIndexBehavior/ConcurrentOperations
    index_test.go:221: 
        	Error Trace:	/home/runner/work/llm-d-kv-cache-manager/llm-d-kv-cache-manager/pkg/kvcache/kvblock/index_test.go:221
        	            				/opt/hostedtoolcache/go/1.24.1/x64/src/runtime/asm_amd64.s:1700
        	Error:      	[]string(nil) does not contain "pod-20-3"
        	Test:       	TestInMemoryIndexBehavior/ConcurrentOperations
```

pkg/kvcache/kvblock/instrumented_index.go
```diff
@@ -55,6 +55,29 @@ func (m *instrumentedIndex) Lookup(
 	metrics.LookupRequests.Inc()
 
 	pods, err := m.next.Lookup(ctx, keys, podIdentifierSet)
+	go recordHitMetrics(pods)
 
 	return pods, err
 }
+
+func recordHitMetrics(keyToPods map[Key][]string) {
```
vMaroon, line 63: For a follow-up PR (will open an issue based on this comment) - wonder if this better fits to be recorded by the `kvblock.Scorer` and not the index.

vMaroon:
@Frapschen please verify your commit signature (following this guide: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification), then good to go! Thank you for the contribution and sorry for the delay.

/lgtm
/approve

PR #150, opened 2025-10-24
Implementation for Tiering in KV-Cache-Manager

Jay-Pd:
This PR is the initial implementation for Tiered Scoring in llm-d KV-Cache-manager.

This PR introduces the following changes.

1. Adds Backend Configuration struct
2. Adds Backend Configurations to Pool, Indexer and the Scorer. 
3. Modifies the Indexer so that PodIdentifier along with the DeviceTier information is returned.
4. Modifies the scorer so that tier-aware scoring is done.
5. Modifies the tests so that Device Tier is also considered.

Changes left to be implemented as part of the [proposal](https://docs.google.com/document/d/1RXVJ8Bmuf_BWxUJoul0OKN8SVJrJzks41_hx14ogSUM/edit?pli=1&tab=t.s0kedl7ipu1j) are:

1. Create Custom Configuration functions in the Pool, Indexer and Scorer for custom Backends config.
2. Create a BackendConfiguration CRD that can be used to load custom backends on startup.
3. Make corresponding changes in llm-d inference-scheduler to load and pass backend configuration to the Indexer and the Pool when plugin is created.
4. Add environment variables in llm-d helm charts so that the directory of the Backend Configuration CR is known to the llm-d inference-scheduler.
5. Add a Config Loader functionality to load the Backend Configuration from the YAML, similar to [EndpointPickerConfig](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/d788a2c2dc45c00b393c46c23bd09f0e949dff79/site-src/guides/epp-configuration/config-text.md) in GIE-EPP.

Please note that there are ongoing discussions and the implementation may change over time. 
Please contribute to the discussion on this [doc](https://docs.google.com/document/d/1RXVJ8Bmuf_BWxUJoul0OKN8SVJrJzks41_hx14ogSUM/edit?pli=1&tab=t.s0kedl7ipu1j). 

Thank you!

@vMaroon


ahg-g:
just for the sake of clarify, this is not actually adding a "CRD", it is just adding a new parameter to the plugin config type

pkg/kvcache/kvevents/pool.go
```diff
@@ -247,6 +246,16 @@ func (p *Pool) digestEvents(ctx context.Context, podIdentifier, modelName string
 	for _, event := range events {
 		switch ev := event.(type) {
 		case BlockStored:
+			deviceTier := ""
+			if ev.Medium == nil {
+				deviceTier = "gpu"
+			} else {
+				deviceTier = *ev.Medium
+			}
```
liu-cong:
```suggestion
            // Default to gpu. 
            // For non-gpu events, vLLM KV event has a non-empty Medium field.
			medium := "gpu"
			if ev.Medium = !nil {
				medium = *ev.Medium
			}
```

pkg/kvcache/kvevents/pool.go
```diff
@@ -270,6 +279,16 @@ func (p *Pool) digestEvents(ctx context.Context, podIdentifier, modelName string
 			}
 
 		case BlockRemoved:
+			deviceTier := ""
```
liu-cong: Same suggestion here.

pkg/kvcache/kvblock_scorer.go
```diff
@@ -58,54 +62,88 @@ type KVBlockScorer interface {
...
-	// set initial score of 1
-	// pods not in the first key will retain the default score of 0.
-	for _, pod := range podsForFirstKey {
-		podScores[pod] = 1
-	}
+	for i := 0; i < len(keys); i++ {
+		podsForKey := keyToPods[keys[i]]
+		currentPodsSet := sets.NewString(podsForKey...)
+
+		if i == 0 {
```
liu-cong: I like the previous code structure (dealing with first key separately) better, it's more readable

pkg/kvcache/kvblock_scorer.go
```diff
@@ -58,54 +62,88 @@ type KVBlockScorer interface {
...
+}
 
 // Strategy returns the strategy type: LongestPrefixMatch.
 func (s *LongestPrefixScorer) Strategy() KVScoringStrategy {
 	return LongestPrefixMatch
 }
 
-// Score implements the longest prefix scoring logic.
+// Score implements the longest prefix scoring logic with weighted sum based on BackendConfig.
 func (s *LongestPrefixScorer) Score(keys []kvblock.Key, keyToPods map[kvblock.Key][]string) (map[string]int, error) {
```
liu-cong: let's change the return score to float64, the consumer (inference scheduler) will need to convert int to float64 anyway

pkg/kvcache/kvblock_scorer.go
```diff
@@ -58,54 +62,88 @@ type KVBlockScorer interface {
...
-		activePods = activePods.Intersection(currentPodsSet)
-		for pod := range activePods {
-			// increment score for each pod in the intersection
-			podScores[pod]++
+			// Add weight for each match using PodIdentifier as key
+			podScores[podIdentifier] += weight
 		}
 	}
 
+	// Convert float64 scores to int
```
liu-cong: save the conversion here by changing return type to float

pkg/kvcache/kvevents/pool.go
```diff
@@ -247,6 +246,16 @@ func (p *Pool) digestEvents(ctx context.Context, podIdentifier, modelName string
...
 		case BlockStored:
+			deviceTier := ""
+			if ev.Medium == nil {
+				deviceTier = "gpu"
+			} else {
+				deviceTier = *ev.Medium
+			}
+
+			// Create PodEntry for this specific event's device tier
+			podEntries := []kvblock.PodEntry{{PodIdentifier: podIdentifier, DeviceTier: deviceTier}}
```
liu-cong, line 261: what happens if the same block hash exists on multiple mediums? Can this code correctly handle that ?

pkg/kvcache/kvblock_scorer.go
```diff
@@ -58,54 +62,88 @@ type KVBlockScorer interface {
...
+// in the format "PodIdentifier@DeviceTier". If no '@' is found, the original string
+// is returned as PodIdentifier and DeviceTier is empty.
+//
+// It returns:
+// 1. podIdentifier: the extracted PodIdentifier.
+// 2. deviceTier: the extracted DeviceTier.
+//
+//nolint:gocritic // unnamedResult: conflicts with nonamedreturns linter preference
+func extractPodIdentifierAndTier(podString string) (string, string) {
+	if idx := strings.Index(podString, "@"); idx != -1 {
```
liu-cong: Don't like the string parsing here. I was wondering if we can just pass a pod struct instead of string to the `Score` function.

pkg/kvcache/kvblock_scorer.go
```diff
@@ -35,12 +37,14 @@ const (
 // KVBlockScorerConfig holds the configuration for the KVBlockScorer.
 type KVBlockScorerConfig struct {
 	ScoringStrategy KVScoringStrategy
+	BackendConfigs  map[string]*backend.BackendConfig `json:"backendConfigs"`
```
liu-cong: I sugest something like this:

```
KVMediumsConfig []*KVMediumConfig 
```

```
type KVMediumConfig struct {
  Name string
  Weight float64
}
```

For the following reasons:
1. I think the config is part of this scorer so let's keep it here instead of adding it to `backend`
2. Using a list instead of map so a) it's self explanatory which fields means what 2) it's more flexible (e.g., we may need to change map key in the future). It does mean we need a bit of validation to ensure no duplicate Names, but this is common in many k8s resources.

pkg/kvcache/kvblock_scorer.go
Jay-Pd: Removed the string parsing function and passed pod struct directly.

pkg/kvcache/kvevents/pool.go
Jay-Pd: Since a pod entry is the pod identifier as well as the medium, it will be able to handle such a case.

pkg/kvcache/kvblock_scorer.go
Jay-Pd: converted to float64.

pkg/kvcache/kvevents/pool.go
Jay-Pd: Refactored to remove duplicate code.

pkg/kvcache/kvblock_scorer.go
Jay-Pd: Added the KVMediumsConfig struct.
The Scorer will parse this above struct and create a map to use while scoring.

pkg/kvcache/kvevents/pool.go
```diff
@@ -247,6 +246,16 @@ func (p *Pool) digestEvents(ctx context.Context, podIdentifier, modelName string
 	for _, event := range events {
 		switch ev := event.(type) {
 		case BlockStored:
+			// Default to gpu.
+			// For non-gpu events, vLLM KV event has a non-empty Medium field.
+			deviceTier := "gpu"
```
vMaroon: `"gpu"` should be in a const and used here and on line 284.

pkg/kvcache/indexer.go
```diff
@@ -38,6 +39,7 @@ type Config struct {
 	KVBlockIndexConfig   *kvblock.IndexConfig          `json:"kvBlockIndexConfig"`
 	KVBlockScorerConfig  *KVBlockScorerConfig          // not exported
 	TokenizersPoolConfig *tokenization.Config          `json:"tokenizersPoolConfig"`
+	BackendConfigs       []*backend.BackendConfig      `json:"backendConfigs"`
```
vMaroon: I get the idea behind "Backend" and it does make sense as `kvcache.BackendConfig` since it reads as a kvcache-backend, but wondering if you had alternatives in mind.

BTW - this should be documented in `/docs/configuration.md`.

pkg/utils/backend/types.go
```diff
@@ -0,0 +1,31 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package backend
+
+type BackendConfig struct {
```
vMaroon: In one comment above I referred to this as `kvcache.BackendConfig` - I think kvcache should be a part of its name. If not the package, then here.

pkg/utils/backend/types.go
```diff
@@ -0,0 +1,31 @@
...
+	// Name is the identifier for this medium (e.g., "gpu", "cpu", "disk")
+	Name string `json:"name"`
+	// Weight is the scoring weight for blocks stored on this medium
+	Weight float64 `json:"weight"`
+}
+
+func DefaultBackendConfig() []*BackendConfig {
+	return []*BackendConfig{
+		{Name: "gpu", Weight: 1.0},
+		{Name: "cpu", Weight: 0.5},
```
vMaroon: In a test we ran with an enhanced CPU offloading connector implementation and a light workload, TTFTs were as follows:
- GPU kvcache prefill: 0.12s
- CPU kvcache prefill (transport and prefill): 0.15s

Based on these, we can give CPU a default of 0.8.

pkg/kvcache/kvevents/pool.go
Jay-Pd: Resolved.

pkg/kvcache/indexer.go
Jay-Pd: Added documentation to `/docs/configuration.md`

pkg/utils/backend/types.go
Jay-Pd: I have renamed it to `backend.KVCacheBackendConfig` to make it verbose.

pkg/utils/backend/types.go
Jay-Pd: Resolved. Is it possible to share more details on the benchmark?

pkg/kvcache/kvblock_scorer.go
```diff
@@ -51,61 +52,92 @@ type KVBlockScorer interface {
...
 	}
 
-	for i := 1; i < len(keys); i++ {
-		if activePods.Len() == 0 {
-			break
+	// For each pod, calculate consecutive score from start
+	for podID := range allPods {
+		score := 0.0
+
+		for _, key := range keys {
```
vMaroon: This implementation can get expensive - in the previous one, we first initialized the scored pods by those mapped for the first key, and then iterated over the keys, intersecting the set of active pods with the mappings for each key.

In this one, for every pod we iterate over all the keys.

Additionally, in today's duplicative offloading behavior, if a prompt has a prefix in GPU, and an extended prefix in CPU, the overlap in CPU should not affect the total score. (In the other cases of GPU only or CPU only, scoring is trivial).

The existing scoring algorithm can be used as-is, but should only get the weighting right + account for not duplicating scoring in overlap case.

vMaroon (commented):
Great work @Jay-Pd, this is a very useful feature. Thank you for the contributions and patience!

/lgtm

Will approve after in-cluster testing is confirmed.

vMaroon (commented):
/lgtm
/approve

pkg/kvcache/indexer.go
```diff
@@ -126,7 +130,7 @@ func (k *Indexer) KVBlockIndex() kvblock.Index {
 // The function returns a map of pod identifiers to scores.
 func (k *Indexer) GetPodScores(ctx context.Context, prompt, modelName string,
 	podIdentifiers []string,
-) (map[string]int, error) {
+) (map[string]float64, error) {
```
vMaroon, line 133: Pointing out for inference-scheduler import update:

This API change will be pulled.

vMaroon: @pierDipi

docs/configuration.md
```diff
@@ -33,6 +34,7 @@ The main configuration structure for the KV Cache Indexer module.
 | `tokenProcessorConfig` | [TokenProcessorConfig](#token-processor-configuration-tokenprocessorconfig) | Configuration for token processing | See defaults |
 | `kvBlockIndexConfig` | [IndexConfig](#index-configuration-indexconfig) | Configuration for KV block indexing | See defaults |
 | `tokenizersPoolConfig` | [Config](#tokenization-pool-configuration-config) | Configuration for tokenization pool | See defaults |
+| `kvCacheBackendConfigs` | [KVCacheBackendConfig](#kv-cache-backend-configuration-kvcachebackendconfig) | Configuration for KV Cache Device Backends | See defaults |
```
vMaroon, line 37: Pointing out for inference-scheduler import update:

This may optionally need to be reflected in the inference-scheduler docs too.

vMaroon: @pierDipi

PR #162, opened 2025-11-17
[KV-Events] Add vLLM KV cache events demo with ZMQ listener

sagearc:
### Summary
Adds a local demo that captures and emits KV events generated by vLLM for custom prompts.

### What's Added
`vllm_kv_cache_demo.py` - async ZMQ listener that captures vLLM KV-events during inference with custom prompts.

### Use Case
Local tool for understanding vLLM cache behavior and examining cache hashes during development.

#189

### DIFF
--- a/examples/kv_events/vllm/vllm_kv_cache_demo.py
+++ b/examples/kv_events/vllm/vllm_kv_cache_demo.py
@@ -1,5 +1,17 @@
-# SPDX-License-Identifier: Apache-2.0
-# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+# Copyright 2025 The llm-d Authors.
+#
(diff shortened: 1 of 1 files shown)


examples/kv_events/vllm/vllm_kv_cache_demo.py
```diff
@@ -0,0 +1,185 @@
...
+from vllm.distributed.kv_events import (
+    AllBlocksCleared,
+    BlockRemoved,
+    BlockStored,
+    KVEventBatch,
+)
+from vllm.engine.arg_utils import EngineArgs
+from vllm.lora.request import LoRARequest
+
+# Update these constants at the top of the file
```
vMaroon: This and other similar comments should be removed (seems to be AI)

PR #165, opened 2025-11-18
feat: KV-block hashing consistency between vLLM and llmd

sagearc:
## Context

This PR updates the LLMD block hashing method to maintain compatibility with a recent change in the upstream vLLM project.

The vLLM block hashing method was updated to use the LoRA adapter name (string) instead of the previous integer ID (vllm-project/vllm#27211). This alignment is necessary for correct KV-cache lookups in LLMD.

## Key Changes

1. **LoRA Hashing Support:** The LLMD block hashing logic is updated to correctly incorporate the LoRA adapter name in the request hash calculation. Previously, LLMD did not support LoRA request hashing.
2. **Test Coverage:** New tests have been added to verify the behavior of the request hash generation both with and without LoRA adapter information.

## Future Work & Notes

- **KV Event Hashing:** The upstream vLLM project is planning an update to the KV block stored event (vllm-project/vllm#27577). This will eventually allow us to retrieve all necessary hashing information directly from the events.
- **Test Data Source:** The KV events used for the current testing were gathered using an offline, local vLLM setup, as referenced in issue #162.


sagiahrac:
Closing as duplicate of #162

Issue #170, opened 2025-11-21
The online example cannot deploy


delavet:
I encountered an error while trying to run the online example:

<img width="643" height="153" alt="Image" src="https://github.com/user-attachments/assets/76e3727b-4ded-4515-b802-2a5765a841f6" />

It seems that there is a small issue with the chart format.

yankay:
fixed by #171

PR #171, opened 2025-11-21
fix online example chart format

delavet:
Fixed two minor issues related to the online example chart. This should solves #170 .


delavet:
It seems that there might be some issues with the test. I fixed it on the fly, so please help review it. Thank you.

vMaroon:
Great, thanks @delavet.

/lgtm
/approve

Issue #108, opened 2025-09-05
Profile the different `kvblock.Index` implementations


vMaroon:
As of writing this issue, we have three `kvblock.Index` implementations. An index contains KV-block locality information and is updated when KV-Events are digested from the vLLM servers, each containing a set of block-hashes, admission or removal type and an identifier of the sender.

The `kvblock.Index` is the heart of the kv-cache-manager and its performance is essential for low-latency and accurate scheduling.

The implementations are:
1. `InMemoryIndex` (default) - an in-memory index backed by a hashicorp LRU cache implementation, with relaxed locking
2. `CostAwareMemoryIndex` - contributed by @yankay, a `hypermodeinc/ristretto` memory bound high-performance in-memory cache
3. `Redis` - an index backed by a Redis server

Profiling the three options in **realistic**/high-scale workloads inputs would be helpful for drawing configuration recommendations and general transparency. The promotion of the 2nd option to default would also be desirable given good profiling data.

kartikx:
@vMaroon I would like to work on this issue.

vMaroon:
@kartikx thank you for your interest! Please note the milestone deadline and feel free to reach out on the [llm-d slack](https://join.slack.com/t/llm-d/shared_invite/zt-3ctao588l-MMCCEL~fn5_wfYXobsFIxg) anytime.

kartikx:
@vMaroon Do you have any suggestions on what a "realistic" workload looks like? Currently, I have the following setup

1. Add 100,000 keys to the index at once, and profile the time for the `Add`. 
2. Lookup all keys and profile the time for the `Lookup`.

All of these are added to the same pod. I repeat experiments 10 times for each index type, the keys are different on each run.

Current set of results for this:
```
[Cost] Add: 22.271475ms Lookup 21.758358ms
[Redis] Add: 22.533808ms Lookup 22.307492ms
[In memory] Add: 22.59525ms Lookup 22.8048ms
```

The numbers are pretty close, and the ordering doesn't stay consistent. Perhaps a better profiling workload can help make differences clear.

I can share a cleaned up version of my script for you to review if that helps.

vMaroon:
Thanks @kartikx - the script looks OK - the TODOs you outlined are a good improvement start, but we can start with a straightforward profiling test and iterate on it in PRs.

PR #175, opened 2025-11-22
feat: profiling kv cache index implementations (#108)

kartikx:
This PR adds a profiling test to compare the various `KVIndex` implementations.

It adds and looks up keys for each implementation, and reports the average over a specified number of trials.

```
go run -ldflags="-extldflags '-L$(pwd)/lib'" ./tests/profiling/kv_cache_index/main.go -trials 10 -keys 10000

[Cost Aware] Add: 7.516833ms Lookup 141.652299ms
[Redis] Add: 10.614779ms Lookup 6.190304ms
[InMemory] Add: 3.515912ms Lookup 127.265687ms
```

Requested Feedback:
1. Would it be better if I set this up as an actual test? I'm not sure if we want this to run in CI and on every build, so I set it up as a script for now.
2. The current workload simply generates a given number of keys. I'm open to suggestions on whether we should improve this.

Issue: #108

### DIFF
--- a/tests/profiling/kv_cache_index/main.go
+++ b/tests/profiling/kv_cache_index/main.go
@@ -1,0 +1,178 @@
+/*
+Copyright 2025 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"math/rand/v2"
+	"time"
+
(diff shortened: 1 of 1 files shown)


tests/profiling/kv_cache_index/main.go
```diff
@@ -0,0 +1,178 @@
...
+
+type IndexProfileResult struct {
+	AddTime time.Duration
+	LookupTime time.Duration
+}
+
+// TODO @kartikx: Use a more realistic workload if possible.
+func generateWorkloadKeys(numKeys int) []kvblock.Key {
+	// Uses time as seed to ensure that different profiling runs get different keys.
+	randGen := rand.New(rand.NewPCG(42, uint64(time.Now().UnixNano())))
```
sagearc, line 40: Can you generate the same workload keys for all profiling sessions? That way, the only difference will be the indexer implementation.

sagiahrac (changes requested):
Thanks for the contribution @kartikx!
Maybe we can refactor this into a standard Go benchmark using the testing package. This approach is better for project stability as it automatically adjusts `b.N` (the number of keys) to achieve stable timing, which allows us to remove the manual numTrials and easily integrate with CI. It also handles warmups automatically and dynamically.

You can split `Add` and `Lookup` into 2 separate benchmarks for simplicity, or reset the timer after the Add phase.

PR #176, opened 2025-11-23
[fs_connector][feat]: Add file management module (file_io) for fs connector

kfirtoledo:
This PR introduces the file management module used by the filesystem offloading connector.

Included in this PR:
- Adds file_io.cpp and file_io.hpp
- Implements low level file operations for the fs connector
- Supports atomic writes, file existence checks, and read operations

### DIFF
--- a/kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
+++ b/kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
@@ -33,10 +33,17 @@ namespace fs = std::filesystem;
 // -------------------------------------------------------------------
 // Define a larger buffer (1MB) to reduce syscall overhead and speed up I/O
 const size_t WRITE_BUFFER_SIZE = 1 * 1024 * 1024;  // 1MB buffer
+// Staging buffer tensor options (CPU, pinned memory)
+static const auto STAGING_BUFFER_TENSOR_OPTIONS = torch::TensorOptions()
+                                                      .dtype(torch::kUInt8)
+                                                      .device(torch::kCPU)
+                                                      .pinned_memory(true);
 // Allocate custom I/O buffer for this thread (replaces small default buffer)
 thread_local std::vector<char> thread_write_buffer(WRITE_BUFFER_SIZE);
-// Thread-local unique id for temporary files
-thread_local uint64_t thread_unique_id = std::random_device{}();
+
+// Thread-local unique suffix for temporary files
+thread_local std::string tmp_file_suffix =
+    "_" + std::to_string(std::random_device{}()) + ".tmp";
 // -------------------------------------------------------------------
 // file-IO Functions
 // -------------------------------------------------------------------
@@ -60,9 +67,8 @@ bool write_tensor_to_file(const torch::Tensor& cpu_tensor,
   }
 
   // Write to a temporary file to ensure atomic replace on rename
-  // Include thread_unique_id so each thread uses a unique temporary file
-  std::string tmp_path =
-      target_path + std::to_string(thread_unique_id) + ".tmp.";
+  // Include tmp_file_suffix so each thread uses a unique temporary file
+  std::string tmp_path = target_path + tmp_file_suffix;
 
   std::ofstream ofs(tmp_path, std::ios::out | std::ios::binary);
   if (!ofs) {
@@ -80,6 +86,13 @@ bool write_tensor_to_file(const torch::Tensor& cpu_tensor,
     std::cerr << "[ERROR] Failed to write to temporary file: " << tmp_path
               << " - " << std::strerror(errno) << "\n";
     std::remove(tmp_path.c_str());  // Clean up temp file
+    return false;
+  }
+
+  ofs.flush();
+  if (!ofs) {
+    std::cerr << "[ERROR] Failed to flush data to temporary file: " + tmp_path
+              << " - " << std::strerror(errno) << "\n";
     return false;
   }
 
@@ -106,7 +119,6 @@ bool read_tensor_from_file(const std::string& path, torch::Tensor& cpu_tensor) {
   }
 
   // Determine file size
-  ifs.seekg(0, std::ios::end);  // Move read pointer to END
   std::ifstream::pos_type end_pos = ifs.tellg();
   if (end_pos == std::streampos(-1)) {
     std::cerr << "[ERROR] Failed to determine file size: " << path << "\n";
@@ -135,18 +147,12 @@ bool read_tensor_from_file(const std::string& path, torch::Tensor& cpu_tensor) {
     return false;
   }
 
-  // Wrap buffer into a Torch tensor (CPU, pinned)
-  auto options = torch::TensorOptions()
-                     .dtype(torch::kUInt8)  // raw bytes
-                     .device(torch::kCPU)
-                     .pinned_memory(true);
-
   // Wrap staging buffer into a tensor without copying the data
   cpu_tensor = torch::from_blob(
       buf.ptr,
-      {static_cast<long>(file_size)},
+      {static_cast<int64_t>(file_size)},
       [p = buf.ptr](void* /*unused*/) {},
-      options);
+      STAGING_BUFFER_TENSOR_OPTIONS);
   return true;
 }
 


kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,117 @@
...
+    try {
+        fs::create_directories(parent_dir);
+    } catch (const fs::filesystem_error& e) {
+        std::cerr << "[ERROR] Failed to create directories: "
+                  << e.what() << "\n";
+        return false;
+    }
+
+    // Write to a temporary file to ensure atomic replace on rename
+    std::string tmp_path = target_path + ".tmp";
```
dannyharnik: If we have a race condition of two requests of the same content, they will both attempt to create the same file and also the exact same tmp file. This will fail the second open. 
Can solve this either by adding a random salt here to avoid such a collision.
Alternative is to understand it is already open and assume that the other writer will do the job...

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,117 @@
...
+    // Acquire pinned buffer of the required size
+    auto [pinned_ptr, pinned_size] = get_thread_local_pinned(file_size);
+    if (!pinned_ptr || pinned_size < file_size) {
+        std::cerr << "[ERROR] Pinned buffer too small for file: " << path << "\n"
+                  << "[INFO] Required size: " << file_size << " bytes, Available size: " << pinned_size << " bytes\n"
+                  << "pinned_ptr: " << pinned_ptr << "\n";
+        throw std::runtime_error("Pinned buffer too small for file: " + path);
+    }
+
+    // Read file into pinned memory
```
dannyharnik: Did you try Posix read() instead? AI assistant claims its faster... 
Not critical for this version

dannyharnik (changes requested):
Looks good but need to ensure we don't hit the race condition I pointed out

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: From my earlier tests, it was slower, and `io_uring` gave the same results, but it is totally something we can improve in the next version

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: From my earlier tests, it was slower, and io_uring gave the same results, but it is totally something we can improve in the next version

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,133 @@
...
+#include "file_io.hpp"
+#include "buffer.hpp"
+
+namespace fs = std::filesystem;
+
+// Thread-local index for CUDA streams
+extern thread_local size_t thread_stream_idx;
+
+// Write a tensor to disk using a temporary file and atomic rename
+bool write_file_to_disk(const std::string& target_path, const torch::Tensor& host_buf) {
```
orozery: nit:
I think `tensor_to_file` is more intuitive.
Also, I would make `tensor` the first argument, and `target_path` the second.

Also, it would be good to add these assertions:
```
assert(tensor.is_contiguous());
assert(tensor.device().is_cpu());
```

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,133 @@
...
+    try {
+        fs::create_directories(parent_dir);
+    } catch (const fs::filesystem_error& e) {
+        std::cerr << "[ERROR] Failed to create directories: " << e.what() << "\n";
+        return false;
+    }
+
+    // Write to a temporary file to ensure atomic replace on rename
+    // Include thread_stream_idx so each thread uses a unique temporary file
+    std::string tmp_path = target_path + std::to_string(thread_stream_idx) + ".tmp.";
```
orozery: What if you have multiple vllm instances writing to the same path?
I think it's thus safer to generate a temporary file name instead of `target_path + std::to_string(thread_stream_idx) + ".tmp."`.

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+    // Define a larger buffer (1MB) to reduce syscall overhead and speed up I/O
+    const size_t WRITE_BUFFER_SIZE = 1 * 1024 * 1024;  // 1MB buffer
```
orozery: nit: can you move this just before the buffer allocation?
Actually, maybe even put it outside the function at the beginning of the file?

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,133 @@
...
+    // Allocate custom I/O buffer for this stream (replaces small default buffer)
+    std::vector<char> buffer(WRITE_BUFFER_SIZE);
+    // Apply the custom buffer to the file stream
+    ofs.rdbuf()->pubsetbuf(buffer.data(), WRITE_BUFFER_SIZE);
+
+    // Write file contents
+    ofs.write(reinterpret_cast<const char*>(data_ptr), nbytes);
+    if (!ofs) {
+        std::cerr << "[ERROR] Failed to write to temporary file: " << tmp_path << " - " << std::strerror(errno) << "\n";
+        return false;
```
orozery: Add `ofs.close()`.
Maybe also delete the file? @dannyharnik what do you think?

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,133 @@
...
+    }
+    ofs.close();
+
+    // Atomically rename temp file to final target name after successful write
```
orozery:
```suggestion
    // Atomically rename temp file to final target name after a successful write
```

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,133 @@
...
+    if (std::rename(tmp_path.c_str(), target_path.c_str()) != 0) {
+        std::cerr << "[ERROR] "
+                  << "Failed to rename " + tmp_path + " to " + target_path + " - " + std::strerror(errno) << "\n";
+        return false;
```
orozery: @dannyharnik do we want to cleanup the temporary file here?

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,133 @@
...
+    const size_t WRITE_BUFFER_SIZE = 1 * 1024 * 1024;  // 1MB buffer
+
+    std::ofstream ofs(tmp_path, std::ios::out | std::ios::binary);
+    if (!ofs) {
+        std::cerr << "[ERROR] Failed to open temporary file for writing: " << tmp_path << " - " << std::strerror(errno) << "\n";
+        return false;
+    }
+
+    // Allocate custom I/O buffer for this stream (replaces small default buffer)
+    std::vector<char> buffer(WRITE_BUFFER_SIZE);
```
orozery: Can you try to avoid buffer allocation each call and instead use a thread-local pre-allocated buffer?
e.g.
```
const size_t WRITE_BUFFER_SIZE = 1 * 1024 * 1024; // 1MB buffer
thread_local std::vector<char> thread_write_buffer(WRITE_BUFFER_SIZE);
```

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,133 @@
...
+        std::cerr << "[ERROR] "
+                  << "Failed to rename " + tmp_path + " to " + target_path + " - " + std::strerror(errno) << "\n";
+        return false;
+    }
+
+    return true;
+}
+
+// Read a file into a pinned CPU tensor using the thread-local pinned buffer
+torch::Tensor read_file_from_disk(const std::string& path) {
```
orozery: nit: I would call this function `read_tensor_from_file`.
Also, I would prefer this function to have a second argument `torch::Tensor& out_tensor`, and return a bool error (like your other function above).

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,133 @@
...
+        std::cerr << "[ERROR] Failed to open file: " << path << "\n";
+        throw std::runtime_error("Failed to open file: " + path);
+    }
+
+    // Determine file size
+    size_t file_size = static_cast<size_t>(ifs.tellg());
+    ifs.seekg(0, std::ios::beg);
+
+    // Acquire pinned buffer of the required size
+    auto [pinned_ptr, pinned_size] = get_thread_local_pinned(file_size);
```
orozery: where is `get_thread_local_pinned` defined?

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,133 @@
...
+    if (!pinned_ptr || pinned_size < file_size) {
+        std::cerr << "[ERROR] Pinned buffer too small for file: " << path << "\n"
+                  << "[INFO] Required size: " << file_size << " bytes, Available size: " << pinned_size << " bytes\n"
+                  << "pinned_ptr: " << pinned_ptr << "\n";
+        throw std::runtime_error("Pinned buffer too small for file: " + path);
+    }
+
+    // Read file into pinned memory
+    ifs.read(reinterpret_cast<char*>(pinned_ptr), file_size);
```
orozery: What about setting a (1MB?) read buffer like you did on the other function?
e.g.
```
ifs.rdbuf()->pubsetbuf(...);
```

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,133 @@
...
+    ifs.close();
+
+    // Wrap pinned buffer into a Torch tensor (CPU, pinned)
+    auto options = torch::TensorOptions()
+                       .dtype(torch::kUInt8)  // raw bytes
+                       .device(torch::kCPU)
+                       .pinned_memory(true);
```
orozery: Keep in mind that not all platforms support memory pinning (e.g. TPUs).
For CPU offloading we only use pinning if available.
I think it is fine to keep it hard-coded for now, as we only support GPUs.
However, I would remove all referenced to `pinned` (e.g. `get_thread_local_pinned` -> `get_thread_local_staging_buffer`, which BTW I think is more self-explaining).

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,133 @@
...
+        [pinned_ptr](void* /*unused*/) {},
+        options);
+    return tensor;
+}
+
+// update_atime update only the atime of a file without changing mtime
+void update_atime(const std::string& path) {
+    struct timespec times[2];
+    times[0].tv_nsec = UTIME_OMIT;  // keep mtime unchanged
+    times[1].tv_nsec = UTIME_NOW;   // update atime to now
```
orozery:
```suggestion
    times[0].tv_nsec = UTIME_NOW;    // update atime to now
    times[1].tv_nsec = UTIME_OMIT;   // keep mtime unchanged
```

orozery (changes requested):
Good job!

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
dannyharnik: Yes, you are probably right. Maybe will handle a case in which it runs out of space in the middle and still creates the file...

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
dannyharnik: The stream_idx was added to avoid such a case (assuming it is unique). 
Could use a random string as well...

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
dannyharnik: yep

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: done, but I will keep the write and read so it will be `write_tensor_to_file`

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: in buffer.cpp https://github.com/llm-d/llm-d-kv-cache-manager/pull/177/files#:~:text=std%3A%3Apair%3Cvoid*%2C%20size_t%3E-,get_thread_local_pinned(,-size_t%20required_bytes%2C%20int%20numa_node)%20%7B
I split it just to make the review easier

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: For `ifstream`, the data is read in one large call directly into the pinned staging buffer, so there is no need for an extra buffer.
For `ofstream`, the internal buffer is small and leads to many write syscalls, so we use a 1MB thread-local buffer to speed up writes.
In the next version, I plan to check more deeply whether `fstream` gives the best read and write performance.

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: I will change it to random to be on the safe side.

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,151 @@
...
+// -------------------------------------------------------------------
+// Constants and thread-local buffers
+// -------------------------------------------------------------------
+// Define a larger buffer (1MB) to reduce syscall overhead and speed up I/O
+const size_t WRITE_BUFFER_SIZE = 1 * 1024 * 1024;  // 1MB buffer
+// Allocate custom I/O buffer for this thread (replaces small default buffer)
+thread_local std::vector<char> thread_write_buffer(WRITE_BUFFER_SIZE);
+// Thread-local unique id for temporary files
+thread_local uint64_t thread_unique_id =
+    std::hash<std::thread::id>{}(std::this_thread::get_id());
```
orozery: This won't help you if you have multiple process or even multiple nodes.
How about:
```
#include <random>
thread_local uint64_t thread_unique_id = std::random_device{}();
```

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+// -------------------------------------------------------------------
+// file-IO Functions
+// -------------------------------------------------------------------
+// Write a tensor to disk using a temporary file and atomic rename
+bool write_tensor_to_file(const torch::Tensor& host_buf,
```
orozery: I think `host_buf` is confusing. 
This is not a buffer, but actually a tensor.
So maybe rename it to `tensor` or `cpu_tensor`?
I would also add an assert that it is a CPU tensor:
```
TORCH_CHECK(cpu_tensor.is_cpu(), "Tensor must be on CPU");
```

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,151 @@
...
+                          const std::string& target_path) {
+    // Pointer and size of data to write
+    const void* data_ptr = host_buf.data_ptr();
+    size_t nbytes = host_buf.nbytes();
+
+    // Create parent directory if needed
+    fs::path file_path(target_path);
+    fs::path parent_dir = file_path.parent_path();
+    try {
+        fs::create_directories(parent_dir);
```
orozery: This could fail if two threads are trying to create the same directories at the same time:
https://github.com/boostorg/filesystem/blob/b6130f0f17c248a57e9a3f52f08430aa3b3b99b0/src/operations.cpp#L3364

Another point is that I'm still afraid this `fs::create_directories` can be pricey, even if the directories exist.
My suggestion is that we implement our own create_directories function, which will:
1. Be thread-safe
2. Cache which directories already exists

Anyhow, I'm good with opening an issue and making these changes in a follow-up.
@dannyharnik WDYT?

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,151 @@
...
+thread_local uint64_t thread_unique_id =
+    std::hash<std::thread::id>{}(std::this_thread::get_id());
+
+// -------------------------------------------------------------------
+// file-IO Functions
+// -------------------------------------------------------------------
+// Write a tensor to disk using a temporary file and atomic rename
+bool write_tensor_to_file(const torch::Tensor& host_buf,
+                          const std::string& target_path) {
+    // Pointer and size of data to write
```
orozery: Can you add?
```
TORCH_CHECK(host_buf.is_contiguous(), "Tensor must be contiguous");
```

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,151 @@
...
+        target_path + std::to_string(thread_unique_id) + ".tmp.";
+
+    std::ofstream ofs(tmp_path, std::ios::out | std::ios::binary);
+    if (!ofs) {
+        std::cerr << "[ERROR] Failed to open temporary file for writing: "
+                  << tmp_path << " - " << std::strerror(errno) << "\n";
+        return false;
+    }
+
+    std::vector<char> buffer(WRITE_BUFFER_SIZE);
```
orozery:
```suggestion
```

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,151 @@
...
+    // Apply the custom buffer to the file stream
+    ofs.rdbuf()->pubsetbuf(thread_write_buffer.data(), WRITE_BUFFER_SIZE);
+
+    // Write file contents
+    ofs.write(reinterpret_cast<const char*>(data_ptr), nbytes);
```
orozery: Add a flush after the write:
```
ofs.flush();
```

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,151 @@
...
+bool read_tensor_from_file(const std::string& path, torch::Tensor& host_buf) {
+    // Open file
+    std::ifstream ifs(path, std::ios::in | std::ios::binary | std::ios::ate);
+    if (!ifs) {
+        std::cerr << "[ERROR] Failed to open file: " << path << "\n";
+        return false;
+    }
+
+    // Determine file size
+    size_t file_size = static_cast<size_t>(ifs.tellg());
```
orozery:
```suggestion
    std::ifstream::pos_type end_pos = ifs.tellg();
    if (end_pos < 0) {
        std::cerr << "[ERROR] Failed to determine file size: " << path << "\n";
        return false;
    }

    size_t file_size = static_cast<size_t>(end_pos);
```

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,151 @@
...
+        std::cerr << "[ERROR] Staging buffer too small for file: " << path
+                  << "\n"
+                  << "[INFO] Required size: " << file_size
+                  << " bytes, Available size: " << buf.size << " bytes\n"
+                  << "ptr: " << buf.ptr << "\n";
+        return false;
+    }
+
+    // Read file into Staging buffer
+    ifs.read(reinterpret_cast<char*>(buf.ptr), file_size);
```
orozery:
```suggestion
    ifs.read(reinterpret_cast<char*>(buf.ptr),
             static_cast<std::streamsize>(file_size));
    std::streamsize bytes_read = ifs.gcount();
    if (bytes_read != static_cast<std::streamsize>(file_size) || !ifs.good()) {
        std::cerr << "[ERROR] Failed to read full file: " << path
                  << " (read " << bytes_read << "/" << file_size << " bytes)\n";
        return false;
    }
```

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,151 @@
...
+    std::vector<char> buffer(WRITE_BUFFER_SIZE);
+    // Apply the custom buffer to the file stream
+    ofs.rdbuf()->pubsetbuf(thread_write_buffer.data(), WRITE_BUFFER_SIZE);
+
+    // Write file contents
+    ofs.write(reinterpret_cast<const char*>(data_ptr), nbytes);
+    if (!ofs) {
+        std::cerr << "[ERROR] Failed to write to temporary file: " << tmp_path
+                  << " - " << std::strerror(errno) << "\n";
+        ofs.close();
```
orozery: apparently `ofs.close();` is unnecessary and even discouraged, as it is automatically called upon destruction.
Let's remove all `close()` calls throughout this file.

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,151 @@
...
+    ifs.close();
+
+    // Wrap buffer into a Torch tensor (CPU, pinned)
+    auto options = torch::TensorOptions()
+                       .dtype(torch::kUInt8)  // raw bytes
+                       .device(torch::kCPU)
+                       .pinned_memory(true);
+
+    // Wrap staging buffer into a tensor without copying the data
+    host_buf = torch::from_blob(
```
orozery: nit: rename `host_buf` to `cpu_tensor` ?

orozery (changes requested):

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: I get your point. I think it's cleaner if I explicitly close it - I don't see how it can impact the performance, if I call it explicitly

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: flush() after every write can hurt performance, and I am okay with the data not being written to disk immediately, so I would prefer to skip it.

orozery: You have a single write per file, so you have to flush anyhow.
This assures the file was actually written.

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
orozery: It's not performance, it's just complicates the code.
Notice you call ofs.close() on each error path.
It's hard to maintain this. So why bother?

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: I think it will get the same behavior, because the thread ID is unique, but I will change it.

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: But it is similar to close(), the destructor will call it when you do close()

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: I will open an issue on that

kfirtoledo:
@orozery PTAL

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
orozery: But if it's flushed on close you don't check it
If you flush here, you will correctly return false and log an error.

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,159 @@
...
+    fs::create_directories(parent_dir);
+  } catch (const fs::filesystem_error& e) {
+    std::cerr << "[ERROR] Failed to create directories: " << e.what() << "\n";
+    return false;
+  }
+
+  // Write to a temporary file to ensure atomic replace on rename
+  // Include thread_unique_id so each thread uses a unique temporary file
+  std::string tmp_path =
+      target_path + std::to_string(thread_unique_id) + ".tmp.";
```
orozery: can we define a `thread_local const std::string temp_file_name_suffix` that will replace `thread_unique_id`?
Also, drop the "." at the end of of ".tmp.".

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+  std::ofstream ofs(tmp_path, std::ios::out | std::ios::binary);
+  if (!ofs) {
+    std::cerr << "[ERROR] Failed to open temporary file for writing: "
```
orozery, line 75: can we add a log_error, log_info, etc. kind util functions that will automatically add the `[LEVEL] ` prefix?

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,159 @@
...
+bool read_tensor_from_file(const std::string& path, torch::Tensor& cpu_tensor) {
+  // Open file
+  std::ifstream ifs(path, std::ios::in | std::ios::binary | std::ios::ate);
+  if (!ifs) {
+    std::cerr << "[ERROR] Failed to open file: " << path << "\n";
+    return false;
+  }
+
+  // Determine file size
+  ifs.seekg(0, std::ios::end);  // Move read pointer to END
```
orozery: this line can be removed as the file is opened with `std::ios::ate`.

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,159 @@
...
+    std::cerr << "[ERROR] Failed to read full file: " << path << " (read "
+              << bytes_read << "/" << file_size << " bytes)\n";
+    return false;
+  }
+
+  // Wrap buffer into a Torch tensor (CPU, pinned)
+  auto options = torch::TensorOptions()
+                     .dtype(torch::kUInt8)  // raw bytes
+                     .device(torch::kCPU)
+                     .pinned_memory(true);
```
orozery: Can you move this to a global const?
```
// Global options for CPU pinned uint8 tensors (reusable, avoids recreation)
static const auto STAGING_BUFFER_TENSOR_OPTIONS = torch::TensorOptions()
    .dtype(torch::kUInt8)
    .device(torch::kCPU)
    .pinned_memory(true);
```

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+  // Wrap staging buffer into a tensor without copying the data
+  cpu_tensor = torch::from_blob(
+      buf.ptr,
+      {static_cast<long>(file_size)},
```
orozery:
```suggestion
      {static_cast<int64_t>(file_size)},
```

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: I will add a proper logger in the next version

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/file_io.cpp
kfirtoledo: Done

orozery (approved):
Thanks @kfirtoledo !

dannyharnik (approved):
Good job!

PR #177, opened 2025-11-23
[fs_connector][feat]: Add NUMA management module

kfirtoledo:
This PR adds the buffer management module used by the filesystem offloading connector.
Included in this PR:

- Adds buffer.cpp and buffer.hpp
- Implements pinned memory allocation and release logic
- Ensures proper alignment for fast GPU transfers
- Provides reusable buffers for read and write operations


kv_connectors/llmd_fs_backend/src/csrc/storage/buffer.cpp
```diff
@@ -0,0 +1,133 @@
...
+#include <thread>
+#include <vector>
+#include <algorithm>
+#include <sys/sysinfo.h>
+
+#include "buffer.hpp"
+#include "debug_utils.hpp"
+
+// Thread-local pinned buffer used by each IO thread
+thread_local PinnedBufferInfo t_pinned_buffer{};
```
orozery: nit: as in the previous PR, I think "pinned" should be removed from the names.
Maybe use "staging" instead.

kv_connectors/llmd_fs_backend/src/csrc/storage/buffer.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+// Global pinned buffers used by IO threads
+std::vector<PinnedBufferInfo> g_pinned_buffers;
+
+// Return thread-local pinned buffer, allocating or reallocating if needed
+std::pair<void*, size_t> get_thread_local_pinned(size_t required_bytes, int numa_node) {
```
orozery: Any reason why not return `PinnedBufferInfo` instead of `std::pair<void*, size_t>`?

orozery: I don't see `numa_node` being used. Remove it?

kv_connectors/llmd_fs_backend/src/csrc/storage/buffer.cpp
```diff
@@ -0,0 +1,133 @@
...
+    if (!t_pinned_buffer.ptr || t_pinned_buffer.size < required_bytes) {
+        if (t_pinned_buffer.ptr) {
```
orozery: Does `get_thread_local_pinned` get used other than by `preallocate_pinned_buffers`?
If not, it looks like the reallocating path is unnecessary.
Also, you won't need `get_thread_local_pinned` to access any thread local.
Then simply rename this function to `allocate_staging_buffer`.

BTW is just occurred to me:
You're using snake_case throughout your code.
I think camelCase is more common for C++, wheras snake_case is more for C code.

kv_connectors/llmd_fs_backend/src/csrc/storage/buffer.cpp
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    // Wait for all threads to complete initialization
+    for (auto& t : workers) t.join();
+
+    std::cout << "[INFO] Pre-allocated pinned buffer " << (alloc_bytes / (1024 * 1024)) << " MB for " << io_threads << " threads"
+              << std::endl;
+}
+
+// Return NUMA node associated with a given GPU
```
orozery: If we don't actually allocate base on numa, maybe we can remove all numa-related code?

kv_connectors/llmd_fs_backend/src/csrc/storage/buffer.cpp
```diff
@@ -0,0 +1,133 @@
...
+int get_gpu_numa_node(int device_id) {
+    CUdevice device;
+    int numa_node = -1;
+
+    cuInit(0);
+    cuDeviceGet(&device, device_id);
+    cuDeviceGetAttribute(&numa_node, CU_DEVICE_ATTRIBUTE_HOST_NUMA_ID, device);
```
orozery: missing error checks

kv_connectors/llmd_fs_backend/src/csrc/storage/buffer.cpp
```diff
@@ -0,0 +1,133 @@
...
+    return numa_node;
+}
+
+// Return list of CPU cores that belong to a NUMA node
+std::vector<int> get_cpus_in_numa_node(int node) {
```
orozery: This may be useful when debugging numa-affinity tweaks.
I don't think we're there yet. 
I think this should be removed.

kv_connectors/llmd_fs_backend/src/csrc/storage/buffer.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/buffer.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/buffer.cpp
kfirtoledo: Pytorch and all the code in VLLM use  snake_case, so I was consistent with them.
and Yes, get_thread_local_pinned use by others, for the safer side, if we need to read or write a larger file, it will reallocate the buffer again

kv_connectors/llmd_fs_backend/src/csrc/storage/buffer.cpp
kfirtoledo: I use this function, I allocate threads only to CPUs that are related to the NUMA node, and if it doesn't find the NUMA node it will allocate to all CPUs

kv_connectors/llmd_fs_backend/src/csrc/storage/buffer.cpp
kfirtoledo: If you prefer, I can add a flag to turn it off, but the feature is useful and is based on input from the storage team.

kv_connectors/llmd_fs_backend/src/csrc/storage/buffer.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/buffer.cpp
```diff
@@ -0,0 +1,150 @@
...
+            cudaFreeHost(t_staging_buffer.ptr);
+            t_staging_buffer.ptr = nullptr;
+            t_staging_buffer.size = 0;
+            std::cerr << "[WARN] Thread " << std::this_thread::get_id()
+                      << " existing staging buffer too small ("
+                      << t_staging_buffer.size << " bytes), reallocating "
+                      << required_bytes << " bytes\n";
+        }
+
+        size_t alloc_size = std::max(required_bytes, (size_t)16 * 1024 * 1024);
```
orozery: nit: can we declare a `MIN_STAGING_BUFFER_SIZE` (or another appropriate name) constant?

kv_connectors/llmd_fs_backend/src/csrc/storage/buffer.cpp
```diff
@@ -0,0 +1,150 @@
...
+                                         << " allocated staging buffer "
+                                         << (alloc_size / (1024 * 1024))
+                                         << " MB");
+        }
+    }
+    return t_staging_buffer;
+}
+
+// Preallocate thread-local staging buffers for all IO threads
+void preallocate_staging_buffers(size_t io_threads, size_t buffer_size_mb) {
```
orozery: Why do we need this function?
I think each worker thread in `thread_pool.cpp` should just allocate its buffer before starting its main loop.

kv_connectors/llmd_fs_backend/src/csrc/storage/numa_utils.hpp
```diff
@@ -0,0 +1,43 @@
...
+void preallocate_staging_buffers(size_t io_threads, size_t buffer_size_mb);
+
+// Return thread-local staging buffer, allocating or reallocating if needed
+StagingBufferInfo get_thread_local_staging_buffer(size_t required_bytes);
+
+// Return NUMA node that a GPU belongs to
+int get_gpu_numa_node(int device_id);
+
+// Return list of CPU cores in the given NUMA node
+std::vector<int> get_cpus_in_numa_node(int node);
```
orozery, line 25: I don't see how these functions relate to this file.
Can we move them to a separate file? (`numa_utils.cpp` or similar)

kv_connectors/llmd_fs_backend/src/csrc/storage/buffer.cpp
```diff
@@ -0,0 +1,150 @@
...
+                "Failed to query NUMA node for GPU ",
+                device_id,
+                ": ",
+                cudaGetErrorString(err));
+
+    return numa_node;
+}
+
+// Return list of CPU cores that belong to a NUMA node
+std::vector<int> get_cpus_in_numa_node(int node) {
```
orozery: Perplexity said there are a couple of error paths that may cause a crash here.
He suggested:
```
std::vector<int> get_cpus_in_numa_node(int node) {
    if (node < 0) return {};
    
    std::string path = "/sys/devices/system/node/node" + std::to_string(node) + "/cpulist";
    std::ifstream f(path);
    if (!f.is_open()) {
        std::cerr << "[WARN] NUMA node " << node << " cpulist not found: " << path << std::endl;
        return {};
    }
    
    std::string line;
    if (!std::getline(f, line)) {
        std::cerr << "[WARN] NUMA node " << node << " cpulist empty: " << path << std::endl;
        return {};
    }
    
    std::vector<int> cpus;
    std::istringstream iss(line);
    std::string token;
    
    while (std::getline(iss, token, ',')) {
        // Trim whitespace
        token.erase(0, token.find_first_not_of(" \t"));
        token.erase(token.find_last_not_of(" \t") + 1);
        
        size_t dash = token.find('-');
        if (dash != std::string::npos) {
            try {
                std::string start_str = token.substr(0, dash);
                std::string end_str = token.substr(dash + 1);
                
                int start = std::stoi(start_str);
                int end = std::stoi(end_str);
                
                if (start >= 0 && end >= start && end < 4096) {
                    for (int c = start; c <= end; ++c) {
                        cpus.push_back(c);
                    }
                } else {
                    std::cerr << "[WARN] NUMA node " << node 
                              << " invalid range '" << token << "' (out of bounds)" << std::endl;
                }
            } catch (const std::exception& e) {
                std::cerr << "[WARN] NUMA node " << node 
                          << " malformed range '" << token << "': " << e.what() << std::endl;
                continue;
            }
        } else {
            try {
                int cpu = std::stoi(token);
                if (cpu >= 0 && cpu < 4096) {
                    cpus.push_back(cpu);
                } else {
                    std::cerr << "[WARN] NUMA node " << node 
                              << " invalid CPU '" << token << "' (out of bounds)" << std::endl;
                }
            } catch (const std::exception& e) {
                std::cerr << "[WARN] NUMA node " << node 
                          << " malformed CPU '" << token << "': " << e.what() << std::endl;
                continue;
            }
        }
    }
    return cpus;
}
```

kv_connectors/llmd_fs_backend/src/csrc/storage/numa_utils.hpp
orozery: Furthermore, I actually think we should move the buffer funcs from `buffer.cpp` to `thread_pool.cpp`, and remove `buffer.cpp`.

kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/buffer.cpp
kfirtoledo: We don't, I removed it in one of the preformance iteration, and I am not really sure why I brought it back

kv_connectors/llmd_fs_backend/src/csrc/storage/buffer.cpp
kfirtoledo: I added all the safety checks

PR #178, opened 2025-11-23
[fs_connector][feat]: Add multithreaded worker pool (thread_pool)

kfirtoledo:
This PR introduces the multithreaded worker pool (thread_pool)  used by the filesystem offloading connector.
Included in this PR:
- Adds thread_pool.cpp and thread_pool.hpp
- Implements a lightweight work queue and task execution model
- Enables concurrent read and write operations


kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+#include <cuda_runtime.h>
+
+#include "buffer.hpp"  // defines PinnedBuffer + get_gpu_numa_node + get_cpus_in_numa_node
+#include "debug_utils.hpp"
+
+// Thread-local storage used by each I/O thread
+extern thread_local size_t thread_stream_idx;
+
+// ThreadPool class
```
orozery: Can you add more comments on the purpose of this class?

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.hpp
```diff
@@ -0,0 +1,84 @@
...
+    std::mutex queue_mutex;             // Protects access to the task queue
+    std::condition_variable condition;  // Signals workers when tasks are available
+
+    std::atomic<bool> stop{false};  // Tells workers to stop and exit
+    int m_device_id;                // CUDA device this thread pool is bound to
+};
+
+// enqueue: submit a task to the thread pool
+template <class F>
+auto ThreadPool::enqueue(F&& f) -> std::future<std::invoke_result_t<F>> {
```
orozery, line 75: Why is this function implemented in the header file (`thread_pool.hpp` instead of `thread_pool.cpp`)?

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.hpp
```diff
@@ -0,0 +1,84 @@
...
+    auto task = std::make_shared<std::packaged_task<return_type()>>(std::forward<F>(f));
+
+    // Future for the caller to wait on
+    std::future<return_type> res = task->get_future();
+
+    {
+        std::unique_lock<std::mutex> lock(queue_mutex);
+
+        // Reject new tasks if the pool is shutting down
+        if (stop) throw std::runtime_error("enqueue on stopped ThreadPool");
```
orozery: Is there a way to reject without throwing an exception?

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+                    // Exit thread if pool is stopping and no tasks remain
+                    if (stop && tasks.empty()) return;
+
+                    // Fetch next task from the queue
+                    task = std::move(tasks.front());
+                    tasks.pop();
+                }
+                // Execute the task
+                task();
```
orozery: maybe wrap this with an exception catcher?

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,135 @@
...
+                                     << ") pinned to CPU " << cpu_id << " (running on CPU " << actual_cpu << ")");
+
+            // Attach preallocated pinned buffer for this thread
+            if (i < g_pinned_buffers.size() && g_pinned_buffers[i].ptr != nullptr) {
+                t_pinned_buffer.ptr = g_pinned_buffers[i].ptr;
+                t_pinned_buffer.size = g_pinned_buffers[i].size;
+                DEBUG_PRINT("IO thread " << i << " attached to preallocated pinned buffer " << (t_pinned_buffer.size / (1024 * 1024))
+                                         << " MB");
+            } else {
+                std::cerr << "[WARN] IO thread " << i << " has no preallocated pinned buffer\n";
```
orozery: Is warning enough? shouldn't we fail?
Also, if any thread crashes, should not the entire pool terminate?

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,135 @@
...
+            CPU_SET(cpu_id, &cpuset);
+
+            if (pthread_setaffinity_np(pthread_self(), sizeof(cpuset), &cpuset) != 0) {
+                std::cerr << "[ERROR] Failed to set affinity for thread " << i << " to CPU " << cpu_id << "\n";
+            }
+
+            int actual_cpu = sched_getcpu();
+            pid_t tid = static_cast<pid_t>(syscall(SYS_gettid));
+            DEBUG_PRINT("IO thread " << i << " set CUDA device to " << device_id << " (tid=" << tid << ", tp_rank=" << tp_rank
+                                     << ") pinned to CPU " << cpu_id << " (running on CPU " << actual_cpu << ")");
```
orozery: Remove?

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,135 @@
...
+    std::cout << "\n";
+
+    // Create all worker threads
+    for (size_t i = 0; i < threads; ++i) {
+        // Launch a new worker thread with a lambda that initializes thread resources and processes queued tasks.
+        workers.emplace_back([this, i, threads, pinned_buffer_mb, tp_rank, device_id, gpu_numa, local_cpus] {
+            cudaSetDevice(device_id);
+
+            // Round-robin CPUs within the NUMA node
+            int cpu_id = local_cpus[i % local_cpus.size()];
```
orozery: Can this cause performance issues if threads are not evenly split across CPUs?
@dannyharnik

orozery (commented):
Nice work!

General comment:
I see very long lines throughout, which is hard to read and looks not so good :)
I think we need to add some kind of a style linter to enforce on all our codebase.

kfirtoledo:
> General comment: I see very long lines throughout, which is hard to read and looks not so good :) I think we need to add some kind of a style linter to enforce on all our codebase.

I added a linter in this PR "https://github.com/llm-d/llm-d-kv-cache-manager/commit/268dcf1bdf3445d4db967dbc33c0f8ccb9f04e35"

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.hpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.hpp
kfirtoledo: `enqueue` must be a template because it needs to accept arbitrary callables with arbitrary return types.
Since templates are instantiated only when used, the full implementation must be visible to every translation unit.
If the definition is placed in a .cpp file, the compiler cannot instantiate the template for other .cpp files, which leads to linker errors (undefined reference).

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
kfirtoledo: This code spreads threads across the NUMA-local CPUs using round-robin assignment (and falls back to all CPUs if none are found). It should keep load balanced, but I can add a TODO to evaluate the impact more deeply and consider relaxing the NUMA restriction if needed.

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
kfirtoledo: It shouldn’t fail, because a missing preallocated buffer only affects performance, not correctness. Other threads may still have buffers, and this thread will allocate its own buffer on first use. I updated the warning comment to reflect that behavior.

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.hpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+// ThreadPool constructor
+ThreadPool::ThreadPool(size_t threads,
+                       size_t staging_buffer_mb,
+                       int tp_rank,
+                       int device_id)
+    : m_device_id(device_id) {
+    // Initialize PyTorch threading globally (main thread only)
+    // at::init_num_threads();
+    // at::set_num_threads(1);
```
orozery: remove?

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+    // Get GPU NUMA node ONCE outside the thread loop
+    int gpu_numa = get_gpu_numa_node(device_id);
```
orozery: What happens if get_gpu_numa_node fails?
I think you need to skip the `get_cpus_in_numa_node` call.

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,170 @@
...
+                              tp_rank,
+                              device_id,
+                              gpu_numa,
+                              local_cpus] {
+            cudaSetDevice(device_id);
+
+            // Round-robin CPUs within the NUMA node
+            // TODO: Re-evaluate whether strict NUMA-based round-robin CPU
+            // assignment is optimal for performance.
+            int cpu_id = local_cpus[i % local_cpus.size()];
```
orozery: Maybe we should adjust the thread count based on len(local_cpus)?
Maybe just let the user set a maximum on the number of threads?

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,170 @@
...
+            if (i < g_staging_buffers.size() &&
+                g_staging_buffers[i].ptr != nullptr) {
+                t_staging_buffer.ptr = g_staging_buffers[i].ptr;
+                t_staging_buffer.size = g_staging_buffers[i].size;
+                DEBUG_PRINT("IO thread "
+                            << i << " attached to preallocated staging buffer "
+                            << (t_staging_buffer.size / (1024 * 1024))
+                            << " MB");
+            } else {
+                std::cerr << "[WARN] IO thread " << i
```
orozery: Recall my suggestion from the buffer PR. Why not allocate here?

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
kfirtoledo: The user decides the number of threads; the only thing here is that the pool is assigned to the CPUs assigned to the GPU NUMA node or the entire available CPU (fallback)

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
orozery: Does it make sense to allow the user to set a number of threads such that some cores get more threads scheduled than others?

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include <mutex>
+#include <queue>
+#include <condition_variable>
+#include <atomic>
+#include <sys/syscall.h>
+#include <unistd.h>
+#include <numa.h>
+
+#include "thread_pool.hpp"
+#include "buffer.hpp"
```
orozery: I saw you changed the buffer PR to numa_utils.
Aren't you moving the buffer parts over here?

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
kfirtoledo: Done

kfirtoledo:
@orozery PTAL

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,207 @@
...
+ThreadPool::ThreadPool(size_t threads,
+                       size_t staging_buffer_mb,
+                       int tp_rank,
+                       int device_id)
+    : m_device_id(device_id) {
+  // Get GPU NUMA node ONCE outside the thread loop
+  std::vector<int> local_cpus;
+  int gpu_numa = get_gpu_numa_node(device_id);
+  if (gpu_numa >= 0) {
+    std::cout << "[INFO] GPU " << device_id << " mapped to NUMA node "
```
orozery, line 58: As I suggested in another PR, we should move to a logging function.
This will also allow the user to set a logging level and get print to cout only the relevant messages.
Another thing is that we should use a mutex when outputting to cout or cerr since the messages could otherwise get interleaved by multiple threads.

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.hpp
```diff
@@ -0,0 +1,106 @@
...
+  std::mutex queue_mutex;  // Protects access to the task queue
+  std::condition_variable
+      condition;  // Signals workers when tasks are available
+
+  std::atomic<bool> stop{false};  // Tells workers to stop and exit
+  int m_device_id;                // CUDA device this thread pool is bound to
+  // Thread-local CUDA stream bound to this worker thread
+  static thread_local at::cuda::CUDAStream thread_stream;
+  // Thread-local buffer used by each IO thread
+  static thread_local StagingBufferInfo t_staging_buffer;
```
orozery: can we add `m_` prefix to variables?

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,207 @@
...
+                          staging_buffer_mb,
+                          tp_rank,
+                          device_id,
+                          gpu_numa,
+                          local_cpus] {
+      cudaSetDevice(device_id);
+
+      // Set thread-local CUDA stream for this worker thread
+      auto& tls_stream = ThreadPool::tls_stream();
+      cudaStreamSynchronize(tls_stream.stream());
```
orozery: You don't need this sync (as the stream is just created).

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
kfirtoledo: You are right, I will fix all the logging in the next version

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.hpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/thread_pool.cpp
kfirtoledo: I added this evaluation to my improvement PR https://github.com/llm-d/llm-d-kv-cache/issues/235

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,213 @@
...
+  static thread_local StagingBufferInfo m_staging_buffer;
+  if (!m_staging_buffer.ptr || m_staging_buffer.size < required_bytes) {
+    if (m_staging_buffer.ptr) {
+      cudaFreeHost(m_staging_buffer.ptr);
+      m_staging_buffer.ptr = nullptr;
+      m_staging_buffer.size = 0;
+      std::cerr << "[WARN] Thread " << std::this_thread::get_id()
+                << " existing staging buffer too small ("
+                << m_staging_buffer.size << " bytes), reallocating "
+                << required_bytes << " bytes\n";
```
orozery: move this before freeing

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,213 @@
...
+  m_condition.notify_all();
+  // Wait for all worker threads to exit
+  for (std::thread& worker : m_workers) {
+    worker.join();
+  }
+}
+// Get thread-local CUDA stream (initialization on first call)
+at::cuda::CUDAStream& ThreadPool::tls_stream() {
+  static thread_local at::cuda::CUDAStream stream =
+      at::cuda::getStreamFromPool(false);
```
orozery: there is already `m_thread_stream`, so we can remove this definition.

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,213 @@
...
+  return stream;
+}
+
+// Return thread-local staging buffer, allocating or reallocating if needed
+StagingBufferInfo& ThreadPool::tls_staging_buffer(size_t required_bytes) {
+  static thread_local StagingBufferInfo m_staging_buffer;
```
orozery: this is already defined in the header file. remove it from here.

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,213 @@
...
+    // Launch a new worker thread with a lambda that initializes thread
+    // resources and processes queued tasks.
+    m_workers.emplace_back([this,
+                            i,
+                            threads,
+                            staging_buffer_bytes,
+                            device_id,
+                            gpu_numa,
+                            local_cpus] {
+      cudaSetDevice(device_id);
```
orozery: can you check the return value of this to assert success?

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,213 @@
...
+          // Lock the task queue before checking it
+          std::unique_lock<std::mutex> lock(m_queue_mutex);
+
+          // Wait until either a new task arrives or the pool is
+          // stopping. (wait() unlocks the mutex while sleeping and
+          // re-locks it when waking)
+          m_condition.wait(lock, [this] { return m_stop || !m_tasks.empty(); });
+
+          // Exit thread if pool is stopping and no tasks remain
+          if (m_stop && m_tasks.empty()) {
```
orozery:
```suggestion
          if (m_stop) {
```

I think that we should stop immediately and not wait for complete waiting tasks
Also, the current logic is not thread safe once m_stop is True.

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,213 @@
...
+}
+// Get thread-local CUDA stream (initialization on first call)
+at::cuda::CUDAStream& ThreadPool::tls_stream() {
+  static thread_local at::cuda::CUDAStream stream =
+      at::cuda::getStreamFromPool(false);
+  return stream;
+}
+
+// Return thread-local staging buffer, allocating or reallocating if needed
+StagingBufferInfo& ThreadPool::tls_staging_buffer(size_t required_bytes) {
```
orozery: I would prefer to remove the reallocating logic and rename this function `allocate_staging_buffer`.

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
kfirtoledo: Done, I split into two functions: allocate_staging_buffer and get_staging_buffer

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
kfirtoledo: Done, but is static member that I need to define anyway

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
orozery: ?

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.hpp
```diff
@@ -0,0 +1,105 @@
...
+  std::queue<std::function<void()>> m_tasks;  // Queue of pending tasks
+
+  std::mutex m_queue_mutex;  // Protects access to the task queue
+  std::condition_variable
+      m_condition;  // Signals workers when tasks are available
+
+  std::atomic<bool> m_stop{false};  // Tells workers to stop and exit
+  int m_device_id;                  // CUDA device this thread pool is bound to
+  // Thread-local CUDA stream bound to this worker thread
+  static thread_local std::optional<at::cuda::CUDAStream> m_thread_stream;
```
orozery: Will this work?
```
static thread_local at::cuda::CUDAStream m_thread_stream =
    at::cuda::getStreamFromPool();
```

kfirtoledo: Done

orozery: ?

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
```diff
@@ -0,0 +1,217 @@
...
+  if (m_staging_buffer.ptr) {
+    if (m_staging_buffer.size >= required_bytes) {
+      return true;  // already allocated and big enough
+    }
+    std::cerr << "[ERROR] staging buffer already allocated ("
+              << m_staging_buffer.size << " bytes) but required "
+              << required_bytes << " bytes\n";
+    return false;
+  }
+
```
orozery: I still think we should remove this (dead code).

kfirtoledo: Done

PR #179, opened 2025-11-23
[fs_connector][feat]: Add tensor copy module (tensor_copy) for GPU and CPU block transfers

kfirtoledo:
This PR adds the tensor copy module used by the filesystem offloading connector.
Included in this PR:
- Adds `tensor_copy.cu` and `tensor_copy.hpp`
- Implements GPU and CPU block transfer routines for GET and PUT operations
- Provides CUDA copy kernels and CPU fallback copy paths
- Support kernel-based copy and standard memcpy by using the flags `USE_KERNEL_COPY_WRITE` and `USE_KERNEL_COPY_READ`

### DIFF
--- a/kv_connectors/llmd_fs_backend/csrc/storage/tensor_copier.cu
+++ b/kv_connectors/llmd_fs_backend/csrc/storage/tensor_copier.cu
@@ -21,36 +21,36 @@
 #include <vector>
 #include <iostream>
 
-#include "tensor_copy.hpp"
+#include "tensor_copier.hpp"
 #include "thread_pool.hpp"
 #include "debug_utils.hpp"
 #include <cstdlib>
 #include <string>
 
 // Constructor - initializes configuration
-TensorCopy::TensorCopy(std::vector<torch::Tensor>& tensors,
-                       int gpu_blocks_per_file)
+TensorCopier::TensorCopier(std::vector<torch::Tensor>& tensors,
+                           int gpu_blocks_per_file)
     : m_gpu_blocks_per_file(gpu_blocks_per_file), m_gpu_tensors(tensors) {
-  TORCH_CHECK(!m_gpu_tensors.empty(), "TensorCopy: tensors is empty");
+  TORCH_CHECK(!m_gpu_tensors.empty(), "TensorCopier: tensors is empty");
   TORCH_CHECK(m_gpu_blocks_per_file > 0,
-              "TensorCopy: gpu_blocks_per_file must be > 0");
+              "TensorCopier: gpu_blocks_per_file must be > 0");
   TORCH_CHECK(tensors[0].is_contiguous(), "GPU tensor must be contiguous");
 
   m_tensor_block_size = tensors[0].stride(0) * tensors[0].element_size();
   // Env flags
   m_use_kernel_copy_read = get_env_flag("USE_KERNEL_COPY_READ", false);
   m_use_kernel_copy_write = get_env_flag("USE_KERNEL_COPY_WRITE", false);
-  std::cout << "[INFO] TensorCopy: use_kernel_copy_read="
+  std::cout << "[INFO] TensorCopier: use_kernel_copy_read="
             << m_use_kernel_copy_read
             << ", use_kernel_copy_write=" << m_use_kernel_copy_write
             << ", m_gpu_blocks_per_file=" << m_gpu_blocks_per_file << std::endl;
 }
 
 // Performs block transfers using cudaMemcpyAsync (DMA-based copy)
-void TensorCopy::copy_blocks_via_cuda_memcpy(
+void TensorCopier::copy_blocks_via_cuda_memcpy(
     uint8_t* cpu_base,
     const std::vector<int64_t>& block_ids_list,
-    bool is_put) {
+    bool is_store) {
   uint8_t** src;
   uint8_t** dst;
   uint8_t* gpu_blk_ptr;
@@ -58,7 +58,7 @@ void TensorCopy::copy_blocks_via_cuda_memcpy(
   cudaMemcpyKind kind;
 
   // Determine source and destination based on direction
-  if (is_put) {
+  if (is_store) {
     kind = cudaMemcpyDeviceToHost;
     src = &gpu_blk_ptr;
     dst = &cpu_blk_ptr;
@@ -97,13 +97,13 @@ void TensorCopy::copy_blocks_via_cuda_memcpy(
 }
 
 // Main transfer function - dispatches to kernel or memcpy path
-void TensorCopy::copy_blocks(uint8_t* cpu_base,
-                             const std::vector<int64_t>& block_ids_list,
-                             bool is_put) {
-  bool use_kernel = is_put ? m_use_kernel_copy_write : m_use_kernel_copy_read;
+void TensorCopier::copy_blocks(uint8_t* cpu_base,
+                               const std::vector<int64_t>& block_ids_list,
+                               bool is_store) {
+  bool use_kernel = is_store ? m_use_kernel_copy_write : m_use_kernel_copy_read;
   if (use_kernel) {
-    copy_blocks_via_kernels(cpu_base, block_ids_list, is_put);
+    copy_blocks_via_kernels(cpu_base, block_ids_list, is_store);
   } else {
-    copy_blocks_via_cuda_memcpy(cpu_base, block_ids_list, is_put);
+    copy_blocks_via_cuda_memcpy(cpu_base, block_ids_list, is_store);
   }
 }
--- a/kv_connectors/llmd_fs_backend/csrc/storage/tensor_copier.hpp
+++ b/kv_connectors/llmd_fs_backend/csrc/storage/tensor_copier.hpp
@@ -20,14 +20,14 @@
 #include <vector>
 #include <cstdint>
 
-class TensorCopy {
+class TensorCopier {
  public:
-  TensorCopy(std::vector<torch::Tensor>& tensors, int gpu_blocks_per_files);
+  TensorCopier(std::vector<torch::Tensor>& tensors, int gpu_blocks_per_files);
 
   // Main transfer function - dispatches to kernel or memcpy path
   void copy_blocks(uint8_t* cpu_base,
                    const std::vector<int64_t>& block_ids_list,
-                   bool is_put);
+                   bool is_store);
 
  private:
   // GPU tensor list
@@ -44,10 +44,10 @@ class TensorCopy {
   // Performs block transfers using cudaMemcpyAsync (DMA-based copy)
   void copy_blocks_via_cuda_memcpy(uint8_t* cpu_base,
                                    const std::vector<int64_t>& block_ids_list,
-                                   bool is_put);
+                                   bool is_store);
 
   // Performs block transfers using a custom CUDA kernel
   void copy_blocks_via_kernels(uint8_t* cpu_base,
                                const std::vector<int64_t>& block_ids_list,
-                               bool is_put);
+                               bool is_store);
 };
--- a/kv_connectors/llmd_fs_backend/csrc/storage/tensor_copier_kernels.cu
+++ b/kv_connectors/llmd_fs_backend/csrc/storage/tensor_copier_kernels.cu
@@ -21,7 +21,7 @@
 #include <vector>
 #include <iostream>
 
-#include "tensor_copy.hpp"
+#include "tensor_copier.hpp"
 #include "thread_pool.hpp"
 #include "debug_utils.hpp"
 
@@ -60,7 +60,7 @@ __global__ void copy_blocks_kernel(
     const int layer_idx,                    // Current layer being processed
     const int num_layers,                   // Total number of layers
     const int gpu_blocks_per_file,  // Blocks per file for modulo calculation
-    const bool is_put)              // Direction flag (true=PUT, false=GET)
+    const bool is_store)            // Direction flag (true=PUT, false=GET)
 {
   const int bi = blockIdx.x;  // block index
   const int tid = threadIdx.x;
@@ -81,8 +81,8 @@ __global__ void copy_blocks_kernel(
 
   // Determine source and destination based on direction
   const uint8_t* src =
-      is_put ? (gpu_base + gpu_offset) : (cpu_base + cpu_offset);
-  uint8_t* dst = is_put ? (cpu_base + cpu_offset) : (gpu_base + gpu_offset);
+      is_store ? (gpu_base + gpu_offset) : (cpu_base + cpu_offset);
+  uint8_t* dst = is_store ? (cpu_base + cpu_offset) : (gpu_base + gpu_offset);
 
   // Copy cooperatively across threads
   for (size_t i = tid; i < tensor_block_size; i += blockDim.x) {
@@ -91,10 +91,10 @@ __global__ void copy_blocks_kernel(
 }
 
 // Performs block transfers using a custom CUDA kernel
-void TensorCopy::copy_blocks_via_kernels(
+void TensorCopier::copy_blocks_via_kernels(
     uint8_t* cpu_base,
     const std::vector<int64_t>& block_ids_list,
-    bool is_put) {
+    bool is_store) {
   const int num_layers = static_cast<int>(m_gpu_tensors.size());
 
   // Wrap block IDs in tensor and copy to GPU for kernel access
@@ -129,7 +129,7 @@ void TensorCopy::copy_blocks_via_kernels(
         layer,  // Pass current layer index
         num_layers,
         m_gpu_blocks_per_file,
-        is_put);
+        is_store);
 
     // Check for kernel launch errors immediately
     cudaError_t launch_err = cudaGetLastError();


kv_connectors/llmd_fs_backend/src/csrc/storage/tensor_copy.hpp
```diff
@@ -0,0 +1,47 @@
...
+    bool layers_before_blocks;
+    // Index of the num_blocks dimension within the tensor shape.
+    int block_axis;
+};
+
+extern ConnectorConfig g_connector_config;
+
+// Copy selected GPU blocks into a staging CPU buffer.
+// Returns a staging CPU tensor containing raw K/V block bytes.
+torch::Tensor copy_gpu_tensors_to_buffer(
```
orozery: nit: rename `buffer` to `cpu_tensor` ?
And in the function below as well

orozery: Can we return bool and pass cpu_tensor explicitly?

kv_connectors/llmd_fs_backend/src/csrc/storage/tensor_copy.hpp
```diff
@@ -0,0 +1,47 @@
...
+    const std::vector<torch::Tensor>& src_tensors,
+    const std::vector<int64_t>& block_ids_list,
+    const c10::cuda::CUDAStream& stream);
+
+// Copy data from a staging CPU buffer back into GPU tensors
+bool copy_buffer_to_gpu_tensors(torch::Tensor cpu_buf,
+                                const std::vector<int64_t>& block_ids_list,
+                                const std::vector<torch::Tensor>& dst_tensors,
+                                int num_blocks_in_file,
```
orozery: Let's remove `num_blocks_in_file` (with all of its occurrences).
Instead, before calling this function, slice the cpu_tensor to skip the required amount of elements (`cpu_buf.slice(0, num_elements_to_skip)`);

kv_connectors/llmd_fs_backend/src/csrc/storage/tensor_copy.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+#pragma once
+
+#include <torch/extension.h>
+#include <vector>
+#include <cstdint>
+
+// Layout metadata used by the connector to interpret KV-cache tensor shapes.
+struct ConnectorConfig {
+    // True if the KV plane (2) comes before the num_blocks dimension.
```
orozery:
```suggestion
    // True if the {k,v} dimension (2) comes before the num_blocks dimension.
```

kv_connectors/llmd_fs_backend/src/csrc/storage/tensor_copy.hpp
```diff
@@ -0,0 +1,47 @@
...
+    bool kv_before_blocks;
+    // True if the layer dimension appears before the num_blocks dimension.
+    bool layers_before_blocks;
+    // Index of the num_blocks dimension within the tensor shape.
+    int block_axis;
```
orozery: nit: I would rename this to `num_blocks_dim` or `num_blocks_dimension`

kv_connectors/llmd_fs_backend/src/csrc/storage/tensor_copy.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include <torch/extension.h>
+#include <vector>
+#include <cstdint>
+
+// Layout metadata used by the connector to interpret KV-cache tensor shapes.
+struct ConnectorConfig {
+    // True if the KV plane (2) comes before the num_blocks dimension.
+    bool kv_before_blocks;
+    // True if the layer dimension appears before the num_blocks dimension.
+    bool layers_before_blocks;
```
orozery: Why do you need this?

kv_connectors/llmd_fs_backend/src/csrc/storage/tensor_copy.hpp
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+#pragma once
+
+#include <torch/extension.h>
+#include <vector>
+#include <cstdint>
+
+// Layout metadata used by the connector to interpret KV-cache tensor shapes.
+struct ConnectorConfig {
```
orozery: Can we have a class that will hold the layout config, and the copy functions will be members of it?
Then we can remove `ConnectorConfig` with the global variable `g_connector_config`.
I would also put the cuda streams as members and you won't have to pass them to the copy functions.

kv_connectors/llmd_fs_backend/src/csrc/storage/tensor_copy.cu
```diff
@@ -0,0 +1,547 @@
...
+    TORCH_CHECK(cpu_buf.is_pinned(),
+                "cpu_buf must be pinned memory for kernel-based copy");
+
+    // Extract tensor geometry
+    auto layout =
+        CacheLayout::from_tensor(ref,
+                                 g_connector_config.block_axis,
+                                 g_connector_config.kv_before_blocks,
+                                 g_connector_config.layers_before_blocks);
+    const int num_layers = static_cast<int>(dst_tensors.size());
```
orozery: you are only serving a single model, so this calculation is redundant.
Layout should be set once in a constructor of a wrapping class.

kv_connectors/llmd_fs_backend/src/csrc/storage/tensor_copy.cu
```diff
@@ -0,0 +1,547 @@
...
+
+    auto* cpu_base = cpu_buf.data_ptr<uint8_t>();
+
+    // Check environment variable to determine copy method (default: kernel for
+    // READ)
+    bool use_kernel = get_env_flag("USE_KERNEL_COPY_READ", false);
```
orozery: I think we should remove the kernel implementation.
The code is complex as it is, and there's no use-case for that.

kv_connectors/llmd_fs_backend/src/csrc/storage/tensor_copy.cu
```diff
@@ -0,0 +1,547 @@
...
+    for (size_t i = tid; i < bytes_per_block; i += blockDim.x) {
+        dst[i] = src[i];
+    }
+}
+
+//----------------------------------------------------------------------
+// Copy Implementation Functions
+//----------------------------------------------------------------------
+// Standard cudaMemcpyAsync path (DMA-based copying)
+void copy_via_cuda_memcpy(uint8_t* cpu_base,
```
orozery: Can we change this function to have just 3 arguments?
`src_tensor`, `dst_tensor`, `block_ids_list`
And it will do this:
```

int64_t device_offset;
int64_t host_offset = 0;

if (check if src_tensor is gpu) {
   // device to host (GPU -> CPU)
   memcpy_type = cudaMemcpyDeviceToHost;
   src_offset = &device_offset;
   dst_offset = &host_offset;
   stream = this.d2h_stream;
} else {
   // host to device (CPU -> GPU)
   ...
}


char* src_ptr = static_cast<char*>(src.data_ptr());
char* dst_ptr = static_cast<char*>(dst.data_ptr());

for (const auto& block_id: block_ids_list) {
    device_offset = block_id * this.physical_block_size_in_bytes;
    cudaMemcpyAsync(dst_ptr + *dst_offset, src_ptr + *src_offset,
                    this.physical_block_size_in_bytes, memcpy_type, stream);
    host_offset += this.logical_block_size_in_bytes;
  }
```

where
```
this.logical_block_size_in_bytes = this.physical_block_size_in_bytes * (num_tensors) * (2 if kv before num_blocks else 1)
```

This will really help to simplify the code.

orozery (changes requested):

kv_connectors/llmd_fs_backend/src/csrc/storage/tensor_copy.hpp
kfirtoledo: Is part of the kv cache layout configuration, and I want to know if I can do a copy cross layout or one layout each time

kv_connectors/llmd_fs_backend/src/csrc/storage/tensor_copy.cu
kfirtoledo: At least for now, let's keep it, if you prefer I can move the 2 functions to diffrent file

kv_connectors/llmd_fs_backend/src/csrc/storage/tensor_copy.hpp
kfirtoledo: Done, I moved it tothe  cfg

kv_connectors/llmd_fs_backend/src/csrc/storage/tensor_copy.hpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/tensor_copy.cu
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/tensor_copy.cu
kfirtoledo: I am now using precomputed offsets via `CacheLayout::get_gpu_offset()` and `get_cpu_offset()` methods instead of inline computation, which handles our 3 different KV cache layouts (KV-first, Block-first, Cross-layer). This is similar to NIXL's approach, but instead of returning lists of block descriptor IDs, I return precomputed offsets, making the block copy loop simple and branch-free

kfirtoledo:
@orozery, I updated the cfg that will be part of TensorCpy class. Please take another look

kv_connectors/llmd_fs_backend/csrc/storage/tensor_copy.cu
```diff
@@ -0,0 +1,113 @@
...
+// Performs block transfers using cudaMemcpyAsync (DMA-based copy)
+void TensorCopy::copy_blocks_via_cuda_memcpy(
+    uint8_t* cpu_base,
+    const std::vector<int64_t>& block_ids_list,
+    const c10::cuda::CUDAStream& stream,
+    bool is_put) {
+  uint8_t** src;
+  uint8_t** dst;
+  uint8_t* gpu_layer_offset;
+  uint8_t* cpu_layer_offset;
```
orozery: these are actually gpu_ptr and cpu_ptr (not offsets)

kv_connectors/llmd_fs_backend/csrc/storage/tensor_copy.cu
```diff
@@ -0,0 +1,113 @@
...
+
+  for (size_t bi = 0; bi < block_ids_list.size(); ++bi) {
+    int64_t gpu_block_idx = block_ids_list[bi];
+
+    // Compute CPU block offset
+    // Each block in CPU memory stores all layers sequentially:
+    // [layer0_data, layer1_data, ..., layerN_data]
+    uint8_t* cpu_block_offset =
+        cpu_base + (gpu_block_idx % m_gpu_blocks_per_file) *
+                       m_gpu_tensors.size() * m_tensor_block_size;
```
orozery: you can just advance `cpu_block_offset += m_tensor_block_size` and avoid this assignment inside the outer loop.

kv_connectors/llmd_fs_backend/csrc/storage/tensor_copy.cu
```diff
@@ -0,0 +1,113 @@
...
+
+    // Process all layers for this block (for cross-layer layout is just one
+    // layer)
+    size_t li = 0;
+    for (const auto& tensor : m_gpu_tensors) {
+      gpu_layer_offset = reinterpret_cast<uint8_t*>(tensor.data_ptr()) +
+                         gpu_block_idx * m_tensor_block_size;
+      cpu_layer_offset = cpu_block_offset + li * m_tensor_block_size;
```
orozery: remove this, and instead advance `cpu_layer_offset += m_tensor_block_size` after the memcpy

kv_connectors/llmd_fs_backend/csrc/storage/tensor_copier.cu
```diff
@@ -0,0 +1,113 @@
...
+// Constructor - initializes configuration
+TensorCopy::TensorCopy(std::vector<torch::Tensor>& tensors,
+                       int gpu_blocks_per_file)
+    : m_gpu_blocks_per_file(gpu_blocks_per_file), m_gpu_tensors(tensors) {
+  TORCH_CHECK(!m_gpu_tensors.empty(), "TensorCopy: tensors is empty");
+  TORCH_CHECK(m_gpu_blocks_per_file > 0,
+              "TensorCopy: gpu_blocks_per_file must be > 0");
+  TORCH_CHECK(tensors[0].is_contiguous(), "GPU tensor must be contiguous");
+
+  m_tensor_block_size = tensors[0].stride(0) * tensors[0].element_size();
```
orozery, line 39: This assumes all tensors has the same block size.
This is OK for now, as we only support uniform models,
but maybe you want to assert that when we build the tensors list of kv_caches.

kv_connectors/llmd_fs_backend/csrc/storage/tensor_copy.cu
```diff
@@ -0,0 +1,113 @@
...
+      TORCH_CHECK(err == cudaSuccess, "cudaMemcpyAsync failed");
+      li++;
+    }
+  }
+}
+
+// Main transfer function - dispatches to kernel or memcpy path
+void TensorCopy::copy_blocks(uint8_t* cpu_base,
+                             const std::vector<int64_t>& block_ids_list,
+                             const c10::cuda::CUDAStream& stream,
```
orozery: This can be removed. you set the current stream outside.
inside use `at::cuda::getCurrentCUDAStream()`

kv_connectors/llmd_fs_backend/csrc/storage/tensor_copy_kernels.cu
```diff
@@ -0,0 +1,141 @@
...
+  uint8_t* dst = is_put ? (cpu_base + cpu_offset) : (gpu_base + gpu_offset);
+
+  // Copy cooperatively across threads
+  for (size_t i = tid; i < tensor_block_size; i += blockDim.x) {
+    dst[i] = src[i];
+  }
+}
+
+// Performs block transfers using a custom CUDA kernel
+void TensorCopy::copy_blocks_via_kernels(
```
orozery: I did not review this file as I don't think it's necessary :)

kv_connectors/llmd_fs_backend/csrc/storage/tensor_copy.cu
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  for (size_t bi = 0; bi < block_ids_list.size(); ++bi) {
+    int64_t gpu_block_idx = block_ids_list[bi];
+
+    // Compute CPU block offset
+    // Each block in CPU memory stores all layers sequentially:
+    // [layer0_data, layer1_data, ..., layerN_data]
+    uint8_t* cpu_block_offset =
+        cpu_base + (gpu_block_idx % m_gpu_blocks_per_file) *
```
orozery: can we read from offset of the file so that we don't read from file bytes we don't need?

kv_connectors/llmd_fs_backend/csrc/storage/tensor_copy_kernels.cu
kfirtoledo: Sure, let's keep it for now

kv_connectors/llmd_fs_backend/csrc/storage/tensor_copy.cu
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/tensor_copy.cu
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/tensor_copier.cu
kfirtoledo: Already have this assertion in the `_get_tensors` code

kv_connectors/llmd_fs_backend/csrc/storage/tensor_copy.cu
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/tensor_copy.cu
kfirtoledo: Ok, I will add this to the improvement PR-https://github.com/llm-d/llm-d-kv-cache/issues/235

kv_connectors/llmd_fs_backend/csrc/storage/tensor_copy.cu
kfirtoledo: Done

kfirtoledo:
@orozery,PTAL

orozery (approved):
Nice and neat! :)
Note I did not review the kernel copy.

PR #180, opened 2025-11-23
[fs_connector][feat]: Add storage_offload CUDA kernel and setup config

kfirtoledo:
This PR adds the main storage offload CUDA module for the filesystem offloading connector.
Included in this PR:
- Adds `storage_offload.cu`
- Implements the high-level CUDA logic that orchestrates block movement between GPU and storage
- Includes the `initialization` and `cleanup` routines for resources used by the module
- Adds `setup.py` configuration required to build and load the CUDA extension
- Adds `debug_utils.hpp`  for debug logging and timing measurement

### DIFF
--- a/kv_connectors/llmd_fs_backend/csrc/storage/file_io.cpp
+++ b/kv_connectors/llmd_fs_backend/csrc/storage/file_io.cpp
@@ -33,11 +33,7 @@ namespace fs = std::filesystem;
 // -------------------------------------------------------------------
 // Define a larger buffer (1MB) to reduce syscall overhead and speed up I/O
 const size_t WRITE_BUFFER_SIZE = 1 * 1024 * 1024;  // 1MB buffer
-// Staging buffer tensor options (CPU, pinned memory)
-static const auto STAGING_BUFFER_TENSOR_OPTIONS = torch::TensorOptions()
-                                                      .dtype(torch::kUInt8)
-                                                      .device(torch::kCPU)
-                                                      .pinned_memory(true);
+
 // Allocate custom I/O buffer for this thread (replaces small default buffer)
 thread_local std::vector<char> thread_write_buffer(WRITE_BUFFER_SIZE);
 
@@ -47,15 +43,9 @@ thread_local std::string tmp_file_suffix =
 // -------------------------------------------------------------------
 // file-IO Functions
 // -------------------------------------------------------------------
-// Write a tensor to disk using a temporary file and atomic rename
-bool write_tensor_to_file(const torch::Tensor& cpu_tensor,
+// Write a buffer to disk using a temporary file and atomic rename
+bool write_buffer_to_file(const StagingBufferInfo& buf,
                           const std::string& target_path) {
-  TORCH_CHECK(cpu_tensor.is_cpu(), "Tensor must be on CPU");
-  TORCH_CHECK(cpu_tensor.is_contiguous(), "Tensor must be contiguous");
-  // Pointer and size of data to write
-  const void* data_ptr = cpu_tensor.data_ptr();
-  size_t nbytes = cpu_tensor.nbytes();
-
   // Create parent directory if needed
   fs::path file_path(target_path);
   fs::path parent_dir = file_path.parent_path();
@@ -81,7 +71,7 @@ bool write_tensor_to_file(const torch::Tensor& cpu_tensor,
   ofs.rdbuf()->pubsetbuf(thread_write_buffer.data(), WRITE_BUFFER_SIZE);
 
   // Write file contents
-  ofs.write(reinterpret_cast<const char*>(data_ptr), nbytes);
+  ofs.write(reinterpret_cast<const char*>(buf.ptr), buf.size);
   if (!ofs) {
     std::cerr << "[ERROR] Failed to write to temporary file: " << tmp_path
               << " - " << std::strerror(errno) << "\n";
@@ -109,8 +99,8 @@ bool write_tensor_to_file(const torch::Tensor& cpu_tensor,
   return true;
 }
 
-// Read a file into a CPU tensor using the thread-local staging buffer
-bool read_tensor_from_file(const std::string& path, torch::Tensor& cpu_tensor) {
+// Read a file into a thread-local staging buffer
+bool read_buffer_from_file(const std::string& path, StagingBufferInfo& buf) {
   // Open file
   std::ifstream ifs(path, std::ios::in | std::ios::binary | std::ios::ate);
   if (!ifs) {
@@ -128,7 +118,6 @@ bool read_tensor_from_file(const std::string& path, torch::Tensor& cpu_tensor) {
   ifs.seekg(0, std::ios::beg);  // Move read pointer to start for reading
 
   // Acquire staging buffer of the required size
-  StagingBufferInfo& buf = ThreadPool::tls_staging_buffer(file_size);
   if (!buf.ptr || buf.size < file_size) {
     std::cerr << "[ERROR] Staging buffer too small for file: " << path << "\n"
               << "[INFO] Required size: " << file_size
@@ -147,12 +136,6 @@ bool read_tensor_from_file(const std::string& path, torch::Tensor& cpu_tensor) {
     return false;
   }
 
-  // Wrap staging buffer into a tensor without copying the data
-  cpu_tensor = torch::from_blob(
-      buf.ptr,
-      {static_cast<int64_t>(file_size)},
-      [p = buf.ptr](void* /*unused*/) {},
-      STAGING_BUFFER_TENSOR_OPTIONS);
   return true;
 }
 
--- a/kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
+++ b/kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
@@ -1,0 +1,321 @@
+/*
+ * Copyright 2025 The llm-d Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <torch/extension.h>
+#include <c10/cuda/CUDAStream.h>
+#include <ATen/cuda/CUDAContext.h>
+#include <c10/cuda/CUDAGuard.h>
+
+#include <iostream>
+#include <algorithm>
+#include <cassert>
+#include <fstream>
+#include <string>
+#include <vector>
+#include <chrono>
+#include <map>
+#include <future>
+#include <mutex>
+#include <thread>
+#include <queue>
+#include <condition_variable>
+#include <memory>
+#include <atomic>
+#include <optional>
+#include <sys/syscall.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <cstring>
+#include <filesystem>
+#include <numa.h>
+
+#include "storage_offload.hpp"
+#include "storage_types.hpp"
+#include "file_io.hpp"
+#include "numa_utils.hpp"
+#include "thread_pool.hpp"
+#include "debug_utils.hpp"
+#include "tensor_copier.hpp"
+
+// Initialize IO threads, CUDA streams, and staging memory pool
+StorageOffloadEngine::StorageOffloadEngine(int io_threads,
+                                           int gpu_blocks_per_file,
+                                           std::vector<torch::Tensor>& tensors)
+    : m_tensor_copier(tensors, gpu_blocks_per_file),
+      m_thread_pool(io_threads,
+                    calc_staging_bytes(gpu_blocks_per_file, tensors),
+                    get_device_id()) {}
+
+// Get current device (should be set by vLLM before calling this)
+int StorageOffloadEngine::get_device_id() {
+  int device_id = 0;
+  cudaError_t err = cudaGetDevice(&device_id);
+  if (err != cudaSuccess) {
+    std::cerr << "[ERROR] cudaGetDevice failed: " << cudaGetErrorString(err)
+              << "\n";
+  }
+  return device_id;
+}
+// Calculate staging buffer size in bytes
+size_t StorageOffloadEngine::calc_staging_bytes(
+    int gpu_blocks_per_file,
+    const std::vector<torch::Tensor>& tensors) {
+  size_t block_size_in_bytes = 0;
+  for (const auto& tensor : tensors) {
+    block_size_in_bytes += static_cast<size_t>(tensor.stride(0)) *
+                           static_cast<size_t>(tensor.element_size());
+  }
+  return block_size_in_bytes * static_cast<size_t>(gpu_blocks_per_file);
+}
+
+// -------------------------------
+// Status and job management
+// -------------------------------
+// Return finished jobs and their success status
+std::vector<std::pair<int, bool>> StorageOffloadEngine::get_finished() {
+  std::lock_guard<std::mutex> lock(m_jobs_mutex);
+
+  std::vector<std::pair<int, bool>> results;
+  std::vector<int> to_erase;
+
+  // Iterate over all active jobs.
+  for (auto& kv : m_jobs) {
+    int job_id = kv.first;
+    auto& job_state = kv.second;
+
+    // Check if the job has completed all its tasks.
+    if (job_state->completed_tasks.load() == job_state->total_tasks) {
+      bool all_ok = job_state->all_success.load();
+      results.emplace_back(job_id, all_ok);
+      to_erase.push_back(job_id);
+    }
+  }
+
+  // Remove all finished jobs from the map.
+  for (int jid : to_erase) {
+    m_jobs.erase(jid);
+  }
+  return results;
+}
+
+// Wait for all tasks in the specified job to complete
+void StorageOffloadEngine::wait_job(int job_id) {
+  std::vector<std::shared_future<bool>> futures;
+
+  {
+    std::lock_guard<std::mutex> lock(m_jobs_mutex);
+    auto it = m_jobs.find(job_id);
+    if (it == m_jobs.end()) return;
+    futures = it->second->futures;
+  }
+
+  for (auto& fut : futures) {
+    fut.wait();
+  }
+}
+// -------------------------------
+// Store and Load operations
+// -------------------------------
+// ScopeGuard: executes stored lambda on destruction.
+class ScopeGuard {
+  std::function<void()> func_;
+
+ public:
+  explicit ScopeGuard(std::function<void()> f) : func_(std::move(f)) {}
+  ~ScopeGuard() {
+    if (func_) func_();
+  }
+};
+
+// Async GPU -> Storage transfer
+bool StorageOffloadEngine::async_store_gpu_blocks(
+    int job_id,
+    std::vector<std::string> dst_files,
+    std::vector<std::vector<int64_t>> all_block_ids) {
+  // Create job state object that will track progress and futures for this
+  // job.
+  auto job_state = std::make_shared<JobState>();
+  job_state->total_tasks = dst_files.size();
+
+  // Record an event on the default stream to ensure all prior work is done
+  cudaEvent_t event;
+  cudaEventCreate(&event);
+  cudaEventRecord(event, at::cuda::getCurrentCUDAStream().stream());
+  // Wrap default_done event in shared_ptr with custom deleter
+  auto gpu_kvs_ready_event = std::shared_ptr<CUevent_st>(
+      event,
+      [](CUevent_st* evt) { cudaEventDestroy(evt); });
+
+  // For each dst_file file, enqueue one async task in the I/O thread pool.
+  for (size_t i = 0; i < dst_files.size(); i++) {
+    std::string dst_file = dst_files[i];
+    auto block_ids = all_block_ids[i];
+
+    auto future = m_thread_pool.enqueue(
+        [this, dst_file, block_ids, job_state, gpu_kvs_ready_event]() -> bool {
+          // Check if dst_file file already exists - skip write if it does
+          if (std::ifstream(dst_file).good()) {
+            update_atime(dst_file);
+            job_state->completed_tasks.fetch_add(1);
+            return true;  // File exists
+          }
+
+          // Wait for default stream to complete before starting the copy
+          auto& tls_stream = ThreadPool::get_tls_stream();
+          cudaStreamWaitEvent(tls_stream.stream(),
+                              gpu_kvs_ready_event.get(),
+                              0);
+
+          StagingBufferInfo& buf = ThreadPool::get_staging_buffer();
+          auto* cpu_base = static_cast<uint8_t*>(buf.ptr);
+          bool is_store = true;
+          bool success = false;
+
+          // Mark task completion on scope exit
+          ScopeGuard completion([&]() {
+            job_state->completed_tasks.fetch_add(1);
+            // if (!success) job_state->all_success = false; // TODO- silent
+            // ignore read failures for now offloading connector not able to
+            // handle failures
+          });
+
+          // Execute the copy operation
+          try {
+            // Stage 1: copy tensors from GPU to staging CPU tensor.
+            TIME_EXPR(
+                "write phase 1: copy_blocks ",
+                m_tensor_copier.copy_blocks(cpu_base, block_ids, is_store),
+                "file: ",
+                dst_file);
+
+            cudaError_t err = cudaStreamSynchronize(tls_stream.stream());
+            if (err != cudaSuccess) {
+              std::cerr << "[ERROR] cudaStreamSynchronize failed: "
+                        << cudaGetErrorString(err) << std::endl;
+              return false;
+            }
+            // Stage 2: Write the cpu tensor to disk.
+            success = TIME_EXPR("write phase 2: write_buffer_to_file",
+                                write_buffer_to_file(buf, dst_file),
+                                "file:",
+                                dst_file,
+                                " size:",
+                                buf.size);
+            if (!success) {
+              std::cerr << "[ERROR] Store failed during file write: "
+                        << dst_file << "\n";
+              return success;
+            }
+          } catch (const std::exception& e) {
+            std::cerr << "[ERROR] Store failed for " << dst_file << ": "
+                      << e.what() << std::endl;
+            success = false;
+          } catch (...) {
+            std::cerr << "[ERROR] Store failed for " << dst_file << "\n";
+            success = false;
+          }
+
+          return success;
+        });
+    // Convert std::future -> std::shared_future, which is copyable and can
+    // be waited on by multiple threads.
+    job_state->futures.push_back(future.share());
+  }
+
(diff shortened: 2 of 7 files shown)


kv_connectors/llmd_fs_backend/src/csrc/storage/storage_offload.cu
```diff
@@ -0,0 +1,362 @@
...
+          py::arg("tp_rank"),
+          py::arg("kv_before_blocks"),
+          py::arg("layers_before_blocks"),
+          py::arg("block_axis"));
+
+    m.def("cleanup_resources", &cleanup_resources);
+
+    m.def("get_finished", &get_finished);
+
+    m.def("transfer_async_put",
```
orozery: nit:
I think `async_save_gpu_blocks` and `async_load_gpu_blocks` are clearer names for `transfer_async_put` and `transfer_async_get`.

kv_connectors/llmd_fs_backend/src/csrc/storage/storage_offload.cu
```diff
@@ -0,0 +1,362 @@
...
+    std::lock_guard<std::mutex> lock(jobs_mutex);
+    jobs[job_id] = std::move(job_state);
+    return true;
+}
+
+// -------------------------------
+// PYBIND11 module
+// -------------------------------
+PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
+    m.def("init_resources",
```
orozery: I think this file is a good place to document each of the `m.def` functions.
In particular, note that `init_resources` must be called before using the transfer functions.

kv_connectors/llmd_fs_backend/copy_files.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
orozery: Remove this file?

kv_connectors/llmd_fs_backend/setup.py
```diff
@@ -0,0 +1,41 @@
...
+
+setup(
+    name="storage_offload",
+    packages=find_packages("src"),
+    package_dir={"": "src"},
+    ext_modules=[
+        CUDAExtension("storage_offload",
+                      sources=[
+                          "src/csrc/storage/storage_offload.cu",
+                          "src/csrc/storage/buffer.cpp",
```
orozery: rename to numa_utils?

kv_connectors/llmd_fs_backend/src/csrc/storage/storage_offload.cu
```diff
@@ -0,0 +1,416 @@
...
+namespace py = pybind11;
+
+// Tracks progress and results for a multi-file async PUT/GET job
+struct JobState {
+    // Futures for each async task in the job
+    std::vector<std::shared_future<bool>> futures;
+    // Number of tasks completed so far
+    std::atomic<int> completed_tasks{0};
+    // Total number of tasks scheduled for this job
+    std::atomic<int> total_tasks{0};
```
orozery: atomic is not needed here

kv_connectors/llmd_fs_backend/csrc/storage/file_io.cpp
```diff
@@ -48,14 +44,8 @@ thread_local std::string tmp_file_suffix =
 // file-IO Functions
 // -------------------------------------------------------------------
 // Write a tensor to disk using a temporary file and atomic rename
-bool write_tensor_to_file(const torch::Tensor& cpu_tensor,
+bool write_tensor_to_file(const StagingBufferInfo& buf,
```
orozery: you should search-replace "tensor" in file_io files and replace by "buffer"

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,298 @@
...
+    block_size_in_bytes += tensor.stride(0) * tensor.element_size();
+  }
+  auto staging_buffer_bytes = block_size_in_bytes * gpu_blocks_per_file;
+  std::cout << "[INFO] Initializing ThreadPool with " << io_threads
+            << " threads on device " << device_id << ", "
+            << staging_buffer_bytes / (1024 * 1024)
+            << " MB staging buffer per thread\n";
+
+  // Initialize TensorCopy class
+  m_tensor_copy = std::make_unique<TensorCopy>(tensors, gpu_blocks_per_file);
```
orozery: maybe TensorCopier, m_tensor_copier is more clear than m_tensor_copy, TensorCopy

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,298 @@
...
+#include "tensor_copy.hpp"
+
+// Initialize IO threads, CUDA streams, and staging memory pool
+StorageOffloadEngine::StorageOffloadEngine(
+    int io_threads,
+    int gpu_blocks_per_file,
+    std::vector<torch::Tensor>& tensors) {
+  // Get current device (should be set by vLLM before calling this)
+  int device_id;
+  cudaGetDevice(&device_id);
```
orozery: Check return?

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,298 @@
...
+  int gpu_numa = get_gpu_numa_node(device_id);
+  numa_set_preferred(gpu_numa);
+
+  // Pass device_id to thread pool
+  m_thread_pool =
+      std::make_unique<ThreadPool>(io_threads, staging_buffer_bytes, device_id);
+}
+
+// Destructor: stops worker threads and releases CUDA and memory resources.
+StorageOffloadEngine::~StorageOffloadEngine() { m_thread_pool.reset(); }
```
orozery: this is unnecessary

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+// StorageOffloadEngine class manages asynchronous storage offload operations
+class StorageOffloadEngine {
+ private:
+  // Mutex protecting access to the jobs map
+  std::mutex m_jobs_mutex;
+  // Global map of job_id to JobState, tracking async job progress
+  std::map<int, std::unique_ptr<JobState>> m_jobs;
+  // Thread pool for scheduling async PUT/GET tasks
+  std::unique_ptr<ThreadPool> m_thread_pool;
```
orozery: can we switch to `ThreadPool m_thread_pool;` ?

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.hpp
```diff
@@ -0,0 +1,76 @@
...
+  // Handles GPU <-> CPU tensor copy operations
+  std::unique_ptr<TensorCopy> m_tensor_copy;
```
orozery: same, can we hold the object directly, not via a pointer?

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,298 @@
...
+            job_state->completed_tasks.fetch_add(1);
+            return true;  // File exists
+          }
+          // Ensure correct device is set (thread-local)
+          int device_id;
+          cudaGetDevice(&device_id);
+
+          // Set thread to a dedicated CUDA stream for async task.
+          auto& tls_stream = ThreadPool::tls_stream();
+          at::cuda::setCurrentCUDAStream(tls_stream);
```
orozery: unnecessary

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,298 @@
...
+                      m_tensor_copy->copy_blocks(cpu_base,
+                                                 block_ids,
+                                                 tls_stream,
+                                                 is_put),
+                      "file: " + dst_file);
+
+            cudaError_t err = cudaStreamSynchronize(tls_stream.stream());
+            if (err != cudaSuccess) {
+              std::cerr << "[ERROR] cudaStreamSynchronize failed: "
+                        << cudaGetErrorString(err) << std::endl;
```
orozery, line 208: fail at this point?

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,298 @@
...
+  // For each source file, enqueue one async task in the I/O thread pool.
+  for (size_t i = 0; i < src_files.size(); i++) {
+    std::string src_file = src_files[i];
+    auto block_ids = all_block_ids[i];
+    auto future = m_thread_pool->enqueue(
+        [this, src_file, block_ids, job_state = job_state.get()]() -> bool {
+          // Set thread to a dedicated CUDA stream for async task.
+          auto& tls_stream = ThreadPool::tls_stream();
+          at::cuda::setCurrentCUDAStream(tls_stream);
+
```
orozery: unnecessary

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
orozery: However, you need to add a wait on the default stream.
Something like:
```
cudaEvent_t compute_event;
cudaEventCreate(&compute_event);
cudaEventRecord(compute_event, nullptr);
cudaStreamWaitEvent(tls_stream, compute_event, 0);
```

kv_connectors/llmd_fs_backend/copy_files.sh
kfirtoledo: Is not part of the commit- I removed it

kv_connectors/llmd_fs_backend/csrc/storage/file_io.cpp
kfirtoledo: Take a look at the latest commit- this was already done

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: I want to be sure that every task is done on the current stream

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.hpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.hpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/src/csrc/storage/storage_offload.cu
kfirtoledo: ok, but I changed it to async_store_gpu_blocks

kv_connectors/llmd_fs_backend/src/csrc/storage/storage_offload.cu
kfirtoledo: `completed_tasks` is updated from multiple worker threads and polled concurrently, so it needs to stay atomic.
`total_tasks` is set once before scheduling and can be non-atomic.

kv_connectors/llmd_fs_backend/src/csrc/storage/storage_offload.cu
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,311 @@
...
+    std::cerr << "[WARN] cudaSetDeviceFlags(cudaDeviceMapHost) failed: "
+              << cudaGetErrorString(flags_err) << "\n";
+  }
+
+  // Bind memory allocations in this thread to the NUMA node local to the GPU.
+  int device_id = get_device_id();
+  int gpu_numa = get_gpu_numa_node(device_id);
+  if (gpu_numa >= 0) {
+    numa_set_preferred(gpu_numa);
+  }
```
orozery: Shouldn't this be moved to TensorCopier?
I think that currently this code runs "too late".

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,311 @@
...
+    std::vector<std::vector<int64_t>> all_block_ids) {
+  // Create job state object that will track progress and futures for this
+  // job.
+  auto job_state = std::make_unique<JobState>();
+  job_state->total_tasks = dst_files.size();
+
+  // Record an event on the default stream to ensure all prior work is done
+  cudaEvent_t event;
+  cudaEventCreate(&event);
+  cudaEventRecord(event, at::cuda::getDefaultCUDAStream().stream());
```
orozery: need to change to `getDefaultCUDAStream` to `getCurrentCUDAStream`.

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,311 @@
...
+  // Wrap default_done event in shared_ptr with custom deleter
+  auto default_done = std::shared_ptr<CUevent_st>(event, [](CUevent_st* evt) {
```
orozery: can you rename `default_done` to `gpu_kvs_ready_event`?

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+  // For each dst_file file, enqueue one async task in the I/O thread pool.
+  for (size_t i = 0; i < dst_files.size(); i++) {
+    std::string dst_file = dst_files[i];
+    auto block_ids = all_block_ids[i];
+
+    auto future = m_thread_pool.enqueue([this,
+                                         dst_file,
+                                         block_ids,
+                                         job_state = job_state.get(),
```
orozery: I think you should remove the `= job_state.get()`
This will make sure job_state is not freed before a thread finishes accessing it.
Or is this an optimization since you `completed_tasks` prevents from freeing job state?

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,311 @@
...
+      bool is_store = true;
+      bool success = false;
+      // Execute the copy operation
+      try {
+        // Stage 1: copy tensors from GPU to staging CPU tensor.
+        TIME_EXPR("write phase 1: copy_blocks ",
+                  m_tensor_copier.copy_blocks(cpu_base, block_ids, is_store),
+                  "file: " + dst_file);
+
+        auto& tls_stream = ThreadPool::get_tls_stream();
```
orozery: You can remove this. Already defined above.

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,311 @@
...
+        cudaError_t err = cudaStreamSynchronize(tls_stream.stream());
+        if (err != cudaSuccess) {
+          std::cerr << "[ERROR] cudaStreamSynchronize failed: "
+                    << cudaGetErrorString(err) << std::endl;
```
orozery: Can you return here, and remove the else?

You can add a scope which executes the mark-task-completion.

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,311 @@
...
+      StagingBufferInfo& buf = ThreadPool::get_staging_buffer();
+      auto* cpu_base = static_cast<uint8_t*>(buf.ptr);
+      bool is_store = true;
+      bool success = false;
+      // Execute the copy operation
+      try {
+        // Stage 1: copy tensors from GPU to staging CPU tensor.
+        TIME_EXPR("write phase 1: copy_blocks ",
+                  m_tensor_copier.copy_blocks(cpu_base, block_ids, is_store),
+                  "file: " + dst_file);
```
orozery: We will pay the price for concatenating "file: " + dst_file even if debug is not enabled.
According to perplexity this can be avoided by changing TIME_EXPR:

```
#define TIME_EXPR(label_fmt, expr, ...)                                    \
  ([&]() -> bool {                                                         \
    if (!(storage_debug_enabled())) {                                      \
      return ((expr), true);                                               \
    }                                                                      \
    /* Format string + args only evaluated when debug enabled */            \
    auto __label = fmt::format(label_fmt, ##__VA_ARGS__);                  \
    auto __t0 = std::chrono::high_resolution_clock::now();                 \
    auto __ret = [&]() {                                                   \
      if constexpr (std::is_void_v<decltype(expr)>) {                      \
        (expr);                                                            \
        return true;                                                       \
      } else {                                                             \
        return (expr);                                                     \
      }                                                                    \
    }();                                                                   \
    auto __t1 = std::chrono::high_resolution_clock::now();                 \
    double __ms =                                                          \
        std::chrono::duration<double, std::milli>(__t1 - __t0).count();    \
    std::cout << "[DEBUG][TIME] " << __label << " took " << __ms << " ms" << std::endl; \
    return __ret;                                                          \
  })()
```

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,311 @@
...
+  // Create job state object to track progress and futures for this job.
+  auto job_state = std::make_unique<JobState>();
+  job_state->total_tasks = src_files.size();
+
+  // For each source file, enqueue one async task in the I/O thread pool.
+  for (size_t i = 0; i < src_files.size(); i++) {
+    std::string src_file = src_files[i];
+    auto block_ids = all_block_ids[i];
+    auto future = m_thread_pool.enqueue(
+        [this, src_file, block_ids, job_state = job_state.get()]() -> bool {
```
orozery: same as above regarding job_state.get()

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,311 @@
...
+          StagingBufferInfo& buf = ThreadPool::get_staging_buffer();
+          bool success = false;
+          try {
+            // Stage 1: Read file to staging CPU tensor.
+            // Read data from disk into a tensor.
+            success = TIME_EXPR("read phase 1: read_tensor_from_file",
+                                read_tensor_from_file(src_file, buf),
```
orozery: Can we open an issue to change this to read from offset, instead of the entire file, depending on gpu_blocks_per_file-len(block_ids)?

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,311 @@
...
+                                ("file:" + src_file));
+            if (!success) {
+              std::cerr << "[ERROR] Stage1 read_tensor_from_file failed for "
```
orozery: like above, can we return and remove the else?

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,311 @@
...
+                  << std::endl;
+        success = false;
+      } catch (...) {
+        std::cerr << "[ERROR] Store failed for " << dst_file << "\n";
+        success = false;
+      }
+
+      // Mark task completion.
+      job_state->completed_tasks.fetch_add(1);
+      // if (!success) job_state->all_success = false; // TODO- silent
```
orozery: I saw in the load you uncomment this. Can you uncomment here as well?

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
orozery: You introduce a shared_ptr, which you tried to avoid in `job_state`, I guess because of performance overhead?
Maybe you can just put an optional event in the job_state, and destroy it at `get_finished`?

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: Yes, I opened an issue with this improvement

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: I will fix the load until the connector can handle failure

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: I prefer the shared_ptr approach because it automatically destroys the event when the last worker thread finishes, without requiring manual cleanup and managing the events per job.

kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: I'm keeping the `raw pointer` capture for performance, `shared_ptr` would add atomic reference counting overhead, and correctness is already guaranteed safe since `get_finished()` only removes jobs after `completed_tasks == total_tasks`, ensuring all threads complete before the JobState is freed.

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: same as above

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
orozery: Then let's make it more explicit:
Use unique_ptr instead of shared_ptr, and add a comment on the `.get()` that explains why this is safe.

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
orozery: I also prefer it, but since I saw you were trying to avoid ref-count changes (in job_state), I thought this is a performance issue.

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: You're right. Since the number of worker threads is limited and both are captured once per job, the shared_ptr atomic overhead is negligible compared to disk I/O operations. I've changed job_state to shared_ptr for consistency.

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: See comment on shared pointer

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+  // For each dst_file file, enqueue one async task in the I/O thread pool.
+  for (size_t i = 0; i < dst_files.size(); i++) {
+    std::string dst_file = dst_files[i];
+    auto block_ids = all_block_ids[i];
+
+    auto future = m_thread_pool.enqueue([this,
+                                         dst_file,
+                                         block_ids,
+                                         job_state = job_state,
```
orozery: can you remove the "= job_state" ?

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,318 @@
...
+      struct Completion {
+        std::shared_ptr<JobState> job_state;
+        bool& success;
+        ~Completion() {
+          job_state->completed_tasks.fetch_add(1);
+          // if (!success) job_state->all_success = false; // TODO- silent
+          // ignore read failures for now offloading connector not able to
+          // handle failures
+        }
+      } completion{job_state, success};
```
orozery: I prefer to avoid the struct definition:
```
auto completion = [&, success](bool s = success) {
  job_state->completed_tasks.fetch_add(1);
  // if (!s) job_state->all_success = false;
}();
```

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -0,0 +1,318 @@
...
+          struct Completion {
+            std::shared_ptr<JobState> job_state;
+            bool& success;
+            ~Completion() {
+              job_state->completed_tasks.fetch_add(1);
+              // if (!success) job_state->all_success = false; // TODO- silent
+              // ignore read failures for now offloading connector not able to
+              // handle failures
+            }
+          } completion{job_state, success};
```
orozery: same as above

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: Done

orozery (dismissed):
LGTM! Thanks @kfirtoledo !

Issue #167, opened 2025-11-19
[RFC] Lora Support


sagearc:
This RFC proposes necessary changes to the llm-d kv cache manager to enable support for multiple LoRA adapters. The proposed solution involves ensuring the KV-Cache manager consistently uses the base model information for tokenization and incorporates the LoRA adapter name into block hashing for compatibility with the vLLM serving framework.

https://docs.google.com/document/d/1kAqPBqZctkXISoGfo6Z2E-vRKbBV4ifDvuP-2y1yTb8/edit?usp=sharing

PR #182, opened 2025-11-24
feat: Simplify tokenization prefix store to single-model architecture

sagearc:
## Summary
This PR refactors the tokenization prefix store architecture to support a single model per store instance, eliminating the need for model name management within the store implementations.

## Changes Made
- Simplified `LRUTokenStore` & `TrieTokenStore` - Removed multi base model support.
- Updated Indexer interface - Removed `modelName` parameters from `AddTokenization` and `FindLongestContainedTokens` methods. Model identity managed at scheduler/indexer level, not storage level
- Updated tests

Configure the base model name in the indexer config
Remove model name parameters from tokenization method calls
The change aligns with the single-model-per-scheduler architecture where each scheduler instance handles one specific base model.

Solves #190 
Part of #167

### DIFF
--- a/pkg/tokenization/prefixstore/lru_store.go
+++ b/pkg/tokenization/prefixstore/lru_store.go
@@ -87,7 +87,7 @@ func NewLRUTokenStore(config *Config) (Indexer, error) {
 }
 
 // AddTokenization adds the full tokenization of a string to the
-// indexer for a given model.
+// indexer.
 // The function assumes tokens and offsets are of the same length.
 // The function assumes that tokens will not be mutated after the call.
 func (c *LRUTokenStore) AddTokenization(prompt string, tokens []uint32,


pkg/tokenization/prefixstore/lru_store.go
```diff
@@ -63,7 +63,7 @@ type LRUTokenStore struct {
 	cacheSize int
 	blockSize int
 
-	store map[string]*lru.Cache[uint64, Block]
+	cache *lru.Cache[uint64, Block]
```
vMaroon, line 66: I think `index` might be a better name here.

sagearc: Considering that the prefix store caches tokenizations, the term index feels too generic. It’s also already used in both `kvcache/kvblock/index.go` and `kvcache/indexer.go`. I think we should choose a name that avoids this overloaded notation and better reflects the actual object the cache is representing.

Issue #155, opened 2025-11-03
Performance benchmark & optimization for the external tokenizer service


delavet:
In https://github.com/llm-d/llm-d-kv-cache-manager/pull/137, an external tokenizer service based on UDS (Unix Domain Socket) was proposed.
This service aims to enhance compatibility with vLLM tokenization by utilizing the Python transformers library. However, due to its external nature, it may impact tokenization performance. Therefore, an appropriate benchmark needs to be proposed so that we can make a reasonable trade-off between external and internal tokenization.
Additionally, this service can still benefit from performance optimization techniques. For example, using gRPC instead of HTTP+JSON could be considered. The benchmark can be used to measure the performance improvements brought by different optimization approaches.

PR #185, opened 2025-11-25
add metrics for tokenizaion and rendering chat template

delavet:
This PR, as part of #155, introduces corresponding observable metrics for the `Encode` call (tokenization) and `RenderChatTemplate` call of the tokenizers.
Additionally, considering the current situation where we have multiple implementations of tokenizers and use a `CompositeTokenizer` to manage them, it is important to observe differences between these implementations using metrics.
This PR also introduces a metrics server for the online example, allowing us to first measure the performance of the `llm-d-kv-cache-manager` module independently through the online example itself, isolating it from the rest of the EPP.


vMaroon (commented):
This is great, thanks!

/lgtm
/approve

PR #186, opened 2025-11-25
fix: Enable local tokenizer chat template rendering

pierDipi:
Fixes https://github.com/llm-d/llm-d-kv-cache-manager/issues/184

Fixes chat template rendering failures when using LocalTokenizer in airgapped
environments. Previously, the local tokenizer would fail to fetch chat templates
because AutoTokenizer.from_pretrained() was being called with a file path to
tokenizer.json instead of the model directory path.

Changes:
- Add IsLocalPath flag to FetchChatTemplateRequest to distinguish local vs remote models
- Update render_jinja_template_wrapper.py to detect file vs directory paths
- When given a file path, extract the parent directory before calling AutoTokenizer
- Set local_files_only=True to prevent network access in airgapped environments
- Update tokenizer.go to automatically set IsLocalPath=true for local tokenizers

This enables chat-completions requests to work correctly with LocalTokenizer in
airgapped environments without requiring network access to HuggingFace.


pkg/preprocessing/chat_completions/requirements.txt
```diff
@@ -3,5 +3,5 @@
 packaging==24.2
 pillow==11.2.1
 torch==2.5.1
-transformers>=4.53.0
+transformers>=4.53.0,<4.57.2
```
pierDipi, line 6: There is an issue in 1.57.2 of transformers https://github.com/huggingface/transformers/issues/42381

pierDipi:
Confirmed with @Jay-Pd that this works now, marking it as ready to review

tests/e2e/redis_mock/testdata/local-llama3/tokenizer.json
vMaroon, line 1: Can this be downloaded on demand instead?

pierDipi: Done

Issue #187, opened 2025-11-25
Lazy Tokenizer Initialization Fails on LoRA and Aliased Models Requests


sagearc:
## Summary
The KV-Cache Indexer's reliance on lazy tokenizer initialization (using the model ID from the inference request) breaks compatibility with LoRA adapters and model aliases (like OPT models using the GPT-2 tokenizer). Since the failure only happens on the first request, the service reports an incorrect healthy status.

A switch to an eager, fail-fast tokenizer load at startup is necessary.

## Detailed Description
The issue stems from the tokenizer lookup failing because the request's model parameter often doesn't match the actual Hugging Face ID of the required tokenizer. This failure is delayed, meaning the EPP is marked as ready for traffic despite being non-functional.

## Proposed Solution: Configuration-Driven Eager Load
To establish system integrity, the tokenizer should be treated as a mandatory, pre-validated dependency.

**Goal:** The ideal state involves removing complex, runtime mappings from various model names to tokenizers and instead **relying on a single, correct tokenizer loaded at launch**.

**Action:** The definitive Base Tokenizer ID (or Base Model ID) is configured and provided to the indexer/pool/tokenizer config component.

**Enforcement:** The service startup process loads the tokenizer using this configured ID. Any failure during this load triggers an immediate crash, ensuring the service is never marked as healthy in a broken state.

part of #167

PR #192, opened 2025-11-25
fix: Single tokenizer initialization at startup

sagearc:
#187 

This PR loads local and HF tokenizers at the pool initialization.
CachedTokenizer instances now carry a single tokenizer with respect to the base model name provided at startup.
Fixed tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,6 @@ require (
 	github.com/vmihailenco/msgpack/v5 v5.4.1
 	go.uber.org/multierr v1.11.0
 	golang.org/x/net v0.38.0
-	golang.org/x/sync v0.12.0
 	google.golang.org/grpc v1.68.1
 	google.golang.org/protobuf v1.36.5
 	k8s.io/apimachinery v0.33.0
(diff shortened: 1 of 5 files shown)


pkg/kvcache/indexer.go
```diff
@@ -34,6 +34,10 @@ import (
 // The configuration cover the different components found in the Indexer
 // module.
 type Config struct {
+	// BaseModelName is the canonical base model identifier.
+	// Used to distinguish base-model requests from LoRA-adapted requests
```
vMaroon: This comment is tainted with implicit context. I think it should just state that this name is required in order to initialize tokenizer if not local.

pkg/kvcache/indexer.go
```diff
@@ -74,7 +79,16 @@ type Indexer struct {
 // NewKVCacheIndexer creates a KVCacheIndex given a Config.
 func NewKVCacheIndexer(ctx context.Context, config *Config) (*Indexer, error) {
 	logger := log.FromContext(ctx)
-	if config != nil && config.TokenProcessorConfig != nil {
+
+	if config == nil {
+		return nil, fmt.Errorf("config cannot be nil")
+	}
+
+	if config.BaseModelName == "" {
```
vMaroon: If we're using a local tokenizer, does this hold?

PR #196, opened 2025-11-28
refactor: Redis memory optimization and test reliability improvements

natoscott:
Small improvements to the KV-block index implementation relating to Redis/Valkey: 
- optimize Redis memory usage by removing unused timestamp values from HSet operations.
- update the Valkey example documentation to clarify RDMA limitations in Go clients and correct the example output.

While testing the code change a flaky concurrent test was encountered - this is fixed here also.


vMaroon:
Thanks @natoscott.

/lgtm
/approve

PR #202, opened 2025-12-04
feat(kvblock): decouple key handling in index

### LINKED ISSUE #195: [Proposal] Decoupling Hash Algorithms between Manager and Inference Engine
When testing prefix cache aware routing, I often encounter issues because the hash seed, hash algorithm, or block size are mismatched between a vllm config and the precie-prefix-cache-scorer config.

Therefore, I propose separating the key into a `requestKey` and an `engineKey`, and storing both. 
The `requestKey` is calculated by the **scheduler**, while the `engineKey` is calculated by the **inference engine** (received via BlockStored and BlockRemoved events).

**Before:**

```go
type Index interface {
	Lookup(ctx context.Context, keys []Key, podIdentifierSet sets.Set[string]) (map[Key][]PodEntry, error)
	Add(ctx context.Context, keys []Key, entries []PodEntry) error
	Evict(ctx context.Context, key Key, entries []PodEntry) error
}
```

**After:**

```go
type Index interface {
	Lookup(ctx context.Context, requestKeys []Key, podIdentifierSet sets.Set[string]) (map[Key][]PodEntry, error)
	Add(ctx context.Context, engineKeys []Key, requestKeys []Key, entries []PodEntry) error
	Evict(ctx context.Context, engineKey Key, entries []PodEntry) error
}
```

**Workflow:**

- Handling a request
  1. Tokenize the prompt.
  2. Compute the `requestKey`s.
  3. Query the keyToPods using these `requestKey`s.
  4. Score pods.
- Handling a BlockStored
  1. Get tokens, parent hash, block hashes (`engineKey`s) from an event.
  2. Find parent `requestKey` by parent `engineKey`.
  3. Compute the `requestKey`s with tokens and parent `requestKey`.
  4. Store the mapping `requestKey` -> metadata.
  5. Store the mapping `engineKey` -> `requestKey`.
- Handling a BlockRemoved
  1. Lookup `requestKey` by `engineKey`.
  2. Remove the mapping `requestKey` -> metadata.
  3. Remove the mapping `engineKey` -> `requestKey` if `requestKey` lacks metadata.

hhk7734:
close #195

### DIFF
--- a/pkg/kvcache/kvblock/cost_aware_memory.go
+++ b/pkg/kvcache/kvblock/cost_aware_memory.go
@@ -176,7 +176,7 @@ func (m *CostAwareMemoryIndex) Add(ctx context.Context, engineKeys, requestKeys 
 	for i, requestKey := range requestKeys {
 		engineKey := engineKeys[i]
 
-		// 1. Store engineKey -> requestKey mapping
+		// Store engineKey -> requestKey mapping
 		m.requestKeys.Add(engineKey, requestKey)
 
 		keyStr := requestKey.String()
--- a/pkg/kvcache/kvblock/in_memory.go
+++ b/pkg/kvcache/kvblock/in_memory.go
@@ -61,15 +61,15 @@ func NewInMemoryIndex(cfg *InMemoryIndexConfig) (*InMemoryIndex, error) {
 		return nil, fmt.Errorf("failed to initialize in-memory index: %w", err)
 	}
 
-	requestKeys, err := lru.New[Key, Key](cfg.Size)
+	engineToRequestKeys, err := lru.New[Key, Key](cfg.Size)
 	if err != nil {
 		return nil, fmt.Errorf("failed to initialize in-memory engine key map: %w", err)
 	}
 
 	return &InMemoryIndex{
-		data:         cache,
-		requestKeys:  requestKeys,
-		podCacheSize: cfg.PodCacheSize,
+		data:                cache,
+		engineToRequestKeys: engineToRequestKeys,
+		podCacheSize:        cfg.PodCacheSize,
 	}, nil
 }
 
@@ -77,8 +77,8 @@ func NewInMemoryIndex(cfg *InMemoryIndexConfig) (*InMemoryIndex, error) {
 type InMemoryIndex struct {
 	// data holds the mapping of requestKeys to sets of pod identifiers.
 	data *lru.Cache[Key, *PodCache]
-	// requestKeys holds the mapping of engineKeys to requestKeys.
-	requestKeys *lru.Cache[Key, Key]
+	// engineToRequestKeys holds the mapping of engineKeys to requestKeys.
+	engineToRequestKeys *lru.Cache[Key, Key]
 	// podCacheSize is the maximum number of pod entries per key.
 	podCacheSize int
 }
@@ -114,10 +114,10 @@ func (m *InMemoryIndex) Lookup(ctx context.Context, requestKeys []Key,
 	podsPerKey := make(map[Key][]PodEntry)
 	highestHitIdx := 0
 
-	for idx, key := range requestKeys {
-		if pods, found := m.data.Get(key); found { //nolint:nestif // TODO: can this be optimized?
+	for idx, requestKey := range requestKeys {
+		if pods, found := m.data.Get(requestKey); found { //nolint:nestif // TODO: can this be optimized?
 			if pods == nil || pods.cache.Len() == 0 {
-				traceLogger.Info("no pods found for key, cutting search", "key", key)
+				traceLogger.Info("no pods found for key, cutting search", "key", requestKey)
 				return podsPerKey, nil // early stop since prefix-chain breaks here
 			}
 
@@ -125,17 +125,17 @@ func (m *InMemoryIndex) Lookup(ctx context.Context, requestKeys []Key,
 
 			if podIdentifierSet.Len() == 0 {
 				// If no pod identifiers are provided, return all pods
-				podsPerKey[key] = pods.cache.Keys()
+				podsPerKey[requestKey] = pods.cache.Keys()
 			} else {
 				// Filter pods based on the provided pod identifiers
 				for _, pod := range pods.cache.Keys() {
 					if podIdentifierSet.Has(pod.PodIdentifier) {
-						podsPerKey[key] = append(podsPerKey[key], pod)
+						podsPerKey[requestKey] = append(podsPerKey[requestKey], pod)
 					}
 				}
 			}
 		} else {
-			traceLogger.Info("key not found in index", "key", key)
+			traceLogger.Info("key not found in index", "key", requestKey)
 		}
 	}
 
@@ -160,7 +160,7 @@ func (m *InMemoryIndex) Add(ctx context.Context, engineKeys, requestKeys []Key, 
 		engineKey := engineKeys[i]
 
 		// 1. Store engineKey -> requestKey mapping
-		m.requestKeys.Add(engineKey, requestKey)
+		m.engineToRequestKeys.Add(engineKey, requestKey)
 
 		// 2. Store requestKey -> PodCache mapping
 		var podCache *PodCache
@@ -216,7 +216,7 @@ func (m *InMemoryIndex) Evict(ctx context.Context, engineKey Key, entries []PodE
 
 	traceLogger := log.FromContext(ctx).V(logging.TRACE).WithName("kvblock.InMemoryIndex.Evict")
 
-	requestKey, found := m.requestKeys.Get(engineKey)
+	requestKey, found := m.engineToRequestKeys.Get(engineKey)
 	if !found {
 		traceLogger.Info("engineKey not found in index, nothing to evict", "engineKey", engineKey)
 		return nil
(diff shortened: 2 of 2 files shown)


pkg/kvcache/kvblock/cost_aware_memory.go
```diff
@@ -69,15 +70,23 @@ func NewCostAwareMemoryIndex(cfg *CostAwareMemoryIndexConfig) (*CostAwareMemoryI
...
 	}, nil
 }
 
 // CostAwareMemoryIndex implements the Index interface using Ristretto cache for cost-aware memory management.
 type CostAwareMemoryIndex struct {
-	// data holds the mapping of keys to sets of pod identifiers.
+	// data holds the mapping of request keys to sets of pod identifiers.
 	data *ristretto.Cache[string, *CostPodCache]
+	// requestKeys holds the mapping of engine keys to request keys.
+	requestKeys *lru.Cache[Key, Key]
```
vMaroon, line 95: I think a comment on the management of this map in relation to `data` can be helpful. Especially to clarify correctness (no hanging keys or missing mappings) and size binding since this is a cost-aware implementation.

pkg/kvcache/kvblock/cost_aware_memory.go
```diff
@@ -255,11 +280,21 @@ func (m *CostAwareMemoryIndex) Evict(ctx context.Context, key Key, entries []Pod
...
 	} else if podCacheLenBefore != podCache.Len() {
 		m.data.Set(keyStr, podCache, podCache.CalculateByteSize(keyStr))
-		traceLogger.Info("evicted pods from key", "key", key, "pods", entries)
+		traceLogger.Info("evicted pods from engineKey", "requestKey", requestKey, "engineKey", engineKey, "pods", entries)
 	}
 	m.data.Wait()
 	return nil
 }
+
+func (m *CostAwareMemoryIndex) GetRequestKey(ctx context.Context, engineKey Key) (Key, error) {
```
vMaroon, line 302: Function should have descriptive comment. This is relevant for all the implementations.

pkg/kvcache/kvblock/cost_aware_memory.go
```diff
@@ -145,18 +160,26 @@ func (c *CostPodCache) CalculateByteSize(keyStr string) int64 {
...
+	}
 
 	traceLogger := log.FromContext(ctx).V(logging.TRACE).WithName("kvblock.CostAwareMemoryIndex.Add")
 
-	for _, key := range keys {
-		keyStr := key.String()
+	for i, requestKey := range requestKeys {
+		engineKey := engineKeys[i]
+
+		// 1. Store engineKey -> requestKey mapping
```
vMaroon: I think `1. ` should be dropped since there isnt a continuation.

pkg/kvcache/kvblock/in_memory.go
```diff
@@ -61,16 +61,24 @@ func NewInMemoryIndex(cfg *InMemoryIndexConfig) (*InMemoryIndex, error) {
...
 	}, nil
 }
 
 // InMemoryIndex is an in-memory implementation of the Index interface.
 type InMemoryIndex struct {
-	// data holds the mapping of keys to sets of pod identifiers.
+	// data holds the mapping of requestKeys to sets of pod identifiers.
 	data *lru.Cache[Key, *PodCache]
+	// requestKeys holds the mapping of engineKeys to requestKeys.
+	requestKeys *lru.Cache[Key, Key]
```
vMaroon: For clarity, this can be called `engineToRequestKeys` (relevant to all other backends).

pkg/kvcache/kvblock/in_memory.go
```diff
@@ -86,27 +94,27 @@ type PodCache struct {
...
+		return nil, fmt.Errorf("no requestKeys provided for lookup")
 	}
 
 	traceLogger := log.FromContext(ctx).V(logging.TRACE).WithName("kvblock.InMemoryIndex.Lookup")
 
 	podsPerKey := make(map[Key][]PodEntry)
 	highestHitIdx := 0
 
-	for idx, key := range keys {
+	for idx, key := range requestKeys {
```
vMaroon: s/key/engineKey/.

pkg/kvcache/kvblock/token_processor.go
```diff
@@ -76,54 +77,44 @@ func NewChunkedTokenDatabase(config *TokenProcessorConfig) TokenProcessor {
...
-		return nil
+		return 0
 	}
 
-	sum := sha256.Sum256(b)
-	return sum[:] // Return the full 32-byte hash
+	h := fnv.New64a()
+	_, _ = h.Write(b)
+	return h.Sum64()
 }
```
sagearc, line 112: The request hashes do not have to be aligned with the Engine hashes. I suggest simplifying this hash function with a more naive and easy-to-read approach. Also, returning 0 as the default value could cause collisions when one of these functions fails.

pkg/kvcache/kvevents/pool.go
```diff
@@ -271,12 +273,31 @@ func (p *Pool) digestEvents(ctx context.Context, podIdentifier, modelName string
...
+					debugLogger.Error(err, "Failed to convert parent block hash for BlockStored event",
+						"rawHash", ev.ParentBlockHash)
+					continue
+				}
+
+				parentEngineKey := kvblock.Key{ModelName: modelName, ChunkHash: hash}
+
+				key, err := p.index.GetRequestKey(ctx, parentEngineKey)
+				if err == nil {
+					parentRequestKey = &key
```
sagearc, line 292: The parent request key probably will not be aligned with the parent engine key. If this is the case, we would calculate subsequent request keys based on a parent that **we likely cannot reach**, as we will not be able to reconstruct the hashing logic. I would expect an error or warning to be raised, and the event to be dropped without updating the index in that case...

pkg/kvcache/kvblock/in_memory.go
```diff
@@ -137,26 +145,35 @@ func (m *InMemoryIndex) Lookup(ctx context.Context, keys []Key,
...
 		var podCache *PodCache
 		var found bool
 
 		// Try to get existing cache first
-		podCache, found = m.data.Get(key)
+		podCache, found = m.data.Get(requestKey)
 		//nolint:nestif // double-checked locking pattern
 		if !found {
 			// Create new cache
 			cache, err := lru.New[PodEntry, struct{}](m.podCacheSize)
```
sagearc, line 174: This might not be directly relevant to the PR, but why do we use an LRU cache for every request key instead of a simple set of keys? If a key were to be evicted from a certain pod, we would receive that information from the evict kv event. My concern is that we might mistakenly evict a system prompt, which acts as a prefix to all keys, and we are limited by the value of m.podCacheSize to the number of unique inputs we can cache
@vMaroon

pkg/kvcache/kvblock/in_memory.go
```diff
@@ -185,23 +202,30 @@ func (m *InMemoryIndex) Add(ctx context.Context, keys []Key, entries []PodEntry)
 		}
 		podCache.mu.Unlock()
```
sagearc, line 203: Regarding previous comment, we could use a simple `sync.Map` from request keys to a set of pod identifiers, I think there's no need for a mutex lock, and it would also fix that potential bug mentioned above. @vMaroon

pkg/kvcache/kvblock/in_memory.go
vMaroon: This cache holds the pods that have the associated individual key. Its capacity should be set to some multiplier to the expected number of serving pods.

>we are limited by the value of m.podCacheSize to the number of unique inputs we can cache

I don't think this is true. This adds a constraint over the number of server hits you can get per sequence of tokens, and not a constraint over the number of the latter.

If we use a simple set of keys then we'll also have to implement pod pruning, which is not justified yet IMO.

pkg/kvcache/kvblock/in_memory.go
vMaroon: A lock is needed on this level since we don't hold a global lock over the index. Besides for concurrency reasons, the above comment explains why an LRU is preferred.

pkg/kvcache/kvblock/in_memory.go
sagearc: Why limit this and not have a precise map of all keys to pods?
number of keys > number of requests >> than the number of pods either way

pkg/kvcache/kvblock/token_processor.go
vMaroon: Can you expand on "naive and easy-to-read approach"? While we do not have to recreate vLLM's exact hashing, we still need to maintain a prefix-hash chain, and allow tainting chains with cache-altering context such as lora/multimodality... etc.

The above constraints are nicely addressed by this hashing template, leaving the serialisation/hashing algorithms configurable.

pkg/kvcache/kvblock/in_memory.go
sagearc: Assuming that `sync.Map` is thread safe, and we have much more reads than writes, I believe we can delegate the concurrency handling to the `sync` library

pkg/kvcache/kvblock/token_processor.go
sagearc: The cbor, endianness, hash256 logic (much of it was modified/added here: https://github.com/llm-d/llm-d-kv-cache/pull/138) are vllm implementation specific

pkg/kvcache/kvblock/in_memory.go
vMaroon: Because you'd have to write the logic that maintains this list. On a happy path admissions and evictions are enough, but pod failures add avoidable complication.

I prefer not to optimize for things we're not sure we need, if the cost is non-trivial.

Issue #101, opened 2025-09-01
OpenAI Chat-Completions API Multi-Modal Support


vMaroon:

yankay:
Hi @vMaroon,

I have a question about handling image URLs when adding multi‑modal support to the OpenAI Chat Completions workflow:

- Option A: Download the image and convert its content into a key/token (for example, like the approach in [image_loader.py](https://github.com/ai-dynamo/dynamo/blob/main/components/src/dynamo/vllm/multimodal_utils/image_loader.py#L41)), then use that content-derived key for KV calculations.
- Option B: Generate a token directly from the image URL itself (note that URL-derived tokens will not be consistent with the tokens used in kvevent).

Which approach would you consider more appropriate or standard for a production system? 

Thanks for any guidance.

sagearc:
Completed as part of v0.7 milestones

PR #219, opened 2025-12-16
feat: Support multi modal chat completion requests

sagearc:
**Context:**
The GIE recently added support for images (https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/1617) as part of ongoing multi-modal request support. However, the KV Cache Manager still uses outdated chat completions request type (`preprocessing.RenderJinjaTemplateRequest`) and Message (`preprocessing.RenderJinjaTemplateRequest.ChatMessage`) type definitions that don't support the new multi-modal capabilities.

**Changes:**
This PR refactors the kv cache manager to reuse the existing GIE request types that already support multi-modal content. 

**Reference:**
Note that the GIE already parses the chat completion fields from the request into these types:
- https://github.com/vMaroon/gateway-api-inference-extension/blob/2f8d627894784fa7fa9aa4cb37f26dd6e9483f22/pkg/epp/requestcontrol/director.go#L106
- https://github.com/vMaroon/gateway-api-inference-extension/blob/2f8d627894784fa7fa9aa4cb37f26dd6e9483f22/pkg/epp/util/request/body.go#L40-L50

### DIFF
--- a/pkg/preprocessing/chat_completions/types.go
+++ b/pkg/preprocessing/chat_completions/types.go
@@ -40,8 +40,6 @@ type RenderJinjaTemplateRequest struct {
 	ContinueFinalMessage      bool                   `json:"continue_final_message,omitempty"`
 	AddGenerationPrompt       bool                   `json:"add_generation_prompt,omitempty"`
 	ChatTemplateKWArgs        map[string]interface{} `json:"chat_template_kwargs,omitempty"`
-	// CacheSalt is an optional request parameter to isolate prefix caches for security reasons.
-	CacheSalt string `json:"cache_salt,omitempty"`
 }
 
 func (r *RenderJinjaTemplateRequest) String() string {


sagiahrac:
Related to #101

go.mod
```diff
@@ -8,67 +8,74 @@ require (
...
-	k8s.io/client-go v0.33.0
-	sigs.k8s.io/controller-runtime v0.21.0
+	go.uber.org/zap v1.27.0
+	golang.org/x/net v0.46.0
+	google.golang.org/grpc v1.76.0
+	google.golang.org/protobuf v1.36.10
+	k8s.io/apimachinery v0.34.2
+	k8s.io/client-go v0.34.2
+	sigs.k8s.io/controller-runtime v0.22.4
+	sigs.k8s.io/gateway-api-inference-extension v1.2.1
```
delavet: Can we achieve this goal by modifying the struct in the current package instead of importing `sigs.k8s.io/gateway-api-inference-extension`?
Considering that `sigs.k8s.io/gateway-api-inference-extension` is a widely referenced component in the GAIE ecosystem, and `llm-d-kv-cache` is also in a position of being referenced (for example, `llm-d-inference-scheduler` references both), allowing `llm-d-kv-cache` to reference `sigs.k8s.io/gateway-api-inference-extension` could introduce many potential dependency conflicts. IMO, adding such extra complexity just for reusing a struct is not worthwhile.

delavet:
Additionally, I'm just curious. Is there any specific implementation plan or related design that I can refer to for the "hybrid approach" mentioned in [#101](https://github.com/llm-d/llm-d-kv-cache/issues/101#issuecomment-3660761481)? This PR currently only adds support for struct fields, but it seems that it still doesn't fully support multi-modal functionality.

go.mod
sagearc: Good point, copied the structs from GIE and removed the extra dependency.

pkg/preprocessing/chat_completions/types.go
```diff
@@ -0,0 +1,162 @@
...
+	"strings"
+)
+
+const nilString = "<nil>"
+
+// ChatCompletionsRequest is a structured representation of the fields we parse out of the v1/chat/completions
+// request body. For detailed body fields, please refer to https://platform.openai.com/docs/api-reference/chat.
+// This struct includes fields usable for plugins and scheduling decisions - and not the entire
+// API spec.
+type ChatCompletionsRequest struct {
```
vMaroon: The API in this library is responsible only for what affects the KV-cache, in this case that is equivalent to fields used in rendering: ones needed by the transformers render-jinja-template function. 
The scheduler is responsible for handling the OpenAI API input.

With the above in mind, this rename is imprecise. What do you think?

pkg/preprocessing/chat_completions/types.go
```diff
@@ -0,0 +1,162 @@
...
+type Message struct {
+	// Role is the message Role, optional values are 'user', 'assistant', ...
+	Role string `json:"role,omitempty"`
+	// Content defines text of this message
+	Content Content `json:"content,omitempty"`
+}
+
+type Content struct {
+	Raw        string
+	Structured []ContentBlock
```
vMaroon, line 80: This is the essence of what's missing, right?

pkg/preprocessing/chat_completions/types.go
sagearc: The fields are exactly the same as in the EPP struct, so this might help reduce ambiguity. That said, your point also makes sense, so I changed it back to `RenderJinjaTemplateRequest`.

pkg/preprocessing/chat_completions/types.go
sagearc: Yes

delavet:
LGTM

pkg/preprocessing/chat_completions/types.go
vMaroon: Thanks. `CacheSalt` is still excessive, we should open an issue for supporting it and implement it on the request level, and not chat templating.

PR #221, opened 2025-12-18
Refactor: Replace daulet/tokenizers with vLLM

hyeongyun0916:
This PR refactors the tokenization system to use vLLM's tokenizer wrapper instead of the daulet/tokenizers library. The changes streamline the tokenization pipeline by removing the external tokenizer dependency and consolidating chat template rendering and encoding through vLLM's unified API.

https://llm-d.slack.com/archives/C0A0SU5J68Y/p1764153758005369

### DIFF
--- a/pkg/preprocessing/chat_completions/cgo_functions.go
+++ b/pkg/preprocessing/chat_completions/cgo_functions.go
@@ -49,7 +49,6 @@ type ChatTemplateRequest struct {
 
 // ApplyChatTemplateRequest represents the request to render a chat template.
 type ApplyChatTemplateRequest struct {
-	// `conversation` is the transformers name, but we use `messages` for consistency with OpenAI API.
 	// The Python wrapper will handle converting this to a batched list if needed.
 	ChatTemplateRequest
 	Conversation              []Conversation         `json:"conversation"`
--- a/pkg/preprocessing/chat_completions/tokenizer_wrapper.py
+++ b/pkg/preprocessing/chat_completions/tokenizer_wrapper.py
@@ -21,7 +21,6 @@ import json
 import logging
 import os
 import sys
-from typing import Optional, Union
 from vllm.transformers_utils.tokenizer import get_tokenizer
 
 # Basic logging setup
@@ -58,18 +57,17 @@ def apply_chat_template(request_json):
             - model (str): The model ID or path (HF model ID, local directory path, or path to tokenizer file).
             - revision (str, optional): Model revision.
             - token (str, optional): Hugging Face token for private models.
-            - conversations (list): List of conversation lists
+            - conversation (list): List of conversation lists
             - chat_template (str, optional): The template to use
             - tools (list, optional): Tool schemas
             - documents (list, optional): Document schemas
             - return_assistant_tokens_mask (bool, optional): Whether to return assistant tokens mask
             - continue_final_message (bool, optional): Whether to continue final message
             - add_generation_prompt (bool, optional): Whether to add generation prompt
-            - kwargs (dict, optional): Additional rendering variables
-            - isVLLM / - isSGLang enum
-            
+            - chat_template_kwargs (dict, optional): Additional rendering variables
+
     Returns:
-        str: JSON string containing 'rendered_chats' and 'generation_indices' keys.
+        str: The rendered chat template as a string.
     """
 
     try:
@@ -118,9 +116,6 @@ def encode(request_json: str) -> str:
             - download_dir (str, optional): Directory to download the model.
             - text (str): The text to encode.
             - token (str, optional): Hugging Face token for private models.
-            - isVLLM (bool, optional): Whether to use VLLM tokenizer.
-            - isSGLang (bool, optional): Whether to use SG-Lang tokenizer.
-            - ....
 
     Returns:
         str: JSON string containing 'encoded_texts' key with list of token ID lists.


pkg/tokenization/tokenizer.go
```diff
@@ -301,69 +293,92 @@ func NewCachedHFTokenizer(modelID string, config *HFTokenizerConfig) (Tokenizer,
...
+	modelName string, req *preprocessing.ApplyChatTemplateRequest,
+) (string, error) {
+	ctx := context.TODO()
+
+	req.IsLocal = true
+	path, ok := t.localTokenizerConfig.ModelTokenizerMap[req.Model]
+	if !ok {
+		return "", fmt.Errorf("tokenizer for model %q not found", modelName)
+	}
+	req.Model = filepath.Dir(path)
```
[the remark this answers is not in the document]
hyeongyun0916: No need to copy the request here because CompositeTokenizer already creates a deep copy before calling this method.

pkg/tokenization/pool.go
```diff
@@ -200,22 +198,37 @@ func (pool *Pool) workerLoop(_ int) {
 // processTask tokenizes the prompt and updates the indexer.
 // It sends exactly one response (success or error) if ResultCh is provided.
 func (pool *Pool) processTask(task Task) error {
+	addSpecialTokens := true
 	if task.RenderReq != nil {
+		task.RenderReq.Model = task.ModelName
```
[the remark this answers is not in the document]
hyeongyun0916: It is safe to mutate task.RenderReq.Model here because ApplyChatTemplate performs a deep copy internally.

pkg/tokenization/tokenizer_test.go
```diff
@@ -171,28 +203,38 @@ func TestCachedLocalTokenizer_Encode(t *testing.T) {
...
+			invalidModelName: "testdata/test-model/tokenizer.json",
 		},
 	}
-	tokenizer, err := NewCachedLocalTokenizer(invalidModelName, config)
-	require.Error(t, err)
-	require.Nil(t, tokenizer)
+	tokenizer, err := NewCachedLocalTokenizer(config)
+	require.NoError(t, err)
+	require.NotNil(t, tokenizer)
 }
```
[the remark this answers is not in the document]
hyeongyun0916: This test verifies that creating a LocalTokenizer with an invalid model in the config fails at construction time.

pkg/preprocessing/chat_completions/tokenizer_wrapper.py
```diff
@@ -0,0 +1,195 @@
...
+    This function is aligned with the Go cgo_functions.go structs.
+
+    Args:
+        request_json (str): JSON string containing the request parameters:
+            - is_local (bool, optional): Whether the model is local.
+            - model (str): The model ID or path (HF model ID, local directory path, or path to tokenizer file).
+            - revision (str, optional): Model revision.
+            - token (str, optional): Hugging Face token for private models.
+            - conversations (list): List of conversation lists
+            - chat_template (str, optional): The template to use
```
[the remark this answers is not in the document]
hyeongyun0916: [apply copilot review](https://github.com/llm-d/llm-d-kv-cache/pull/221/commits/b198108724269203b70ba18c1bb6a48cf8455bb7)

pkg/preprocessing/chat_completions/cgo_functions.go
```diff
@@ -33,17 +33,26 @@ import (
...
+	DownloadDir string `json:"download_dir,omitempty"`
+	Model       string `json:"model"`
+	Revision    string `json:"revision,omitempty"`
+	Token       string `json:"token,omitempty"`
+}
+
+// ApplyChatTemplateRequest represents the request to render a chat template.
+type ApplyChatTemplateRequest struct {
+	// `conversation` is the transformers name, but we use `messages` for consistency with OpenAI API.
 	// The Python wrapper will handle converting this to a batched list if needed.
```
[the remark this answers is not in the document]
hyeongyun0916: [apply copilot review](https://github.com/llm-d/llm-d-kv-cache/pull/221/commits/b198108724269203b70ba18c1bb6a48cf8455bb7)

pkg/preprocessing/chat_completions/tokenizer_wrapper.py
```diff
@@ -0,0 +1,195 @@
...
+#!/usr/bin/env python3
+"""
+Standalone wrapper for tokenizer from vllm.
+"""
+
+import json
+import logging
+import os
+import sys
+from typing import Optional, Union
```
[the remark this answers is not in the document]
hyeongyun0916: [apply copilot review](https://github.com/llm-d/llm-d-kv-cache/pull/221/commits/b198108724269203b70ba18c1bb6a48cf8455bb7)

pkg/preprocessing/chat_completions/tokenizer_wrapper.py
```diff
@@ -0,0 +1,195 @@
...
+            - documents (list, optional): Document schemas
+            - return_assistant_tokens_mask (bool, optional): Whether to return assistant tokens mask
+            - continue_final_message (bool, optional): Whether to continue final message
+            - add_generation_prompt (bool, optional): Whether to add generation prompt
+            - kwargs (dict, optional): Additional rendering variables
+            - isVLLM / - isSGLang enum
+            
+    Returns:
+        str: JSON string containing 'rendered_chats' and 'generation_indices' keys.
+    """
```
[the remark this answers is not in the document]
hyeongyun0916: [apply copilot review](https://github.com/llm-d/llm-d-kv-cache/pull/221/commits/b198108724269203b70ba18c1bb6a48cf8455bb7)

pkg/tokenization/pool.go
```diff
@@ -200,22 +198,37 @@ func (pool *Pool) workerLoop(_ int) {
...
 		var err error
-		task.Prompt, err = pool.tokenizer.RenderChatTemplate(pool.modelName, task.RenderReq)
+		task.Prompt, err = pool.tokenizer.ApplyChatTemplate(task.ModelName, task.RenderReq)
 		if err != nil {
-			log.Log.Error(err, "failed to render chat template")
+			log.Log.Error(err, "failed to render chat template", "modelName", task.ModelName)
 			return err
 		}
+		addSpecialTokens = false
```
hyeongyun0916, line 210: [completion default is True](https://github.com/vllm-project/vllm/blob/v0.11.2/vllm/entrypoints/openai/protocol.py#L1127)
[chatCompletion default is False](https://github.com/vllm-project/vllm/blob/v0.11.2/vllm/entrypoints/openai/protocol.py#L613)

hyeongyun0916:
> On the CI issue, could you try adding a cleanup step? https://github.com/marketplace/actions/free-disk-space-ubuntu
Feel free to defer.

Fixed the CI failure by replacing the vllm package with the CPU-only version to save space.

pkg/preprocessing/chat_completions/tokenizer_wrapper.py
```diff
@@ -0,0 +1,190 @@
...
+        lock = _get_tokenizer_cache_lock()
+        with lock:
+            cache_key = f"{model_name}:{revision or 'main'}:{is_local}"
+            tokenizer = _tokenizer_cache.get(cache_key)
+            if tokenizer is None:
+                os.environ["HF_TOKEN"] = token
+                tokenizer = get_tokenizer(model_name, trust_remote_code=True, revision=revision, download_dir=download_dir)
+                _tokenizer_cache[cache_key] = tokenizer
+            
+            request["tokenize"] = False
```
hyeongyun0916, line 100: For now, I've set tokenize to False to support completions instead of chat completions. It's worth considering unify this into a single process later on.

vMaroon: I haven't gotten to review to depth yet - but the fix that removes BOS duplicates would make it work for chat, right?

hyeongyun0916: Regarding the BOS duplicates, that part is specifically related to this discussion: https://github.com/llm-d/llm-d-kv-cache/pull/221#discussion_r2634052334

As for the `request["tokenize"] = False` setting:

Currently, Completion only performs encode.
ChatCompletion performs applyChatTemplate followed by encode.

To ensure both follow the same [encoding logic](https://github.com/llm-d/llm-d-kv-cache/blob/main/pkg/tokenization/pool.go#L216), I’ve configured it so that in this stage, we only perform applyChatTemplate without the actual encoding. This allows the subsequent encoding step to be handled consistently.

pkg/preprocessing/chat_completions/tokenizer_wrapper.py
```diff
@@ -0,0 +1,190 @@
...
+            model_name = os.path.dirname(model_name)
+
+        lock = _get_tokenizer_cache_lock()
+        with lock:
+            cache_key = f"{model_name}:{revision or 'main'}:{is_local}"
+            tokenizer = _tokenizer_cache.get(cache_key)
+            if tokenizer is None:
+                os.environ["HF_TOKEN"] = token
+                tokenizer = get_tokenizer(model_name, trust_remote_code=True, revision=revision, download_dir=download_dir)
+                _tokenizer_cache[cache_key] = tokenizer
```
sagearc, line 144: Do we need to acquire the cache lock on every access, even after the tokenizer is initialized and in memory? If `tokenizer.encode` / `tokenizer(...)` is thread-safe here, as in the daulet tokenizers that rely on stateless HF tokenizers, we could likely restrict locking to the initialization path only and avoid holding the lock during encode calls.

hyeongyun0916: I agree with your point regarding the lock. I initially kept the cache lock as it was in the [existing implementation](https://github.com/llm-d/llm-d-kv-cache/blob/main/pkg/preprocessing/chat_completions/render_jinja_template_wrapper.py#L167) to avoid any sudden structural changes.

Since I'm planning to split this PR as you suggested, would it be better to remove the lock for the `encode` and `chatTemplate` in the new PR? I'd like to hear your thoughts on this before I start refactoring.

sagearc: I believe this should work. I think in most cases vllm relies on the underlying `transformers` tokenizer, not sure about any side effects.

Are you sure it’s thread-safe? It would probably be best to run a benchmark before and after the change to confirm there’s no regression.

hyeongyun0916: Regarding thread-safety, get_tokenizer in vLLM already utilizes get_lock internally to prevent race conditions during the download and initialization phase. For the subsequent calls, I believe the underlying tokenizer is stateless, but as you suggested, I will run a concurrency benchmark to confirm there are no regressions or race conditions when the lock is removed.

PR #224, opened 2025-12-22
Limit task requeue retries on faiures

alpe:
The `TokenizationPool` implemented a worker pool that re-queued failed tasks indefinitely without a retry limit. Specifically, the `Tokenize` method blocked on a result channel that was never closed if the task failed, leading to a permanent goroutine leak for every poisonous input.


vMaroon:
Sorry for the delay, happy holidays @alpe!
Thank you for fixing this.

/lgtm
/approve

PR #225, opened 2025-12-22
Async tokenize exec

alpe:
The blocking call was offloaded to a thread executor using `loop.run_in_executor`, ensuring the event loop remains responsive.


vMaroon:
@alpe thank you for the contribution!

@delavet pls take a look.

osswangxining (approved):
LGTM,  Execute synchronous tokenization operation in thread pool

PR #227, opened 2025-12-23
fix(kvcache): move in-memory indexer to fallback to respect external configs

hyeongyun0916:
Currently, the precise_prefix_cache in the scheduler initializes the configuration with default values (in-memory). Because the switch statement in the indexer evaluates the inmemory case first, it unintentionally overrides explicit Redis or Valkey configurations.

This PR refactors the switch logic to move the inmemory case to the bottom as a fallback. This ensures that user-defined configurations (Redis/Valkey) are correctly recognized and prioritized

https://llm-d.slack.com/archives/C09QG40A3GW/p1766471542097329

### DIFF
--- a/pkg/kvcache/kvblock/index.go
+++ b/pkg/kvcache/kvblock/index.go
@@ -65,11 +65,6 @@ func NewIndex(ctx context.Context, cfg *IndexConfig) (Index, error) {
 	var err error
 
 	switch {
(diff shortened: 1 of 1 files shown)


vMaroon:
Good catch @hyeongyun0916, thanks!

Sorry for the delay in reviews and happy holidays!

/lgtm
/approve

PR #233, opened 2025-12-24
fix(redis): atomic engine key eviction to prevent race condition

### LINKED ISSUE #232: [Redis] Race condition in index eviction
**What happened**:

There is a race condition in the `Evict` method when cleaning up keys from the kv cache indexer.

Currently, the logic for removing an engine key when its request key becomes empty (has no pods) is not atomic:

```mermaid
sequenceDiagram
    participant Evict as Thread A (Evict)
    participant Add as Thread B (Add)
    participant Redis

    Evict->>Redis: 1. Check if RequestKey has no pods (Result: 0)
    
    Note over Evict, Add: ⚠ RACE GAP ⚠
    
    Add->>Redis: 2. Add Pod to RequestKey (Count becomes 1)
    
    Evict->>Redis: 3. DEL Engine Key
    Note right of Evict: Deletes because Step 1 saw 0
    
    Note over Evict, Add: BUG: Engine Key deleted, but Pod exists!
```

**What you expected to happen**:

Atomic `HLEN RequestKey` and `DEL EngineKey`

sagearc:
closes #232 
Should fix CI tests

### DIFF
--- a/pkg/kvcache/kvblock/redis.go
+++ b/pkg/kvcache/kvblock/redis.go
@@ -143,6 +143,16 @@ type RedisIndex struct {
 
 var _ Index = &RedisIndex{}
 
+// pruneEngineKeyScript atomically verifies that a request key contains no pods, deleting the corresponding engine key if true.
+var pruneEngineKeyScript = redis.NewScript(`
+	local hashLen = redis.call('HLEN', KEYS[1])
+	if hashLen == 0 then
+		redis.call('DEL', KEYS[2])
+		return 1
+	end
+	return 0
+`)
+
 // Lookup receives a list of keys and a set of pod identifiers,
 // and retrieves the filtered pods associated with those keys.
 // The filtering is done based on the pod identifiers provided.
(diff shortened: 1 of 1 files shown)


sagearc:
@shmuelk You are right that `HDEL` handles the hash key (`RequestKey`) cleanup automatically. However, the issue here is cleaning up the index key (`EngineKey`) that points to it.

As shown below, we manage two separate keys: an index pointer and the actual data hash.

```
[ GLOBAL KEYSPACE ]
│
├── [ EngineKey X ] ---------------------------> [ RequestKey Y ]
│
│   ... (Empty Slots) ...
│
├── [ RequestKey Y ] ---------------------------> [ PODS ]
│                                                  {
│                                                    "pod-abc": "",
│                                                    "pod-xyz": ""
│                                                  }
│
└── ...
```

Since `EngineKey` and `RequestKey` are separate, we have to manually delete the index when the hash becomes empty. Without atomicity, we hit this race condition:
1. Thread **A** removes the last pod from `RequestKey`.
2. Thread **A** checks `HLEN` → returns 0 (empty).
3. Race Condition: Thread **B** pushes a **NEW** pod to `RequestKey`.
4. Thread **A** resumes and deletes EngineKey (based on the stale check in step 2).

Result: `EngineKey` is deleted, but `RequestKey` still exists and holds valid data. We have effectively orphaned the new pod because the lookup key is gone.

PR #243, opened 2026-01-05
feat: salt KV block hashes with LoRA adapter name

### LINKED ISSUE #167: [RFC] Lora Support

sagearc:
Previously, KV blocks were hashed using only the base model name, which could 
cause collisions between the base model and its lora adapters.

We now use the LoRA name (if present) to generate unique block hashes. 
This ensures cache isolation and completes the support needed for LoRA 
adapters.

closes #167

### DIFF
--- a/pkg/kvcache/kvevents/pool.go
+++ b/pkg/kvcache/kvevents/pool.go
@@ -218,6 +218,12 @@ func (p *Pool) digestEvents(ctx context.Context, podIdentifier, modelName string
 				deviceTier = strings.ToLower(*ev.Medium)
 			}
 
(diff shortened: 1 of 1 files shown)


vMaroon:
Awesome work @sagearc 

/lgtm
/approve

PR #234, opened 2025-12-27
Refactor: Replace transformers with vLLM

hyeongyun0916:
This PR refactors the tokenization system to use vLLM's tokenizer wrapper instead of the transformers library.

https://llm-d.slack.com/archives/C0A0SU5J68Y/p1764153758005369

PR #249, opened 2026-01-06
fix(cgo): resolve CStrings memory leaks

### LINKED ISSUE #248: CGO memory leaks
**Context**

From CGO docs:
> One important thing to remember is that `C.CString()` will allocate a new string of the appropriate length, and return it. That means **the C string is not going to be garbage collected and it is up to you to free it.**

ref: https://go.dev/wiki/cgo#go-strings-and-c-strings

**Leaks:**

- https://github.com/llm-d/llm-d-kv-cache/blob/6f627516023dd2c8f853cfe462913839439ce08f/pkg/preprocessing/chat_completions/cgo_functions.go#L151
- https://github.com/llm-d/llm-d-kv-cache/blob/6f627516023dd2c8f853cfe462913839439ce08f/pkg/preprocessing/chat_completions/cgo_functions.go#L185

sagearc:
closes #248

### DIFF
--- a/pkg/preprocessing/chat_completions/cgo_functions.go
+++ b/pkg/preprocessing/chat_completions/cgo_functions.go
@@ -148,7 +148,9 @@ func (w *ChatTemplatingProcessor) RenderChatTemplate(ctx context.Context,
 		return nil, fmt.Errorf("failed to marshal request: %w", err)
 	}
 	// Call the cached Python function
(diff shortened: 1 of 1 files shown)


hyeongyun0916:
This looks great. I think we should apply the same logic to [this](https://github.com/llm-d/llm-d-kv-cache/pull/234) as well.

hyeongyun0916 (approved):

PR #257, opened 2026-01-19
Align the interface of the uds tokenizer service with Tokenizer

### LINKED ISSUE #207: The uds tokenizer does not comply with the Tokenizer interface specification.
The tokenizer configuration currently used by the UDS tokenizer is read from the environment variables and initialized at program startup.
Reference:

https://github.com/llm-d/llm-d-kv-cache/blob/98db134628e59dfe527c40000f8d3829c002d99b/services/uds_tokenizer/server.py#L318C1-L354C1

However, the interface definition for the Tokenizer is

```golang
// Tokenizer interface defines the methods for tokenization.
type Tokenizer interface {
	RenderChatTemplate(string, *preprocessing.RenderJinjaTemplateRequest) (string, error)
	// Encode tokenizes the input string and returns the token IDs and offsets.
	Encode(input, modelName string) ([]uint32, []tokenizers.Offset, error)
	Type() string
}
```

This means that the tokenizer should read the `modelName` parameter passed in `Encode` or `RenderChatTemplate`, and dynamically select the appropriate tokenizer to use based on the modelName. Currently, the uds tokenizer does not comply with this requirement.

delavet:
fix #207 

This PR introduces two behavioral changes:
1. The UDS tokenizer service no longer identifies the model to load from its own environment variables. Instead, it provides an `InitializeTokenizer` method, which allows the service to dynamically load the tokenizer for a model based on the model name passed in. This also enables us to configure the model loaded by the tokenizer through a unified configuration.
2. The UDS tokenizer service now respects the `addSpecialTokens` parameter passed into the `Encode` method, rather than using a fixed value from the environment variables. This implementation aligns more closely with the definition of the `Tokenizer` interface.


services/uds_tokenizer/tokenizer_service/tokenizer.py
```diff
@@ -32,18 +32,24 @@
...
 
 
 class TokenizerService:
     """Service for handling tokenizer operations"""
 
-    def __init__(self, config: TokenizerConfig):
-        """Initialize service with configuration"""
-        self.tokenizer = self._create_tokenizer(config.model)
-        self.config = config
+    def __init__(self, config: TokenizerConfig = None):
```
vMaroon, line 42: Do you think this module can directly reuse the vLLM wrapping defined in the preprocessing package?

delavet: It is possible. I am actually preparing another PR to make the uds tokenizer use the vLLM wrapper. If you think it's OK, I can also add this directly to this PR.

delavet: Actually, I suggest preparing another PR to replace `uds_tokenizer` with the vLLM wrapper after the merge of #221. This is because, in the current main branch code, the vLLM wrapper only provides the `apply_chat_template` method so far. What do you think?

vMaroon: Sounds great!

PR #260, opened 2026-01-20
Fix: override block_size in storage backend

kfirtoledo:
Override block_size in the storage backend and not the offloading connector default.


dannyharnik (approved):
LGTM

Issue #263, opened 2026-01-21
Improve lookup in manager


rshavitt:
This issue addresses the task: "Parallelize lookup operations in the manager" mentioned in #235.

I investigated two potential approaches to improve the performance of lookup operations:

**1. Parallelization (Multi-threading)**
I attempted to parallelize the lookup operations using threads as originally suggested.
* **Result:** This approach actually **degraded performance**. The overhead of context switching and thread management outweighed the benefits for this specific operation structure.
* **Conclusion:** We should not proceed with threading for this specific lookup task.

**2. Optimizing `file_mapper` (Chosen Solution)**
I analyzed the bottleneck and found that refactoring `get_file_name` in the `file_mapper` offered significant gains without the complexity of concurrency.
* **Result:** This change improved lookup performance by approximately **3x** compared to the previous version.

Issue #235, opened 2025-12-28
FS Backend Improvements and Enhancements


kfirtoledo:
Several improvements can be made in the FS backend:
1. Evaluate performance improvements by allocating all threads to CPUs on the same NUMA node versus using all available CPUs.
2. Add a [safety mechanism](https://github.com/llm-d/llm-d-kv-cache/pull/176#discussion_r2625730055) to handle cases where two threads open the same folder at the same time.

kaushik-rohit:
@kfirtoledo Hi, what's left on this? and are you planning to address the remaining? if not I will like to take it

kfirtoledo:
llmd-fs-backend is now the FS tier of vLLM's multi-tier offloading connector (TieringOffloadingSpec). All new features and support continue there.

PR #264, opened 2026-01-21
Optimize get_file_name to improve lookup performance

rshavitt:
This PR addresses the performance of lookup operations.

Originally, we considered parallelizing these lookups (see issue #263, #235), but my testing showed that threading introduced overhead that slowed down the system.

Instead, this PR change the `get_file_name` function within `file_mapper` to work with strings instead of python Path, which improved the lookup performance to be ~3x faster.

### DIFF
--- a/kv_connectors/llmd_fs_backend/llmd_fs_backend/file_mapper.py
+++ b/kv_connectors/llmd_fs_backend/llmd_fs_backend/file_mapper.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from pathlib import Path
 
 
 class FileMapper:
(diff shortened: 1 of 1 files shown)


kv_connectors/llmd_fs_backend/llmd_fs_backend/file_mapper.py
```diff
@@ -83,6 +80,4 @@ def get_file_name(self, block_hash: int | bytes) -> str:
         assert isinstance(block_hash, int)
 
         block_hash_hex = f"{block_hash & ((1 << 64) - 1):016x}"
-        subfolder1, subfolder2 = block_hash_hex[:3], block_hash_hex[3:5]
-        full_path = self.base_path / subfolder1 / subfolder2 / f"{block_hash_hex}.bin"
-        return str(full_path)
+        return f"{self.base_path}/{block_hash_hex[:3]}/{block_hash_hex[3:5]}/{block_hash_hex}.bin"
```
kfirtoledo: Plead add end of file

PR #266, opened 2026-01-21
refactor: Accept tokens in GetPodScores

### LINKED ISSUE #244: Update GetPodScores of Indexer to get tokens as an input instead of preprocessing.RenderJinjaTemplateRequest
This change will help the pd profile handler to be able to make decision about disaggregation based on the precise prefix scorer's logic.
Tokenization will be done in the scorer's code before calling the GetPodScores function. 
Relevant [issue](https://github.com/llm-d/llm-d-inference-scheduler/issues/530) is in llm-d-inference-scheduler project.

acardace:
## Summary

Refactors `GetPodScores` to accept `tokens []uint32` instead of `renderReq` and `prompt` parameters.

- Adds `Tokenize` method to `Indexer` that returns `([]uint32, error)`
- Updates `GetPodScores` signature to accept `tokens []uint32` directly
- Propagates tokenization errors to callers instead of silently returning nil

Fixes #244

### DIFF
--- a/examples/kv_cache_aware_scorer/kvcache_aware_scorer.go
+++ b/examples/kv_cache_aware_scorer/kvcache_aware_scorer.go
@@ -214,7 +214,11 @@ func (s *PrecisePrefixCacheScorer) Score(ctx context.Context, cycleState *types.
 		return nil
 	}
 
-	tokens := s.kvCacheIndexer.Tokenize(nil, request.Prompt)
+	tokens, err := s.kvCacheIndexer.Tokenize(nil, request.Prompt)
+	if err != nil {
+		logger.Error(err, "Failed to tokenize prompt")
+		return nil
+	}
 	scores, err := s.kvCacheIndexer.GetPodScores(ctx, tokens, request.TargetModel, nil)
 	if err != nil {
 		logger.Error(err, "Failed to get pod scores")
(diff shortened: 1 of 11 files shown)


pkg/tokenization/pool.go
```diff
@@ -163,8 +164,10 @@ func (pool *Pool) Tokenize(renderReq *preprocessing.ApplyChatTemplateRequest, pr
 	})
 
 	res := <-resultCh
-	tokens := res.Tokens
-	return tokens
+	if res.Err != nil {
```
liu-cong, line 167: nit: I think `return res.Tokens, res.Err` should be enough

examples/kv_cache_aware_scorer/kvcache_aware_scorer.go
```diff
@@ -248,67 +251,3 @@ func (s *PrecisePrefixCacheScorer) Score(ctx context.Context, cycleState *types.
...
-		"isCompletions", request.Body != nil && request.Body.Completions != nil)
-
-	// The upstream parser guarantees exactly one body is populated, but we defensively prioritize chat completions.
-	// If an unexpected dual payload slips through (parser regression/new client), log it and use chat semantics.
-	if request.Body != nil && request.Body.ChatCompletions != nil {
-		if request.Body.Completions != nil {
-			traceLogger.Info("Both chat/completions and completions present; defaulting to chat/completions")
-		}
-
-		renderReq := &preprocessing.ApplyChatTemplateRequest{
```
liu-cong, line 270: It seems that the chat template is lost with this refactoring.

PR #279, opened 2026-02-02
fix: Fix CI wheels to support arm64

kfirtoledo:
This PR fixes the CI workflow to build and publish CUDA wheels on arm64 and amd64. 
The update moves arm64 builds into Docker, fixes CUDA environment handling, and ensures only newly built wheels are attached to releases.


Gregory-Pereira (approved):
/lgtm

Issue #191, opened 2025-11-25
[bug] tokenizer memory-leak on aggressive startup


vMaroon:
## Summary

In scenarios where the tokenizer is not loaded yet and input is received at high QPS, the tokenizer is loaded multiple times leading to potentially massive memory use.

PR #291, opened 2026-02-04
fix: Tokenizer initialization race condition (multiple parallel downloads)

albertoperdomo2:
## Summary

Under high QPS during cold start, multiple concurrent requests triggered duplicate tokenizer downloads due to a check-then-act race condition. It is mentioned in #191 and this caused:
  - Parallel downloads for a single model tokenizer
  - High request failure rate from file corruption
  - High memory usage (for parallel downloads)
 
The steps I followed to verify the condition with [meta-llama/Llama-3.2-1B](https://huggingface.co/meta-llama/Llama-3.2-1B) are:

1. Create venv with requirements

```
cd services/uds_tokenizer
python3 -m venv venv
source venv/bin/activate
pip install -r requirements.txt
pip install psutil  # for memory monitoring
```

2. Clear Tokenizer cache

```
rm -rf ~/.cache/huggingface/hub/models--*
rm -rf models/meta-llama
```

3. Ensure socket directory exists: `mkdir -p /tmp/tokenizer`

4. Start the gRPC service

```
python run_grpc_server.py
```

In a new terminal window, I ran a [script](https://gist.github.com/albertoperdomo2/0a52aef8eed9f8ad6d008fb53ac7e508) to verify the race condition existed. 

To solve it, I implemented double-checked locking with per-model locks in, with a fast path (cache hit) with zero locking overhead and a slow path (cache miss) using `threading.Lock` to ensure only one thread downloads each model.

## Test plan

The initial runs of the script (which launches 50 requests at the same time) reported:

```
============================================================
Memory Summary
============================================================
Initial:  75.7 MB
Peak:     494.5 MB
Final:    298.6 MB
Increase: 223.0 MB

Total test duration: 31.33s
```

with several errors in the service logs like:

```
2026-02-04 10:30:15,891 [ERROR] [root] Failed to load tokenizer from /<path>/<to>/llm-d-kv-cache/services/uds_tokenizer/models/meta-llama/Llama-3.2-1B: No such file or directory (os error 2)

2026-02-04 10:30:15,891 [ERROR] [root] Failed to initialize tokenizer for model meta-llama/Llama-3.2-1B: Failed to load tokenizer: No such file or directory (os error 2)

2026-02-04 10:30:20,153 [ERROR] [root] Failed to initialize tokenizer for model meta-llama/Llama-3.2-1B: Failed to load tokenizer: ...
```

And after the proposed fix:

```
============================================================
Memory Summary
============================================================
Initial:  75.3 MB
Peak:     173.7 MB
Final:    173.7 MB
Increase: 98.4 MB

Total test duration: 2.16s
```

Ensuring that the initial memory leak is gone, which directly impacts performance too. 

## Related issues

- #191

### DIFF
--- a/services/uds_tokenizer/tokenizer_service/tokenizer.py
+++ b/services/uds_tokenizer/tokenizer_service/tokenizer.py
@@ -21,6 +21,7 @@ from typing import Optional, List, Dict, Union
 from concurrent.futures import ThreadPoolExecutor
 import threading
 from collections import defaultdict
+
 from transformers import (AutoTokenizer, PreTrainedTokenizer, PreTrainedTokenizerFast)
 from transformers.tokenization_utils_base import BatchEncoding
 from modelscope import snapshot_download
@@ -45,7 +46,8 @@ class TokenizerService:
         """Initialize service with optional configuration"""
         self.tokenizers = {}  # Dictionary to store multiple tokenizers by model name
         self.configs = {}     # Dictionary to store configurations by model name
-        self._tokenizer_locks = defaultdict(threading.Lock)  # Per model locks
+        self._tokenizer_locks = defaultdict(threading.Lock)
+        self._locks_guard = threading.Lock()  # Protects access to _tokenizer_locks
 
         # If a config is provided, initialize the default tokenizer
         if config:
@@ -54,6 +56,14 @@ class TokenizerService:
             self.tokenizers[config.model] = self.tokenizer
             self.configs[config.model] = config
     
+    def _get_model_lock(self, model_name: str) -> threading.Lock:
+        """Get or create a per-model lock (thread-safe)."""
+        lock = self._tokenizer_locks.get(model_name)
+        if lock is not None:
+            return lock
+        with self._locks_guard:
+            return self._tokenizer_locks[model_name]
+
     def _create_tokenizer(self, model_identifier: str) -> AnyTokenizer:
         """Create a tokenizer, using cached files if available or downloading from ModelScope or Hugging Face"""
         # Check if the model_identifier is a remote model name or a local path
@@ -222,7 +232,7 @@ class TokenizerService:
             logging.info(f"Tokenizer for {model_name} already loaded")
             return True
 
-        lock = self._tokenizer_locks[model_name]
+        lock = self._get_model_lock(model_name)
         with lock:
             if model_name in self.tokenizers:
                 logging.info(f"Tokenizer for {model_name} already loaded")


pkg/preprocessing/chat_completions/tokenizer_wrapper.py
```diff
@@ -21,19 +21,34 @@
...
+    keys_to_lock = sorted(_cache_locks.keys())
+    acquired_locks = []
+    try:
+        for key in keys_to_lock:
+            lock = _cache_locks[key]
+            lock.acquire()
+            acquired_locks.append(lock)
+
+            _tokenizer_cache.clear()
+            return "Tokenizer caches cleared"
```
liu-cong: does this mean you will never have more than 1 item in acquired_locks given you return immediately after acquiring one?

pkg/preprocessing/chat_completions/tokenizer_wrapper.py
```diff
@@ -74,15 +89,24 @@ def get_or_create_tokenizer_key(request_json):
...
-        tokenizer = get_tokenizer(
-            model_name,
-            trust_remote_code=True,
-            revision=revision,
-            download_dir=download_dir,
-        )
-        _tokenizer_cache[key] = tokenizer
-        return key
+
+        lock = _cache_locks[key]
```
liu-cong: don't we need a lock to synchronize access to the `_cache_locks` dict?

pkg/preprocessing/chat_completions/tokenizer_wrapper.py
```diff
@@ -21,19 +21,34 @@
...
 _tokenizer_cache = {}
+_cache_locks = defaultdict(threading.Lock)
 
 
 def clear_caches():
     """Clear the tokenizer cache for testing purposes."""
-    _tokenizer_cache.clear()
-    return "Tokenizer caches cleared"
+    # Sorted locks to avoid deadlock
+    keys_to_lock = sorted(_cache_locks.keys())
```
liu-cong: It looks like we never insert any item to `_cache_locks` map (relying on defaultdict to get the lock). So this will always be empty?

pkg/preprocessing/chat_completions/tokenizer_wrapper.py
albertoperdomo2: This should be outside of the loop, my bad.

pkg/preprocessing/chat_completions/tokenizer_wrapper.py
albertoperdomo2: It is a `defaultdict()` so if you call it and the key does not exist, it will automatically create it on first access.

pkg/preprocessing/chat_completions/tokenizer_wrapper.py
albertoperdomo2: You are right, that would be necessary. This is primarily based on my approach to have per model locking, but if we do not need that, we can make it simpler with a single global lock.

services/uds_tokenizer/tokenizer_service/tokenizer.py
```diff
@@ -215,22 +218,31 @@ def _is_remote_model(self, model_identifier: str) -> bool:
...
-                add_generation_prompt=add_generation_prompt
-            )
+        if model_name in self.tokenizers:
+            logging.info(f"Tokenizer for {model_name} already loaded")
+            return True
 
-            tokenizer = self._create_tokenizer(model_name)
-            self.tokenizers[model_name] = tokenizer
-            self.configs[model_name] = config
+        lock = self._tokenizer_locks[model_name]
```
liu-cong: I think we need a `_cache_locks_guard` here as well.

albertoperdomo2: you are right 👍🏽

Issue #295, opened 2026-02-09
[RFC] Remove embedded tokenizers and enforce UDS Tokenizer usage


sagearc:
**Description**

Following the roadmap established in https://github.com/llm-d/llm-d-inference-scheduler/issues/552#issuecomment-3778097256
 and the recent updates in llm-d-kv-cache (specifically PR #278 #280), we are moving to a fully disaggregated tokenization architecture.

The goal of this issue is to remove all embedded python/cgo dependencies from the llm-d-inference-scheduler binary. It will rely exclusively on the external UDS tokenizer service for all tokenization and chat templating tasks.

**Proposed plan**

- [ ] **Remove Embedded Tokenizers:** Delete `CachedHFTokenizer`, `CachedLocalTokenizer`, `CompositeTokenizer`, and all `prefixstore` dependencies.
- [ ] **Remove Tokenization Pool:** Deprecate `tokenization.Pool` and its worker queue. The Indexer should invoke the UDS Tokenizer directly, relying on gRPC's native concurrency handling rather than a manual worker pool.
- [ ] **Update Testing Strategy:** Refactor tests to run against a mock UDS service or a sidecar binary, removing all tests that depend on local model loading or HF downloads.
- [ ] **Update Deployment Examples** in llm-d/llm-d, llm-d/llm-d-inference-scheduler

cc @vMaroon @hyeongyun0916 @delavet @elevran

vMaroon:
I think for initial stages, rather than remove the embedded modules, it would be good to isolate them in build through Go build tags and containing them in flags in the Makefile/Dockerfile - and deprecate them overtime.

delavet:
After fully introducing the vllm wrapper, the uds tokenizer might still require some refactoring to keep the code concise (the current code is influenced by the first version of the implementation and contains some redundant parts). I believe I can complete it within 1-2 weeks after the PR using the vllm wrapper is merged.

sagearc:
@hyeongyun0916 Not that I'm aware of. It might be worth running a quick benchmark to quantify that latency difference.

vMaroon:
> Do we have any foreseeable release plan to include this change? say, 0.6.0, maybe?

We're trying to push for this ASAP - independently of the llm-d v0.6.0 release timeline. I think if there won't be any surprises, then by a week it should be done? @sagearc @delavet

PR #302, opened 2026-02-11
refactor: extract shared tokenization types from CGO preprocessing package

sagearc:
Part of #295 - removing embedded python/cgo dependencies from the binary.

This PR extracts shared request/response types out of the CGO-dependent `chat_completions` package into a new standalone `types` package.

This is a prerequisite step: by decoupling these types from the CGO layer, consumers (UDS tokenizer, indexer, pool, prefixstore, examples, tests) no longer pull in the Python/C embedding dependency just to reference common structs.


sagearc:
contained in #303

Issue #286, opened 2026-02-04
FS Backend: Create priority in thread pools for reads


kfirtoledo:
In the current storage connector implementation, read and write operations share the same thread pool. This can cause read latency to increase when the pool is busy handling write tasks.
Add prioritization for read operations, for example, by introducing separate thread pools or a priority scheduling mechanism, so load operations are handled faster, and overall performance improves.

PR #337, opened 2026-02-17
feat: Threadpool priority

albertoperdomo2:
## Summary

The FS backend's storage connector uses a single thread pool for both read (GET: Storage->GPU) and write (PUT: GPU->Storage) operations. When write-heavy workloads fill the task queue, read operations (which are latency critical for serving inference requests)get stuck behind writes in a FIFO queue. 

This PR introduces a dual-queue with shared workers, so we keep a single pool of workers but maintain two internal queues: a high-priority read queue and a normal-priority write queue. Workers check the read queue first; if empty, they fall through to the write queue.

Other approaches such as separated `ThreadPool` instances are simpler but double the thread-local resource usage. With this approach, the thread-local resources (staging buffers, CUDA streams...) are not affected since we are only changing how tasks are queued and dequeued, not how they execute.

## Test plan

**(Edit)**

In order to have a sense of how this affects performance, three new test cases were added. 

1. Test that reads submitted after writes complete before them, by submitting 20 writes, wait for the queue to fill up, then submit 5 read operations and verify that reads complete early. 
2. Measure latency distribution under write pressure, by submitting writes continuously (cache evictions) and submitting intersperse reads (cache hits), then measuring the latency distribution. 
3. Test that checks starvation of writes, when a lot of reads are submitted, write should still be processed. 

The tests pass showing that prioritization is working as expected.

## Related issues

- #286

### DIFF
--- a/kv_connectors/llmd_fs_backend/tests/test_priority_queue.py
+++ b/kv_connectors/llmd_fs_backend/tests/test_priority_queue.py
@@ -539,28 +539,3 @@ def test_write_starvation_prevention(default_vllm_config):
         cleanup_files(file_mapper, write_hashes)
 
     del handler, put, get
-
-
-if __name__ == "__main__":
-    import sys
-
-    from vllm.config import VllmConfig, set_current_vllm_config
-
-    with set_current_vllm_config(VllmConfig()):
-        if len(sys.argv) > 1:
-            test_name = sys.argv[1]
-            if test_name == "order":
-                test_priority_completion_order()
-            elif test_name == "percentiles":
-                test_read_latency_percentiles()
-            elif test_name == "starvation":
-                test_write_starvation_prevention()
-            else:
-                print(f"Unknown test: {test_name}")
-                print("Available tests: order, percentiles, starvation")
-        else:
-            print("Running all priority queue tests...")
-            test_priority_completion_order()
-            test_read_latency_percentiles()
-            test_write_starvation_prevention()
-            print("\nAll tests passed!")


kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
```diff
@@ -42,8 +42,25 @@ thread_local at::cuda::CUDAStream ThreadPool::m_thread_stream =
...
+  // Initialize preferences
+  m_worker_preferences.reserve(threads);
+  for (size_t i = 0; i < threads; i++) {
+    WorkerPreference::Type pref = (i < num_read_first)
+                                          ? WorkerPreference::READ_FIRST
+                                          : WorkerPreference::WRITE_FIRST;
+    m_worker_preferences.push_back(pref);
+  }
+
+  FS_LOG_INFO("ThreadPool: "
```
kfirtoledo: Can you add also the total

kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.hpp
```diff
@@ -31,19 +31,31 @@
 #include "logger.hpp"
 #include "storage_types.hpp"
 
+enum class TaskPriority { kHigh = 0, kNormal = 1 };
+struct WorkerPreference {
+  enum Type {
+    READ_FIRST,  // Check high priority queue first, fall back to normal
```
kfirtoledo: Maybe this should be HIGH)FIRST and NORMAL_FIRST

kv_connectors/llmd_fs_backend/tests/test_fs_backend.py
```diff
@@ -324,3 +358,161 @@ def test_fs_backend_roundtrip_param(
...
+          f"max: {max(read_times):.3f}s)")
+    print(f"Total write time:  {total_write_time:.3f}s")
+
+    # Assertion: Average read latency should be reasonable
+    # Expected max avg read latency based on ratio:
+    # - 90% read-preferring: reads should avg <0.20s (most workers prioritize reads)
+    # - 75% read-preferring: reads should avg <0.25s (majority workers prioritize reads)
+    # - 50% read-preferring: reads should avg <0.35s (balanced, but still some priority)
+    if read_preferring_ratio >= 0.9:
+        max_avg_read = 0.30
```
kfirtoledo: Is there a better way to measure that? Because I am not sure it will be consistent across different environments.

kv_connectors/llmd_fs_backend/tests/test_fs_backend.py
```diff
@@ -324,3 +358,161 @@ def test_fs_backend_roundtrip_param(
         gpu_blocks_per_file=gpu_blocks_per_file,
         threads_per_gpu=threads_per_gpu,
     )
+
+
+@pytest.mark.parametrize("read_preferring_ratio", [0.5, 0.75, 0.9])
```
kfirtoledo: Can we move it to diffrent test file?

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
```diff
@@ -187,6 +188,7 @@ def __init__(
         gpu_blocks_per_file: int,
         threads_per_gpu: int,
         max_staging_memory_gb: int = DEFAULT_MAX_STAGING_MEMORY_GB,
+        read_preferring_ratio: float = DEFAULT_READ_PREFERRING_WORKERS_RATIO,
```
kfirtoledo, line 191: Can you make it configurable in spec.py?

kv_connectors/llmd_fs_backend/tests/test_priority_queue.py
```diff
@@ -0,0 +1,572 @@
...
+if __name__ == "__main__":
+    # Allow running individual tests
+    import sys
+
+    if len(sys.argv) > 1:
+        test_name = sys.argv[1]
+        if test_name == "order":
+            test_priority_completion_order()
+        elif test_name == "percentiles":
+            test_read_latency_percentiles()
```
kfirtoledo: This will work without default_vllm_config?

albertoperdomo2: yes, pytest discovers conftest.py automatically for all test files in the same directory, so there is no extra import needed.

kfirtoledo: I meant without the pytest

albertoperdomo2: you are right, now it explicitly creates the default vllm config 👍🏽

kfirtoledo (approved):

kv_connectors/llmd_fs_backend/tests/test_priority_queue.py
```diff
@@ -0,0 +1,396 @@
...
+    # Cleanup
+    cleanup_files(file_mapper, read_hashes)
+    for i in range(num_write_files):
+        _, write_hashes = make_storage_specs(1, start_offset=write_offset + i)
+        cleanup_files(file_mapper, write_hashes)
+
+    del handler, put, get
+
+
+def test_read_latency_percentiles(default_vllm_config):
```
kfirtoledo, line 249: Can we add a test that also checks that we don't have a starvation for writes

kfirtoledo (approved):

kv_connectors/llmd_fs_backend/tests/test_priority_queue.py
```diff
@@ -0,0 +1,566 @@
...
+
+    cleanup_files(file_mapper, read_hashes)
+    for i in range(num_writes):
+        _, write_hashes = make_storage_specs(1, start_offset=write_offset + i)
+        cleanup_files(file_mapper, write_hashes)
+
+    del handler, put, get
+
+
+if __name__ == "__main__":
```
kfirtoledo: I think we can remove the main because we will use only pytest

PR #342, opened 2026-02-22
build: Add pre-commit checks to CI

kfirtoledo:
Adds automated pre-commit hook validation to the CI pipeline, ensuring code quality standards are enforced before merging for Python, C++, CUDA, and GitHub Actions files.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,0 +1,22 @@
+pull_request_rules:
+  - name: comment-pre-commit-failure
+    description: Comment on PR when pre-commit check fails
+    conditions:
+      - check-failure=Run pre-commit hooks
+      - -closed
+      - -draft
+    actions:
+      comment:
+        message: |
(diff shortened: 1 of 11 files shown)


sagearc (changes requested):
Thanks! Since the scope is `--all-files`, please exclude this commit from git blame so it won't override our entire PRs history :)

sagearc:
Thanks for the PR! Just an FYI for the team, many leading OSS projects (examples below) use a specific standard to keep our git blame history clean, especially when pre-commit or formatters touch a lot of lines.

- vLLM:
  - https://github.com/vllm-project/vllm/blob/main/.git-blame-ignore-revs
- Django:
  - https://github.com/django/django/blob/main/.git-blame-ignore-revs
- Pytorch:
  - https://github.com/pytorch/pytorch/blob/main/.git-blame-ignore-revs

I’ll open a follow-up PR to set this up for us so we can maintain meaningful blame history as we grow.

@vMaroon @kfirtoledo @nilig

PR #346, opened 2026-02-23
events: add extra_keys to BlockStored event 

RishabhSaini:
https://github.com/vllm-project/vllm/pull/33304 introduces this extra field

### DIFF
--- a/pkg/kvevents/engineadapter/vllm_adapter.go
+++ b/pkg/kvevents/engineadapter/vllm_adapter.go
@@ -137,6 +137,7 @@ type msgpackVLLMBlockStoredEvent struct {
 	LoraID          *int    `msgpack:",omitempty"`
 	Medium          *string `msgpack:",omitempty"`
 	LoraName        *string `msgpack:",omitempty"`
+	ExtraKeys       []any   `msgpack:",omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


pkg/kvevents/process_event_test.go
```diff
@@ -63,6 +63,7 @@ func TestBlockStoredAllFieldsPresent(t *testing.T) {
 		42,                                // LoraID
 		"GPU",                             // Medium
 		"test-lora",                       // LoraName
+		nil,                               // ExtraKeys (optional, set to nil)
```
yankay: Thanks @RishabhSaini, 

A Test can add, like:
https://github.com/yankay/llm-d-kv-cache/commit/26f27694527dd8349e0978483dc8cbd7b4725052#diff-bb18bd9fc5e2c6d3c0242a3098bb74aa3ad326efdce33da6af50169a65b5ead9R90

```
	require.NotNil(t, blockStored.ExtraKeys, "Expected ExtraKeys to be non-nil")
	require.Equal(t, expectedExtraKeys, blockStored.ExtraKeys, "Expected ExtraKeys to match the initialized hash payload")
```

RishabhSaini: agreed, updated

Issue #344, opened 2026-02-23
Remove exclusion of the services folder from pre-commit checks


kfirtoledo:
In PR #342, the pre-commit configuration excludes the services folder from pre-commit checks. This exclusion should be removed from .pre-commit-config.yaml so that checks run on all files in the repository.

sagearc:
ref: https://github.com/llm-d/llm-d-kv-cache/pull/342#issuecomment-3950780861

PR #362, opened 2026-02-26
Remove services/ exclusion from pre-commit checks

yankay:
This PR removes the exclude: '^services/'directive from .pre-commit-config.yaml, enabling pre-commit hooks (ruff linting and formatting) to run on the services directory.

fix: #344


.pre-commit-config.yaml
kfirtoledo: @yankay, can you also remove this line? It was added by mistake

yankay: OK, It has been changed

Issue #353, opened 2026-02-24
Speculative Indexing for the KV-Cache Aware Routing


vMaroon:
## Summary

Details in [google doc](https://docs.google.com/document/d/1Ao-XYDFBXe_kOgVWFzNR38P5LqRtF7lShPMlC9x53AQ/edit?tab=t.0).

TL;DR: Feed routing decisions into the KV-cache index as speculative block entries alongside engine-confirmed blocks. A side TTL cache expires unconfirmed entries via `Index.Evict()` callbacks.
Block annotations distinguish speculative from confirmed, generalizing to future use cases (HMA full-attention vs SWA). The `PrecisePrefixCacheScorer` gains `PrepareDataPlugin` and `PreRequest` hooks, aligning with the gateway plugin framework. This functionally unifies the IGW approximate scorer and the llm-d precise scorer.

Issue #336, opened 2026-02-17
Support HMA (Hybrid Memory Attention) in Prefix-Cache Aware Scheduling


vMaroon:
## Overview

Prefix-cache-aware routing currently assumes uniform attention across all layers - a block is either cached or not. With HMA (hybrid models mixing full-attention and sliding-window layers), this breaks down:

- vLLM can evict SWA blocks independently of full-attention blocks, creating partial hits: full-attention prefix is cached but SWA blocks for that range are gone
- A partial hit is not a miss. The last N blocks (where N = window size in blocks) need SWA state, but earlier blocks don't. The indexer must distinguish between full hits, partial hits (full-attention present, SWA evicted outside the window), and misses
- The `BlockStored`/`BlockRemoved` event format and the indexer's scoring logic are unaware of per-group block identity, so the scheduler cannot reason about this today

## Proposed Direction

- Keep `BlockStored` as-is - a stored block implies all layer groups are cached (full-attention + SWA)
- Extend `BlockRemoved` to indicate eviction type: full (all layers, equivalent to today's behavior) vs partial (SWA layers only, full-attention retained). This avoids per-group tagging on every block hash
- The indexer/scorer needs the model's SWA window size to determine whether a partial hit qualifies as a routable hit
- Per-layer heterogeneous window sizes don't exist in production models today — defer support, but keep the format extensible
- Upstream dependency: vLLM KVEvents emission needs to be HMA-aware ([ref](https://github.com/vllm-project/vllm/blob/192ad4648b2066ebdf1fa04ad84f24bdf0cd6533/vllm/config/vllm.py#L1004-L1006))

This was identified by @orozery.

vMaroon:
Should align with #353

zhouyou9505:
Hi, I’d like to take this issue if it’s currently unblocked? @vMaroon

sagearc:
I think we can keep this simpler if we separate the architectural model from the implementation detail.

The design I would suggest is **one index plus group presence**, instead of one index per attention group.

TLDR:
- one kv blocks index
- learned group metadata at runtime from received events
- per-block group presence

vLLM already gives us the pieces we need:

- `BlockStored` is self-describing. It carries both the structural group identity (`group_idx`) and the semantic group metadata.
- `BlockRemoved` is intentionally smaller and contains only the `group_idx`.

HMA support only needs the following additional information:

1. A small group metadata catalog, learned from received `BlockStored` at runtime:

```text
pod A:
[...]

PR #369, opened 2026-02-27
feat: add speculative indexing support for PodEntry and Indexer

bongwoobak:
Part of https://github.com/llm-d/llm-d-kv-cache/issues/353

## Summary
- Add `Annotation` field to `PodEntry` to distinguish speculative vs confirmed entries
- Extract `ComputeBlockKeys()` method from `Indexer` for reuse by PrepareDataPlugin
- Support nil `engineKeys` in `InMemoryIndex.Add()/Evict()` for speculative entries without engine key mappings

## Details

Part 1 of Speculative Indexing (https://github.com/llm-d/llm-d-kv-cache/issues/353).

Speculative indexing solves the "blind spot" between routing decision and KV event arrival. When the scheduler picks a pod, it immediately adds speculative entries to the Index so subsequent requests with the same prefix can be routed to the same pod, before the engine confirms via KV events.

### PodEntry Annotation
- New `Annotation string` field (comparable, LRU-safe)
- Speculative: `Annotation: "speculative"`, Confirmed: `Annotation: ""`
- Same pod can have both entry types simultaneously (different LRU keys)
- Evicting speculative doesn't affect confirmed entries

### ComputeBlockKeys
- Extracted tokenize --> truncation --> block hash logic from `GetPodScores()`
- Enables PrepareDataPlugin to compute block keys once and reuse across Score/PreRequest
- `GetPodScores()` refactored to delegate to this method

### Nil engineKeys Support
- `Add()` accepts nil engineKeys, skips engine --> request key mapping
- `Evict()` falls back to treating key as requestKey when no mapping exists

## Test plan
- `TestSpeculativeAnnotation`: speculative + confirmed coexistence, eviction isolation
- `TestSpeculativeEvictThenEmpty`: cleanup after sole speculative eviction
- `TestAddWithNilEngineKeys`: nil engineKeys handling
- `TestPodEntryString`: string representation with/without annotation
- `TestComputeBlockKeys`: deterministic block key computation
- `TestComputeBlockKeysConsistentWithGetPodScores`: consistency with scoring
- `TestComputeBlockKeysEmptyPrompt`: edge case

### DIFF
--- a/pkg/kvcache/indexer.go
+++ b/pkg/kvcache/indexer.go
@@ -221,6 +221,22 @@ func (k *Indexer) ScoreTokens(
 	traceLogger.Info("found block keys", "block-keys", blockKeys,
 		"pods", podsPerKeyPrintHelper(keyToPods))
 
+	// Calculate block-level hit ratio (blocks found / blocks requested).
+	blocksFound := 0
+	for _, pods := range keyToPods {
+		if len(pods) > 0 {
+			blocksFound++
+		}
+	}
+	blockHitRatio := 0.0
+	if len(blockKeys) > 0 {
+		blockHitRatio = float64(blocksFound) / float64(len(blockKeys))
+	}
+	span.SetAttributes(
+		attribute.Float64("llm_d.kv_cache.block_hit_ratio", blockHitRatio),
+		attribute.Int("llm_d.kv_cache.blocks_found", blocksFound),
+	)
+
 	podScores, err := k.kvBlockScorer.Score(ctx, blockKeys, keyToPods)
 	if err != nil {
 		span.SetStatus(codes.Error, err.Error())


pkg/kvcache/indexer_test.go
```diff
@@ -0,0 +1,133 @@
+//go:build embedded_tokenizers
+
+/*
+Copyright 2025 The llm-d Authors.
```
vMaroon: s/2025/2026/.

pkg/kvcache/indexer.go
```diff
@@ -123,6 +123,36 @@ func (k *Indexer) KVBlockIndex() kvblock.Index {
 	return k.kvBlockIndex
 }
 
+// ComputeBlockKeys computes the KV-block keys for a given prompt and model name.
+// This method extracts the tokenization and block key computation logic so that
+// callers (e.g., PrepareDataPlugin) can compute block keys once and reuse them
```
vMaroon: maybe `IGW::EPP::PrepareDataPlugin` for better context?

pkg/kvcache/kvblock/index.go
```diff
@@ -152,9 +152,17 @@ type PodEntry struct {
 	PodIdentifier string
 	// DeviceTier is the tier of the device where the KV-block is stored.
 	DeviceTier string
+	// Annotation is an optional label for the entry (e.g., "speculative").
+	// An empty string represents a confirmed (default) entry.
+	// Because PodEntry is used as a comparable LRU cache key, this is a string
+	// rather than a map to preserve comparability.
+	Annotation string
```
vMaroon: Think this should be a map of annotations (or equivalent). An Immediate extension would be to track block type per entry for #336 (partial/full).

bongwoobak: Instead of a map (which would break LRU cache comparability), I'll change Annotation string to an Annotations struct:

```
type Annotations struct {
    Source    string // "speculative" or "" (confirmed)
    BlockType string // for #336: "partial", "full", or ""
}

type PodEntry struct {
    PodIdentifier string
    DeviceTier    string
    Annotations   Annotations
}
```

pkg/kvcache/kvblock/index.go
```diff
@@ -147,15 +148,39 @@ func (c BlockHash) String() string {
...
+	if e.Annotations.Source != "" || e.Annotations.BlockType != "" {
+		parts := []string{}
+		if e.Annotations.Source != "" {
+			parts = append(parts, e.Annotations.Source)
+		}
+		if e.Annotations.BlockType != "" {
+			parts = append(parts, e.Annotations.BlockType)
+		}
+		suffix = "[" + strings.Join(parts, ",") + "]"
+	}
```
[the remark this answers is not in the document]
bongwoobak: Source and BlockType are constrained enums ("speculative", "partial", "full"), not free-form strings. So no collision risk in practice. Will revisit if the field set grows.

pkg/kvcache/kvblock/in_memory.go
```diff
@@ -218,15 +223,21 @@ func (m *InMemoryIndex) Evict(ctx context.Context, engineKey BlockHash, entries
 	traceLogger := log.FromContext(ctx).V(logging.TRACE).WithName("kvblock.InMemoryIndex.Evict")
 
 	requestKey, found := m.engineToRequestKeys.Get(engineKey)
+	hasEngineKeyMapping := found
 	if !found {
-		traceLogger.Info("engineKey not found in index, nothing to evict", "engineKey", engineKey)
-		return nil
+		// Fallback: treat engineKey as requestKey directly.
+		// This is used for speculative entries added without engineKey mapping.
```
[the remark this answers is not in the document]
bongwoobak: The fallback only applies to speculative entries added with nil engineKeys. Eviction uses the exact PodEntry including Annotations{Source: "speculative"}, so confirmed entries are never affected.

pkg/kvcache/kvblock/index.go
```diff
@@ -147,15 +148,39 @@ func (c BlockHash) String() string {
...
 }
 
 // String returns a string representation of the PodEntry.
 func (e *PodEntry) String() string {
-	return fmt.Sprintf("%s@%s", e.PodIdentifier, e.DeviceTier)
+	suffix := ""
+	if e.Annotations.Source != "" || e.Annotations.BlockType != "" {
+		parts := []string{}
+		if e.Annotations.Source != "" {
+			parts = append(parts, e.Annotations.Source)
```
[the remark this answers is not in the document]
bongwoobak: Updated RedisIndex to handle annotated PodEntries: nil engineKeys in Add, fallback in Evict, annotation stripping in Lookup.

pkg/kvcache/kvblock/in_memory.go
```diff
@@ -147,21 +147,23 @@ func (m *InMemoryIndex) Lookup(ctx context.Context, requestKeys []BlockHash,
 }
 
 // Add adds a set of engineKeys/requestKeys and their associated pod entries to the index backend.
+// If engineKeys is nil, only requestKey -> PodEntry mappings are created (no engineKey -> requestKey mapping).
+// This is used for speculative entries where engine keys are not yet known.
 func (m *InMemoryIndex) Add(ctx context.Context, engineKeys, requestKeys []BlockHash, entries []PodEntry) error {
```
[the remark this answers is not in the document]
bongwoobak: Updated RedisIndex to support nil engineKeys in Add, fallback in Evict, and annotation stripping in Lookup.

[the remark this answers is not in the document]
vMaroon: This is a good point, there is also the cost-aware index. Perhaps this should be modeled explicitly as a new field?

pkg/kvcache/kvblock/in_memory.go
```diff
@@ -203,13 +205,16 @@ func (m *InMemoryIndex) Add(ctx context.Context, engineKeys, requestKeys []Block
...
 
 	return nil
 }
 
-// Evict removes a engineKey and its associated pod entries from the index backend.
+// Evict removes an engineKey and its associated pod entries from the index backend.
+// If the engineKey is not found in the engineToRequestKeys mapping, it falls back to
+// treating the key as a requestKey directly. This supports eviction of speculative entries
+// that were added without engineKey mapping (nil engineKeys in Add).
 func (m *InMemoryIndex) Evict(ctx context.Context, engineKey BlockHash, entries []PodEntry) error {
```
vMaroon: Similarly to the above, we can add a new field that sets whether it is an engineKey or requestKey (the key is just key, and an additional field).

pkg/kvcache/kvblock/index.go
```diff
@@ -147,15 +148,39 @@ func (c BlockHash) String() string {
...
+}
+
 // PodEntry struct represents a pod entry in the KV-block index.
 type PodEntry struct {
 	// PodIdentifier is the unique identifier for the pod.
 	PodIdentifier string
 	// DeviceTier is the tier of the device where the KV-block is stored.
 	DeviceTier string
+	// Annotations holds optional metadata for the entry.
+	Annotations Annotations
```
vMaroon: Looking at this now - this choice was premature. We don't know whether a single `BlockType` will suffice for HMA, as it seems like heterogenous SWA window-size models exist already.

I think we can tie the speculative-property to identity, and just make it a boolean. What do you think?

bongwoobak: Done. Simplified Annotations struct to Speculative bool on PodEntry.

pkg/kvcache/kvblock/in_memory.go
bongwoobak: Done. Added KeyType to Evict() interface.

pkg/kvcache/indexer.go
```diff
@@ -166,23 +170,27 @@ func (k *Indexer) ScoreTokens(
...
 
-	traceLogger := log.FromContext(ctx).V(logging.TRACE).WithName("kvcache.ScoreTokens")
-
-	blockKeys := k.tokenProcessor.TokensToKVBlockKeys(kvblock.EmptyBlockHash, tokens, modelName)
-
 	// Set initial attributes
 	span.SetAttributes(
 		attribute.String("gen_ai.request.model", modelName),
 		attribute.Int("llm_d.kv_cache.pod_count", len(podIdentifiers)),
-		attribute.Int("llm_d.kv_cache.token_count", len(tokens)),
```
vMaroon, line 177: Lets keep this.

bongwoobak: `ScoreTokens` already has this attribute, so it's preserved through the delegation. No information loss.

pkg/kvcache/indexer.go
```diff
@@ -170,52 +199,30 @@ func (k *Indexer) ScoreTokens(
...
 	keyToPods, err := k.kvBlockIndex.Lookup(ctx, blockKeys, sets.New(podIdentifiers...))
 	if err != nil {
 		span.SetStatus(codes.Error, err.Error())
 		return nil, fmt.Errorf("failed to query kvblock indexer: %w", err)
 	}
-	traceLogger.Info("found block keys", "block-keys", blockKeys,
-		"pods", podsPerKeyPrintHelper(keyToPods))
-
-	// Calculate block-level hit ratio (blocks found / blocks requested).
-	blocksFound := 0
```
vMaroon, line 197: Wes this trace block accidentally removed?

bongwoobak: Thank you for catching, I restored them.

PR #370, opened 2026-02-28
Add DP-aware routing support to KVEvents and indexing pipeline

### LINKED ISSUE #357: Ensure DP-aware routing works with KVEvents/indexing
## Summary

In DP deployments (data-parallel), different ranks maintain their own KV-cache. Ranks could be deployed in separate vLLM processes (different API servers/frontends), or within the same vLLM frontend.

KVEvents have a `data_parallel_rank` field that is sent in every message but is currently ignored. While this makes sense in the latter kind of deployments, within the first, the DP rank can be assigned on the connection/subscription level.

This gap should be closed in accommodation with the DP-aware wide-ep work.

satyamg1620:
### Summary

- Propagate DataParallelRank from EventBatch through the KV event processing pipeline into PodEntry, enabling the index and scorer to distinguish KV blocks cached by different data-parallel ranks on the same pod.
- Add DataParallelRank int field to PodEntry with sentinel value -1 (NoDataParallelRank) for backward compatibility with non-DP deployments.
- Update LongestPrefixScorer to produce DP-aware scoring keys ("pod-1@dp0") so different ranks on the same pod receive independent scores.
- Add optional int32 data_parallel_rank field to the PodScore proto message and update the gRPC server to populate it.

Fixes: #357 

Follow up PR: https://github.com/llm-d/llm-d-inference-scheduler/pull/750

### DIFF
--- a/pkg/kvcache/kvblock/index.go
+++ b/pkg/kvcache/kvblock/index.go
@@ -20,10 +20,18 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
-	"github.com/llm-d/llm-d-kv-cache/pkg/kvcache/metrics"
 	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/llm-d/llm-d-kv-cache/pkg/kvcache/metrics"
+)
+
+const (
+	// NoDataParallelRank indicates that no data parallel rank is set.
+	// This is the default value for non-DP deployments.
+	NoDataParallelRank = -1
 )
 
 // IndexConfig holds the configuration for the KV-block index.
@@ -184,9 +192,31 @@ type PodEntry struct {
 	HasGroup bool
 	// GroupIdx identifies the vLLM KV cache group for HMA events.
 	GroupIdx GroupID
+	// DataParallelRank is the data parallel rank of the pod.
+	// A value of NoDataParallelRank (-1) indicates no DP rank is set (non-DP deployment).
+	DataParallelRank int
+}
+
+// NewPodEntry creates a PodEntry, converting a *int DP rank to the int sentinel form.
+// A nil dpRank is stored as NoDataParallelRank (-1). A negative dpRank is also
+// treated as NoDataParallelRank: valid DP ranks are non-negative, and storing
+// a negative value would break roundtripping through ParsePodEntry (which
+// rejects them) and cause the pod to drop out of lookups.
+func NewPodEntry(podIdentifier, deviceTier string, dpRank *int) PodEntry {
+	rank := NoDataParallelRank
+	if dpRank != nil && *dpRank >= 0 {
+		rank = *dpRank
+	}
+	return PodEntry{
+		PodIdentifier:    podIdentifier,
+		DeviceTier:       deviceTier,
+		DataParallelRank: rank,
+	}
 }
 
 // String returns a string representation of the PodEntry.
+// Format: "pod@tier" (no DP rank) or "pod@tier@dpN" (with DP rank).
+// The [speculative] annotation, if present, is always appended last.
 func (e *PodEntry) String() string {
 	suffix := ""
 	if e.Speculative {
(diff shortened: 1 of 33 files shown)


vMaroon:
@satyamg1620 thank you for the contribution!

Generally I think the approach is correct - this is the only way to currently cover all deployments mentioned in https://docs.vllm.ai/en/stable/serving/data_parallel_deployment. Today in llm-d we only support the external LB mode, in which every rank is a separate deployment, and the pipeline works as follows:
1. Each rank publishes to a `kv@<IP>:<PORT>@<MODEL>` topic, and a `PodIdentifier` is then <IP>:<PORT>
2. The scheduler treats each rank as a separate, normal endpoint, identified by <IP>:<PORT>

Though for other DP modes, there is one gap on consumption from the scheduler side: the scheduler currently has nothing that connects actual DP-rank information with the port. 

The bridge between the scheduler and the indexed data is missing here. It would be great if you can prepare an overview of how llm-d would support each of the modes in the vllm doc. Once a plan is conceived on this kind of coverage - we can proceed in a phased approach where this PR is one, and scheduler updates are another.

pkg/kvcache/kvblock_scorer.go
```diff
@@ -96,13 +108,15 @@ func fillMaxWeights(dst map[string]float64, entries []kvblock.PodEntry, mediumWe
 				weight = w
 			}
 		}
-		if cur, exists := dst[entry.PodIdentifier]; !exists || weight > cur {
-			dst[entry.PodIdentifier] = weight
+		if cur, exists := dst[podScoringKey(entry)]; !exists || weight > cur {
+			dst[podScoringKey(entry)] = weight
```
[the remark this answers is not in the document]
gyliu513: +1

gyliu513 (commented):
nit: https://github.com/llm-d/llm-d-kv-cache/blob/main/tests/e2e/uds_tokenizer/uds_e2e_mm_test.go#L297 also need update

pkg/kvcache/kvblock_scorer.go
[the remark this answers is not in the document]
satyamg1620: ok @gyliu513 , i will make the corresponding changes

gyliu513:
looks good, thanks @satyamg1620 . I think all the comments from Copilot are also valid, can you help address those comments? Thanks!

satyamg1620:
> looks good, thanks @satyamg1620 . I think all the comments from Copilot are also valid, can you help address those comments? Thanks!

Sure @gyliu513 , I will address those comments. Let me know if any other changes are required.

satyamg1620:
@vMaroon @liu-cong  @dannyharnik  @kfirtoledo @gyliu513 , can you please review this PR?

satyamg1620:
@vMaroon @liu-cong @dannyharnik @kfirtoledo @gyliu513 , can you please review this PR?

liu-cong:
Apologies for the delay. I just came back from a long leave. I'll try to get to this as soon as possible.

satyamg1620:
> Apologies for the delay. I just came back from a long leave. I'll try to get to this as soon as possible.

Thanks @liu-cong , let me know, if any changes required

satyamg1620:
@vMaroon @liu-cong @dannyharnik @kfirtoledo @gyliu513 , can you please review this PR?

pkg/kvcache/kvblock/index.go
```diff
@@ -180,13 +188,106 @@ type PodEntry struct {
 	DeviceTier string
 	// Speculative indicates the entry was added predictively before a KV event confirmed it.
 	Speculative bool
+	// DataParallelRank is the data parallel rank of the pod.
+	// A value of NoDataParallelRank (-1) indicates no DP rank is set (non-DP deployment).
+	DataParallelRank int
```
liu-cong: Why not using *int to be consistent with the proto definition (nil means no DP rank).

pkg/kvcache/kvblock_scorer.go
```diff
@@ -51,7 +53,9 @@ type KVBlockScorer interface {
 	// Strategy returns the scoring strategy type.
 	Strategy() KVScoringStrategy
 	// Score scores the blocks based on the scoring strategy.
-	// It returns a map of pod names to their scores.
+	// The returned map is keyed by pod scoring key (not the raw pod identifier):
+	// for non-DP pods the key is the pod identifier; for DP-aware pods the key is
+	// "<pod>@dp<rank>". Use ParsePodScoringKey to decompose these keys.
 	Score(ctx context.Context, keys []kvblock.BlockHash,
```
liu-cong, line 62: We are doing string - PodEntry parsing back and forth, we should just change this interface to return map[PodEntry]float64 instead and avoid the parsing. Can you try this?

satyamg1620: Ok @liu-cong. I will try this change

satyamg1620: @liu-cong I have made the changes accordingly

satyamg1620:
@vMaroon @dannyharnik @kfirtoledo @liu-cong Can you please review this PR, if possible

PR #387, opened 2026-03-04
perf: Cache CBOR EncMode in token processor to avoid per-hash allocation

gyliu513:
## Summary

Cache `cbor.CanonicalEncOptions().EncMode()` once at `chunkedTokenDatabase` construction instead of recreating it on every `hash()` call

## Why

The `hash()` method is on the hot path, it is called once per token block chunk for every incoming request during prefix cache lookup. Previously, each call to `hash()` invoked `cbor.CanonicalEncOptions().EncMode()` to create a new CBOR encoder, which involves allocating internal encoding state and validating options every time. This is entirely redundant because `EncMode` is immutable and thread-safe once created.

## How it helps

**Eliminates redundant allocations**: Removes one `EncMode` object creation per `hash()` call. For a typical request with N token blocks, this saves N allocations.

### DIFF
--- a/pkg/kvcache/kvblock/token_processor.go
+++ b/pkg/kvcache/kvblock/token_processor.go
@@ -63,7 +63,7 @@ type TokenProcessor interface {
 // It mimics the chunkedTokenDatabase in the Python code.
 type chunkedTokenDatabase struct {
 	TokenProcessorConfig
-	encMode cbor.EncMode // cached CBOR encoder for deterministic encoding
+	encoder cbor.EncMode // cached CBOR encoder for interoperable encoding
 }
 
 var _ TokenProcessor = &chunkedTokenDatabase{}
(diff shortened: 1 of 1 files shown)


pkg/kvcache/kvblock/token_processor.go
```diff
@@ -63,6 +63,7 @@ type TokenProcessor interface {
 // It mimics the chunkedTokenDatabase in the Python code.
 type chunkedTokenDatabase struct {
 	TokenProcessorConfig
+	encMode cbor.EncMode // cached CBOR encoder for deterministic encoding
```
vMaroon: serializer/encoder might be a clearer name. The comment could be tightened by replacing "deterministic" with "interoperable".

gyliu513: Done, thanks @vMaroon

PR #415, opened 2026-03-10
Add ScoreTokens for external tokenization support

acardace:
## Summary
- Export `GetPodScoresFromTokens` on the `Indexer` to allow callers with pre-tokenized input to obtain KV-cache prefix scores directly, bypassing the internal tokenizers pool
- Extract a `TokenizersPool` interface from the concrete `*tokenization.Pool` type for testability
- Fix the UDS tokenizer health probe endpoint path from `/health` to `/healthz`
- Add unit and e2e test coverage for both `GetPodScores` and `GetPodScoresFromTokens`

## Motivation
This is the llm-d-kv-cache counterpart to the external tokenization support being added in [gateway-api-inference-extension#2542](https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2542). The design ([doc](https://docs.google.com/document/d/1nXeWJZQlzMY43t_Yn5FJran_EhSE_z1n--KCk9cxxjk/edit?usp=sharing)) introduces a `TokenizedPrompt` field on `LLMRequest` so that an external tokenizer plugin can populate token IDs before scheduling. On the llm-d side, the `PrecisePrefixCacheScorer` needs a way to consume those pre-tokenized inputs without re-tokenizing — `GetPodScoresFromTokens` provides that entry point.

## Changes
| Commit | Description |
|--------|-------------|
| `Add GetPodScoresFromTokens` | New exported method on `Indexer` that accepts `[]uint32` tokens directly |
| `Fix health probe endpoint path to /healthz` | Corrects the UDS tokenizer gRPC server health endpoint |
| `Extract TokenizersPool interface for testability` | Replaces concrete `*tokenization.Pool` field with an interface to enable mocking in tests |
| `Add unit and e2e tests for GetPodScores and GetPodScoresFromTokens` | 21 unit tests (table-driven, covering scoring, prefix break, device tiers, filtering, truncation) + 4 e2e tests (cache hit/miss, consistency between methods, prefix reduction) |

### DIFF
--- a/pkg/kvcache/indexer_test.go
+++ b/pkg/kvcache/indexer_test.go
@@ -1,0 +1,366 @@
+/*
+Copyright 2026 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kvcache_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/llm-d/llm-d-kv-cache/pkg/kvcache"
+	"github.com/llm-d/llm-d-kv-cache/pkg/kvcache/kvblock"
+	"github.com/llm-d/llm-d-kv-cache/pkg/tokenization"
+	"github.com/llm-d/llm-d-kv-cache/pkg/tokenization/types"
+	"github.com/llm-d/llm-d-kv-cache/pkg/utils/logging"
+)
+
+// --- mock implementations ---------------------------------------------------
+
+// mockTokenProcessor implements kvblock.TokenProcessor for testing.
+// It records the tokens it receives so tests can assert on them.
+type mockTokenProcessor struct {
+	blockKeys      []kvblock.BlockHash
+	receivedTokens []uint32
+}
+
+func (m *mockTokenProcessor) TokensToKVBlockKeys(_ kvblock.BlockHash, tokens []uint32, _ string) []kvblock.BlockHash {
+	m.receivedTokens = tokens
+	return m.blockKeys
+}
+
+// mockTokenizersPool implements kvcache.TokenizersPool for testing.
+type mockTokenizersPool struct {
+	tokens []uint32
+}
+
+func (m *mockTokenizersPool) Tokenize(_ *types.RenderChatRequest, _ string) []uint32 {
+	return m.tokens
+}
+
+func (m *mockTokenizersPool) Run(_ context.Context) {}
+
+func (m *mockTokenizersPool) SetTokenizer(_ tokenization.Tokenizer, _ string) {}
+
+// --- helpers ----------------------------------------------------------------
+
+const (
+	testModel = "test-model"
+	testPodA  = "pod-a"
+	testPodB  = "pod-b"
+)
+
+func u64ToBlockKeys(keys []uint64) []kvblock.BlockHash {
+	out := make([]kvblock.BlockHash, len(keys))
+	for i, k := range keys {
+		out[i] = kvblock.BlockHash(k)
+	}
+	return out
+}
+
+// newTestIndexer creates an Indexer backed by an in-memory index, a mock
+// tokenizers pool, and a LongestPrefixScorer using the project's default
+// backend weights.
+func newTestIndexer(t *testing.T, tp kvblock.TokenProcessor, pool kvcache.TokenizersPool) *kvcache.Indexer {
+	t.Helper()
+
+	idx, err := kvblock.NewInMemoryIndex(kvblock.DefaultInMemoryIndexConfig())
+	require.NoError(t, err)
+
+	scorer, err := kvcache.NewKVBlockScorer(kvcache.DefaultKVBlockScorerConfig())
+	require.NoError(t, err)
+
+	return kvcache.NewIndexerForTest(tp, idx, scorer, pool)
+}
+
+// populateIndex inserts block-key -> pod entries into the index.
+func populateIndex(t *testing.T, idx kvblock.Index, entries map[kvblock.BlockHash][]kvblock.PodEntry) {
+	t.Helper()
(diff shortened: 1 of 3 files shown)


services/uds_tokenizer/run_grpc_server.py
```diff
@@ -105,7 +105,7 @@ async def health_handler(request):
 def create_probe_app():
     """Create aiohttp application for probes and config"""
     app = web.Application()
-    app.router.add_get("/health", health_handler)
+    app.router.add_get("/healthz", health_handler)
```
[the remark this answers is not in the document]
acardace: Fixed.

pkg/kvcache/indexer_test.go
```diff
@@ -0,0 +1,368 @@
...
+	// whatever tokens it receives to 3 block keys, and only those 3
+	// blocks exist in the index.
+	//
+	// If truncation were broken (all 5 tokens passed through), the token
+	// processor would still return the same canned block keys, so to make
+	// the test meaningful we verify the scoring outcome is correct for the
+	// truncated token set.
+	blockKeys := u64ToBlockKeys([]uint64{10, 20, 30})
+	tp := &mockTokenProcessor{blockKeys: blockKeys}
+	pool := &mockTokenizersPool{tokens: []uint32{100, 200, 300, 400, 500}}
```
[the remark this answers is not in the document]
acardace: Fixed.

pkg/kvcache/indexer.go
```diff
@@ -159,8 +151,33 @@ func (k *Indexer) GetPodScores(ctx context.Context, renderReq *types.RenderChatR
...
+
 	blockKeys := k.tokenProcessor.TokensToKVBlockKeys(kvblock.EmptyBlockHash, tokens, modelName)
+
+	// Start tracing span for main operation
+	tracer := otel.Tracer(telemetry.InstrumentationName)
+	ctx, span := tracer.Start(ctx, "llm_d.kv_cache.get_scores",
+		trace.WithSpanKind(trace.SpanKindInternal),
+	)
+	defer span.End()
+
```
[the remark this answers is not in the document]
acardace: Fixed.

pkg/kvcache/indexer.go
```diff
@@ -159,8 +151,38 @@ func (k *Indexer) GetPodScores(ctx context.Context, renderReq *types.RenderChatR
...
+	modelName string,
+	podIdentifiers []string,
+) (map[string]float64, error) {
+	// Start tracing span for main operation
+	tracer := otel.Tracer(telemetry.InstrumentationName)
+	ctx, span := tracer.Start(ctx, "llm_d.kv_cache.get_scores",
+		trace.WithSpanKind(trace.SpanKindInternal),
+	)
+	defer span.End()
+
```
vMaroon, line 168: Perhaps add something like `attribute.Int("llm_d.kv_cache.token_count", len(tokens))` to potentially help correlate token volume with scoring latency.

pkg/kvcache/indexer.go
```diff
@@ -59,6 +59,13 @@ func NewDefaultConfig() (*Config, error) {
 	}, nil
 }
 
+// TokenizersPool abstracts the tokenization pool for testability/mocking.
+type TokenizersPool interface {
```
vMaroon: I don't think this is the right place for this interface, perhaps in `tokenization` package?

pkg/kvcache/indexer.go
```diff
@@ -159,8 +151,38 @@ func (k *Indexer) GetPodScores(ctx context.Context, renderReq *types.RenderChatR
...
+}
+
+// GetPodScoresFromTokens retrieves the pod scores for a given array of tokens and model name.
+// The function receives the mentioned information and a list of relevant pod
+// identifiers. A Pod identifier should be its address.
+// If the set of pod identifiers is empty, the function assumes all pods are
+// relevant.
+//
+// The function returns a map of pod identifiers to scores.
+func (k *Indexer) GetPodScoresFromTokens(
```
vMaroon: The name reads off - what do you think about `ScoreTokens`?

pkg/kvcache/indexer.go
```diff
@@ -159,8 +151,38 @@ func (k *Indexer) GetPodScores(ctx context.Context, renderReq *types.RenderChatR
 		}
 	}
 
-	// 3. get block keys
+	return k.GetPodScoresFromTokens(ctx, tokens, modelName, podIdentifiers)
+}
+
+// GetPodScoresFromTokens retrieves the pod scores for a given array of tokens and model name.
+// The function receives the mentioned information and a list of relevant pod
```
vMaroon: nit: while not introduced here, "retrieves the pod scores" makes it sound like they exist and are just fetched. Hardening the process that happens within this call can be beneficial.

pkg/kvcache/indexer.go
acardace: Yup, will do.

pkg/kvcache/indexer.go
acardace: This is actually the idiomatic Go placement — interfaces belong in the package that uses them, not the one that implements them. The Go wiki Code Review Comments (https://go.dev/wiki/CodeReviewComments#interfaces) calls this out explicitly, and the stdlib follows the same pattern throughout. Keeping it here means tokenization doesn't need to know what its consumers care about, and the interface stays minimal.

pkg/kvcache/indexer.go
acardace: Sounds good.

pkg/kvcache/indexer.go
acardace: I've changed the comment, what do you mean about hardening the process exactly?

pkg/kvcache/indexer.go
vMaroon: That is fine - still:
1. The interface can be defined in a different file, within this package
2. The rest of the codebase does not follow this pattern. Each package defines an interface and a set of implementations.

I do get the logic behind this pattern, but partial enforcement would cause confusion IMO.
I think we can start an issue on investigating a refactor that follows it, e.g., move all types consumed by the central Indexer into a `types.go` and see how that would look like / the clarity it adds. I think it might be even better to check out the practices in the IGW project and follow that.

acardace: IGW is a mixed bag, the epp/framework/interface interfaces all live in dedicated interface-only pkgs but there are example of interfaces like `DataStore/shardProcessor/Scheduler` whom are defined at the consumer. That said IGW interfaces are meant to be implemented by code importing the lib, it's a different case than llm-d-kv-cache which uses this interface internally mostly just for mocking. I'm fine with either solution to be honest, do you have a preference?

acardace: I've extracted the interface to a dedicated types.go file in the kvcache package.

PR #419, opened 2026-03-12
feat: add GPUDirect Storage (GDS) support to llmd_fs_backend

kfirtoledo:
Adds GPUDirect Storage (GDS) support via cuFile library, enabling direct DMA transfers between GPU memory andstorage without CPU staging buffers.

Changes:

- **New StorageHandler abstraction (storage_handler.hpp):** polymorphic interface with GdsFileIO (GDS path) and FileIO (CPU staging path) implementations, selected at runtime based on gds_mode

- **New GdsFileIO backend (backends/fs_gds/):** cuFile-based read/write with per-buffer GPU registration, supports all GDS modes including Bounce Buffer (BB) variants for filesystems

- **Restructured file_io:** moved to backends/fs_io/ for consistency

- **New gds_mode parameter** added throughout the stack (worker.py, spec.py, storage_offload.cpp, storage_offload_bindings.cpp, tests). Controls GDS behavior per direction. Valid values: disabled, read_only, write_only, read_write, bb_read_only, bb_write_only, bb_read_write

- **setup.py:** GDS support is detected at build time via ldconfig -p; USE_CUFILE is set automatically when libcufile.so is found. Falls back to CPU-only if not available.
 Build without cuFile: pip install -e . on a machine without libcufile.so → builds CPU-only
 Build with cuFile: pip install -e . on GDS-capable machine → builds with USE_CUFILE

- **New test file tests/test_gds_backend.py:** GDS roundtrip, mode-specific read/write, BB mode tests; skips automatically when GDS is unavailable

- **Docs: docs/gds.md** — setup, modes reference, performance tuning; docs/cufile_rdma.json — annotated cuFile config template for storage.

### DIFF
--- a/kv_connectors/llmd_fs_backend/csrc/storage/backends/fs_gds/gds_file_io.cpp
+++ b/kv_connectors/llmd_fs_backend/csrc/storage/backends/fs_gds/gds_file_io.cpp
@@ -38,7 +38,8 @@ GdsFileIO::GdsFileIO(const std::vector<std::pair<void*, size_t>>& gpu_buffers,
                      size_t block_size,
                      GdsMode gds_mode,
                      TensorCopier& tensor_copier)
-    : m_gds_initialized(false),
+    : m_cufile(CuFileApi::instance()),
+      m_gds_initialized(false),
       m_gds_mode(gds_mode),
       m_tensor_copier(tensor_copier),
       m_use_for_read(gds_mode == GdsMode::READ_ONLY ||
@@ -70,34 +71,30 @@ GdsFileIO::GdsFileIO(const std::vector<std::pair<void*, size_t>>& gpu_buffers,
                                          << (total_bytes / (1024.0 * 1024.0))
                                          << " MB total)");
   } else {
-    FS_LOG_WARN(
-        "GdsFileIO: GDS initialization failed, using CPU "
-        "buffer staging");
+    FS_LOG_WARN("GdsFileIO: GDS initialization failed");
   }
 }
 
 // Destructor - cleanup all GDS resources
 GdsFileIO::~GdsFileIO() {
-#ifdef USE_CUFILE
   if (!m_gds_initialized) {
     return;
   }
 
   // Deregister all buffers
   for (const auto& [ptr, size] : m_registered_buffers) {
-    cuFileBufDeregister(ptr);
+    m_cufile.cuFileBufDeregister(ptr);
   }
   m_registered_buffers.clear();
 
   // Close driver
-  CUfileError_t status = cuFileDriverClose();
+  CUfileError_t status = m_cufile.cuFileDriverClose();
   if (status.err != CU_FILE_SUCCESS) {
     FS_LOG_WARN(
         "GdsFileIO: cuFileDriverClose failed with error code: " << status.err);
   }
 
   m_gds_initialized = false;
-#endif
 }
 
 // Helper to check if current mode uses Bounce Buffer
@@ -109,29 +106,31 @@ bool GdsFileIO::is_bb_mode() const {
 
 // Static capability check
 bool GdsFileIO::is_gds_supported() {
-#ifdef USE_CUFILE
-  // Check if cuFile library is available
-  CUfileError_t status;
+  auto& cufile = CuFileApi::instance();
+  if (!cufile.is_loaded()) {
+    return false;
+  }
 
   // Try to get cuFile version as a simple availability check
   int version = 0;
-  status = cuFileGetVersion(&version);
+  CUfileError_t status = cufile.cuFileGetVersion(&version);
 
   if (status.err != CU_FILE_SUCCESS) {
     return false;
   }
 
   return true;
-#else
-  return false;
-#endif
 }
 
 // Initialize GDS driver - opens cuFile driver and queries capabilities
 bool GdsFileIO::initialize_gds() {
-#ifdef USE_CUFILE
-  CUfileError_t status = cuFileDriverOpen();  // Initialize cuFile driver (must
-                                              // be called once per process)
+  if (!m_cufile.is_loaded()) {
+    return false;
+  }
+
+  CUfileError_t status = m_cufile.cuFileDriverOpen();  // Initialize cuFile
+                                                       // driver (must be called
+                                                       // once per process)
 
   if (status.err != CU_FILE_SUCCESS) {
     FS_LOG_ERROR(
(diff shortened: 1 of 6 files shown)


kv_connectors/llmd_fs_backend/csrc/storage/backends/fs_gds/gds_file_io.cpp
```diff
@@ -0,0 +1,443 @@
...
+    return false;
+  }
+
+  // Check GPU capability
+  int gds_supported = 0;
+  cuda_err = cudaDeviceGetAttribute(&gds_supported,
+                                    cudaDevAttrGPUDirectRDMASupported,
+                                    device_id);
+
+  if (cuda_err != cudaSuccess || gds_supported == 0) {
```
effi-ofer: Superfluous?  Will already return false on line 123.

kv_connectors/llmd_fs_backend/csrc/storage/backends/fs_gds/gds_file_io.hpp
```diff
@@ -0,0 +1,108 @@
...
+#include <string>
+#include <memory>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+#include <utility>
+#include <cuda_runtime.h>
+
+// Forward declare cuFile types to avoid header dependency when GDS is not
+// available
```
effi-ofer: Is the forward declaration needed?  I see every instance of CUfileHandle_t and CUfileDescr_t is surrounded by ifdef USE_CUFILE.

kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/backends/fs_gds/gds_file_io.cpp
kfirtoledo: You are right, I removed it

kv_connectors/llmd_fs_backend/csrc/storage/backends/fs_gds/gds_file_io.cpp
```diff
@@ -0,0 +1,445 @@
...
+  status = cuFileDriverGetProperties(&props);
+  if (status.err == CU_FILE_SUCCESS) {
+    FS_LOG_INFO("GdsFileIO: cuFile driver properties:\n"
+                << "  - max_device_cache_size: " << props.max_device_cache_size
+                << "\n"  // Device cache limit
+                << "  - max_device_pinned_mem_size: "
+                << props.max_device_pinned_mem_size);
+  }
+
+  return true;
```
effi-ofer, line 160: if we fail on the if (status.err == CU_FILE_SICCESS) do we still want to return true?

kv_connectors/llmd_fs_backend/csrc/storage/backends/fs_gds/gds_file_io.cpp
```diff
@@ -0,0 +1,445 @@
...
+        total_bytes += size;
+      }
+    }
+    FS_LOG_INFO("GdsFileIO: Registered " << gpu_buffers.size() << " buffers ("
+                                         << (total_bytes / (1024.0 * 1024.0))
+                                         << " MB total)");
+  } else {
+    FS_LOG_WARN(
+        "GdsFileIO: GDS initialization failed, using CPU "
+        "buffer staging");
```
effi-ofer: The comment confused me at first, making me question whether we stayed in gds mode but switched to one of the BB using setups.  Looking at the callers I see they check for gds is available and if not, print a similar message, reset gds_io to nullptr and swtich to cpu buffering.  Perhaps it would be clearer to just say gds initalization failed and let the caller specify the resolution.

kv_connectors/llmd_fs_backend/csrc/storage/backends/fs_gds/gds_file_io.cpp
```diff
@@ -0,0 +1,445 @@
...
+               << (size / (1024.0 * 1024.0)) << " MB)"
+               << (block_size > 0
+                       ? " block_size " + std::to_string(block_size) + " bytes"
+                       : ""));
+
+  // Check if current GDS mode uses Bounce Buffer
+  bool use_bb = is_bb_mode();
+
+  // If block_size is 0 or BB mode is enabled, register entire buffer at once
+  if (block_size == 0 || use_bb) {
```
effi-ofer: I don't see the code ever setting block_size to 0.  Assuming we do want to support this option, would it make sense to mix the use_bb scenario with the register entire buffer at once?  Wouldn't we still want to use CU_FILE_RDMA_REGISTER if !use_bb?

kv_connectors/llmd_fs_backend/csrc/storage/backends/fs_gds/gds_file_io.cpp
```diff
@@ -0,0 +1,445 @@
...
+                  << status.err);
+      return false;
+    }
+
+    m_registered_buffers[gpu_ptr] = size;
+    return true;
+  }
+
+  // TODO: Group blocks into larger chunks to reduce GDS registration entries.
+  // Currently 1 (one block at a time) due to GDS driver limitations.
```
effi-ofer: Can you spell the limitation?

kv_connectors/llmd_fs_backend/csrc/storage/backends/fs_gds/gds_file_io.cpp
```diff
@@ -0,0 +1,440 @@
...
+  }
+
+  // Write all blocks sequentially
+  bool success = true;
+  off_t file_offset = 0;
+
+  for (size_t bi = 0; bi < block_ids.size() && success; ++bi) {
+    int64_t gpu_block_idx = block_ids[bi];
+
+    for (const auto& tensor : tensors) {
```
effi-ofer, line 290: Given the new cross layout, would we ever have more than one tensor?

kv_connectors/llmd_fs_backend/csrc/storage/backends/fs_gds/gds_file_io.cpp
```diff
@@ -0,0 +1,440 @@
...
+  for (size_t i = 0; i < num_chunks; i++) {
+    void* block_ptr = static_cast<uint8_t*>(gpu_ptr) + (i * chunk_size);
+
+    CUfileError_t status = m_cufile.cuFileBufRegister(block_ptr,
+                                                      chunk_size,
+                                                      CU_FILE_RDMA_REGISTER);
+    if (status.err != CU_FILE_SUCCESS) {
+      FS_LOG_WARN("GdsFileIO: cuFileBufRegister failed for block "
+                  << i << " with error code: " << status.err);
+      return false;
```
effi-ofer, line 224: If the registration fails and we return false, what happens to the other registrations?  I see we will clean them up in the destructor, so no registration leakage, but the caller (the constructor) ignores the return code and just logs a warning. Isn't m_gds_initialized == true at this point?

kv_connectors/llmd_fs_backend/csrc/storage/backends/fs_gds/gds_file_io.cpp
kfirtoledo: Done, I removed it

kv_connectors/llmd_fs_backend/csrc/storage/backends/fs_gds/gds_file_io.cpp
kfirtoledo: You're right, I've added a warning log when cuFileDriverGetProperties fails. We still return true because the driver is already open and functional — this call only fetches info for logging, it doesn't affect GDS operations.

kv_connectors/llmd_fs_backend/csrc/storage/backends/fs_gds/gds_file_io.cpp
kfirtoledo: Done, updated the comment

kv_connectors/llmd_fs_backend/csrc/storage/backends/fs_gds/gds_file_io.cpp
kfirtoledo: If registration fails, m_gds_initialized remains true but with no registered buffers — in that case GDS I/O will fall back to bounce buffering. The destructor cleans up any partially registered buffers, so no leakage.

kv_connectors/llmd_fs_backend/csrc/storage/backends/fs_gds/gds_file_io.cpp
kfirtoledo: Yes, but we should also support cases without the new cross-layer layout

kv_connectors/llmd_fs_backend/csrc/storage/backends/fs_gds/gds_file_io.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/docs/gds.md
```diff
@@ -0,0 +1,142 @@
...
+
+GDS uses the **cuFile** library by default for all GPU↔storage transfers. cuFile behavior (I/O threads,
+bounce buffer sizes, RDMA settings, filesystem-specific tuning) can be customized via a `cufile_rdma.json`
+config file — see [Tuning cuFile](#tuning-cufile-cufilejson) for an example.
+
+## Requirements
+
+- NVIDIA GPU with GDS support
+- cuFile library installed at runtime (`nvidia-fs` kernel module + `libcufile`)
+- Filesystem that supports O_DIRECT (e.g., ext4, xfs, NVMe-oF — **not** most NFS/FUSE mounts)
```
dannyharnik: can comment that "it will work with such filesystems but will not give the expected benefit of bypassing the CPU."

kfirtoledo: Done

PR #422, opened 2026-03-16
fix: Close data race in InMemoryIndex Add/Evict with RWMutex

### LINKED ISSUE #421: InMemoryIndex: lost writes under concurrent Add/Evict due to non-atomic compound operations
InMemoryIndex has no coarse-grained lock. It relies on individual LRU operations being thread-safe and a per-PodCache mutex (podCache.mu), but multi-step sequences across Add and Evict are not atomic, leading to silent data loss.

Race sequence:

-   step 1: Add (goroutine A): calls m.data.Get(requestKey) → gets *PodCache P, found=true
-   step 2: Evict (goroutine B): calls m.data.Get(requestKey) → gets the same *PodCache P
-   step 3: Evict: locks P.mu, removes entries from P.cache, observes P.cache.Len() == 0, unlocks
-   step 4: Evict: double-check path: re-fetches from m.data, locks P.mu, confirms Len() == 0, calls m.data.Remove(requestKey), unlocks
-   step 5: Add (goroutine A): locks P.mu, adds entries to P.cache, unlocks

Result: P has been removed from m.data, but Add just wrote entries into it. Subsequent Lookup calls will not find these entries: data is silently lost.

Root cause: The gap between m.data.Get() in step 1 and podCache.mu.Lock() in step 5 is unprotected. Evict's double-check (step 4) only verifies that the cache is empty while holding P.mu, but it cannot detect that another goroutine has already obtained a reference to P and is about to write to it.

Impact: Pods that were just registered via Add become unreachable, causing cache misses on Lookup and degraded KV cache hit rates.

gyliu513:
Fixed #421 

- Add `sync.RWMutex` to `InMemoryIndex` to protect compound read-modify-write sequences in `Add` and `Evict` that were not atomic, allowing concurrent `Evict` to remove a `PodCache` from `data` while `Add` was writing entries into it, silently losing those entries.
- `Lookup` and `GetRequestKey` use `RLock` (concurrent reads allowed); `Add` and `Evict` use `Lock` (mutually exclusive).
- Remove the now-redundant `PodCache.mu` and the double-check relocking pattern in `Evict`, simplifying the eviction path.

### DIFF
--- a/pkg/kvcache/kvblock/in_memory.go
+++ b/pkg/kvcache/kvblock/in_memory.go
@@ -94,6 +94,9 @@ type PodCache struct {
 	cache *lru.Cache[PodEntry, struct{}]
 	// mu protects the cache from concurrent access during check-and-set operations.
 	mu sync.Mutex
+	// removed indicates this PodCache has been evicted from the parent map.
+	// Checked by Add after acquiring mu to avoid writing into an orphaned cache.
+	removed bool
 }
 
 // Lookup receives a list of requestKeys and a set of pod identifiers,
@@ -178,46 +181,26 @@ func (m *InMemoryIndex) Add(ctx context.Context, engineKeys, requestKeys []Block
 
 	// Store requestKey -> PodCache mappings for all request keys.
 	for _, requestKey := range requestKeys {
-		var podCache *PodCache
-		var found bool
+		// Retry on stale cache: needed only when a concurrent Evict marks the
+		// PodCache as removed between getOrCreatePodCache and Lock. The window
+		// is tiny, so this loop almost never iterates more than once.
+		for {
+			podCache := m.getOrCreatePodCache(requestKey)
 
-		// Try to get existing cache first
-		podCache, found = m.data.Get(requestKey)
-		//nolint:nestif // double-checked locking pattern
-		if !found {
-			// Create new cache
-			cache, err := lru.New[PodEntry, struct{}](m.podCacheSize)
-			if err != nil {
-				return fmt.Errorf("failed to create pod cache for key %s: %w", requestKey.String(), err)
+			podCache.mu.Lock()
+			if podCache.removed {
+				podCache.mu.Unlock()
+				continue // retry — this cache was evicted
 			}
 
-			newPodCache := &PodCache{
-				cache: cache,
+			for _, entry := range entries {
+				podCache.cache.Add(entry, struct{}{})
 			}
+			podCache.mu.Unlock()
 
-			// Try to add, but use existing if another thread added it first
-			// This is a bounded retry (1) - not perfectly safe but for practical use-cases and scenarios
-			// this should be sufficient
-			contains, _ := m.data.ContainsOrAdd(requestKey, newPodCache)
-			if contains {
-				podCache, found = m.data.Get(requestKey)
-				if !found { // Extremely irregular workload pattern - key evicted
-					m.data.Add(requestKey, newPodCache)
-					podCache = newPodCache
-				}
-			} else {
-				// We successfully added our cache
-				podCache = newPodCache
-			}
+			traceLogger.Info("added pods to key", "requestKey", requestKey, "pods", entries)
+			break
 		}
-
-		podCache.mu.Lock()
-		for _, entry := range entries {
-			podCache.cache.Add(entry, struct{}{})
-		}
-		podCache.mu.Unlock()
-
-		traceLogger.Info("added pods to key", "requestKey", requestKey, "pods", entries)
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


yankay (changes requested):
## Review: fix: Close data race in InMemoryIndex Add/Evict with RWMutex

The PR correctly identifies and fixes the race described in #421. However, the global `RWMutex` approach introduces a significant concurrency regression and removes the existing fine-grained locking entirely. I'd suggest a more targeted fix.

### What the PR does right

The race analysis in #421 is spot-on. The gap between `m.data.Get(requestKey)` and `podCache.mu.Lock()` in `Add` allows a concurrent `Evict` to remove the PodCache from `m.data`, and subsequent writes by `Add` go into an orphaned object. The PR closes this gap.

### Concern 1: Global write lock serializes all mutations — even for unrelated keys

`Add` and `Evict` both acquire `mu.Lock()` (exclusive) for their entire duration. This means:

- Two `Add` calls for **completely different** `requestKeys` cannot run concurrently.
- `Add` and `Evict` for **unrelated keys** cannot run concurrently.
- A single large `Add(requestKeys=[k1,k2,...,kN])` holds the write lock while looping over all N keys, blocking every other operation.

The original per-PodCache `mu` allowed different keys to be modified in parallel. This PR replaces that with full serialization.

### Concern 2: Lookup blocked by any mutation

`Lookup` takes `RLock` for its entire loop over `requestKeys`. This means every `Lookup` (the hot path — called on every inference request) is blocked whenever any `Add` or `Evict` is in progress, regardless of whether they touch the same keys.

### Concern 3: GetRequestKey does not need a read lock

`GetRequestKey` is a single `m.engineToRequestKeys.Get()` call. `lru.Cache` is already internally thread-safe, so this is atomic without the outer lock. The added `RLock`/`RUnlock` is unnecessary overhead.

### Concern 4: No regression test

This is the root cause behind the flaky test that [PR #366](https://github.com/llm-d/llm-d-kv-cache/pull/366) worked around by removing correctness assertions from `ConcurrentOperations`:

> _"These assertions are invalid because Evict can remove the entire PodCache when it becomes temporarily empty, causing pods from other goroutines to be lost."_

The fix for #421 should include a targeted test that reproduces the specific race (concurrent Add + Evict-to-empty on the same key) and asserts that newly added entries are not lost. This would also validate that #366's workaround is no longer needed.

### Suggested alternative: per-PodCache `removed` flag with retry

The race is specifically between one PodCache's fetch and lock. A surgical fix that preserves fine-grained concurrency:

1. Add a `removed bool` field to `PodCache`.
2. **Evict**: when the cache empties, set `removed = true` **while holding `podCache.mu`**, then call `m.data.Remove()` — both under the same lock hold.
3. **Add**: after acquiring `podCache.mu`, check `removed`. If true, unlock and retry (the retry will either find a new PodCache or create one).

```go
// PodCache — add a removed flag
type PodCache struct {
    cache   *lru.Cache[PodEntry, struct{}]
    mu      sync.Mutex
    removed bool
}

// Evict — set removed + remove from map atomically under podCache.mu
podCache.mu.Lock()
for _, entry := range entries {
    podCache.cache.Remove(entry)
}
if podCache.cache.Len() == 0 {
    podCache.removed = true
    if cur, ok := m.data.Peek(requestKey); ok && cur == podCache {
        m.data.Remove(requestKey)
        m.engineToRequestKeys.Remove(engineKey)
    }
}
podCache.mu.Unlock()

// Add — check removed after acquiring lock, retry if stale
for attempt := 0; attempt < maxRetries; attempt++ {
    podCache := getOrCreatePodCache(requestKey)
    podCache.mu.Lock()
    if podCache.removed {
        podCache.mu.Unlock()
        continue
    }
    for _, entry := range entries {
        podCache.cache.Add(entry, struct{}{})
    }
    podCache.mu.Unlock()
    return nil
}
```

This approach:
- Preserves concurrent Add/Evict on **different** keys (the original benefit of per-PodCache locking).
- Does **not** block Lookup at all (no locking needed in the read path).
- Only serializes operations on the **same** key, which is exactly the scope of the #421 race.
- Uses `Peek` + pointer equality to avoid removing a replacement PodCache that a concurrent Add may have inserted.

yankay:
Thanks @gyliu513 for the update! The `removed` flag + retry approach looks correct to me — nice work adopting the per-PodCache locking strategy.

One minor suggestion: the `for {}` retry loop in `Add` is unbounded. In practice it retries at most once (since a freshly created empty PodCache can't be marked `removed` due to the `prevLen > 0` guard), but adding a `maxRetries` bound would be good defensive programming against unexpected edge cases. Something like:

```go
const maxAddRetries = 3

for attempt := 0; attempt < maxAddRetries; attempt++ {
    podCache := m.getOrCreatePodCache(requestKey)
    podCache.mu.Lock()
    if podCache.removed {
        podCache.mu.Unlock()
        continue
    }
    // ... write entries ...
    podCache.mu.Unlock()
    break
}
```

Tests look great — `TestConcurrentAddEvictToEmpty` (500 iterations) and `TestConcurrentAddWithStaleEvicts` (anti-spin guard) are well-targeted regression tests for the exact race in #421.

gyliu513:
Hi @yankay , thanks for the suggestion! I agree it's good defensive practice, but since the prevLen > 0 guard in Evict guarantees a freshly created empty PodCache is never marked as removed, the retry is bounded to at most one iteration in practice. How about keep it as-is for now to avoid adding complexity?

Issue #425, opened 2026-03-16
[MM] update sidecar to use `OpenAIServingRender` and parse `GenerateRequest`


sagearc:
- Update the kv-cache-manager's renderer client to wrap `OpenAIServingRender` (from vLLM, replacing the current transformers-based tokenization)
- Update the protocol to use [GenerateRequest](https://github.com/sergey-zinchenko/vllm/blob/44102dbbd39f8be4e4b973ed1f824c7bc5ee4dc7/vllm/entrypoints/serve/disagg/protocol.py#L54-L112) as the response format
- Once [vLLM gRPC renderer](vllm-project/vllm#36102 ) is ready, we can replace the wrapper with a direct gRPC call (tracked for 0.7).

PR #280, opened 2026-02-03
uds tokenizer using vllm wrapper

delavet:
fix #126

This PR updates the way the disaggregated uds tokenizer service works. Now, the service completes `ApplyChatTemplate` and `Encode` by using the vllm wrapper.

PR #428, opened 2026-03-16
Add `RenderChatCompletion` and `RenderCompletion` gRPC RPCs via vLLM `OpenAIServingRender`

sagearc:
Add `RenderChatCompletion` and `RenderCompletion` gRPC RPCs via vLLM `OpenAIServingRender`

Related to #425.

Extends the tokenizer sidecar to support full OpenAI request preprocessing by wrapping vLLM's `OpenAIServingRender` — the CPU-only path that converts chat/completion requests into `GenerateRequest`-compatible token IDs and multimodal features, without requiring a running inference engine.

**Changes:**

- **Proto** — adds `RenderChatCompletion` and `RenderCompletion` RPCs with new message types for requests, responses, and multimodal features. Regenerated Python bindings.

- **`RendererService`** (`tokenizer_service/renderer.py`) — lazily loads one `OpenAIServingRender` per model, thread-safe. Explicitly sets `DeviceConfig(device="cpu")` so initialization works on CPU-only environments without GPU platform detection.

- **gRPC servicer** — adds the two RPC handlers with a shared `_generate_request_to_proto` helper. Uses a single persistent background event loop to safely dispatch async renderer calls from `ThreadPoolExecutor` threads.

- **Tests** — integration tests covering text-only, multi-turn, multi-prompt, determinism, invalid JSON, and MM feature absence for text inputs.

**Based on / credits:**

This work is based on llm-d/llm-d-kv-cache#280 and vllm-project/vllm#36102. Co-authors from those PRs have been added to this PR accordingly.

### DIFF
--- a/services/uds_tokenizer/tokenizer_service/renderer.py
+++ b/services/uds_tokenizer/tokenizer_service/renderer.py
@@ -18,6 +18,18 @@ import json
 import logging
 import threading
 
+from vllm.config import VllmConfig
+from vllm.config.device import DeviceConfig
+from vllm.engine.arg_utils import AsyncEngineArgs
+from vllm.entrypoints.openai.chat_completion.protocol import ChatCompletionRequest
+from vllm.entrypoints.openai.completion.protocol import CompletionRequest
+from vllm.entrypoints.openai.engine.protocol import ErrorResponse
+from vllm.entrypoints.openai.models.protocol import BaseModelPath
+from vllm.entrypoints.openai.models.serving import OpenAIModelRegistry
+from vllm.entrypoints.serve.render.serving import OpenAIServingRender
+from vllm.plugins.io_processors import get_io_processor
+from vllm.renderers import renderer_from_config
+
 
 class RendererError(Exception):
     pass
@@ -55,15 +67,6 @@ class RendererService:
                 return False
 
     def _build_serving_render(self, model_name: str, chat_template: str | None):
-        from vllm.config import VllmConfig
-        from vllm.config.device import DeviceConfig
-        from vllm.engine.arg_utils import AsyncEngineArgs
-        from vllm.entrypoints.serve.render.serving import OpenAIServingRender
-        from vllm.entrypoints.openai.models.serving import OpenAIModelRegistry
-        from vllm.entrypoints.openai.models.protocol import BaseModelPath
-        from vllm.renderers import renderer_from_config
-        from vllm.plugins.io_processors import get_io_processor
-
         engine_args = AsyncEngineArgs(model=model_name)
         model_config = engine_args.create_model_config()
         vllm_config = VllmConfig(
@@ -94,11 +97,6 @@ class RendererService:
 
     async def render_chat(self, request_json: str, model_name: str):
         """Render an OpenAI chat completion request, returning a vLLM GenerateRequest."""
-        from vllm.entrypoints.openai.chat_completion.protocol import (
-            ChatCompletionRequest,
-        )
-        from vllm.entrypoints.openai.engine.protocol import ErrorResponse
-
         serving_render = self._get_renderer(model_name)
 
         try:
@@ -116,9 +114,6 @@ class RendererService:
 
     async def render_completion(self, request_json: str, model_name: str):
         """Render an OpenAI completion request, returning a list of vLLM GenerateRequests."""
-        from vllm.entrypoints.openai.completion.protocol import CompletionRequest
-        from vllm.entrypoints.openai.engine.protocol import ErrorResponse
-
         serving_render = self._get_renderer(model_name)
 
         try:


services/uds_tokenizer/tokenizer_grpc_service.py
```diff
@@ -28,17 +31,28 @@
...
-    def __init__(self, tokenizer_service: TokenizerService):
+    def __init__(
+        self, tokenizer_service: TokenizerService, renderer_service: RendererService
+    ):
         self.tokenizer_service = tokenizer_service
+        self.renderer_service = renderer_service
+        self._loop = asyncio.new_event_loop()
+        self._loop_thread = threading.Thread(target=self._loop.run_forever, daemon=True)
+        self._loop_thread.start()
         logging.info("TokenizationServiceServicer initialized")
```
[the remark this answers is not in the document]
sagearc: Redundant. Shutdown happens with process

services/uds_tokenizer/tokenizer_service/renderer.py
```diff
@@ -0,0 +1,131 @@
...
+
+            try:
+                self._renderers[model_name] = self._build_serving_render(
+                    model_name, chat_template
+                )
+                logging.info("Renderer loaded for model: %s", model_name)
+                return True
+            except Exception as e:
+                logging.error("Failed to load renderer for model %s: %s", model_name, e)
+                return False
```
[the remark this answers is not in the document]
sagearc: This service loads one model — the model the vLLM engine is serving. Multiple models is not a real use case here.

services/uds_tokenizer/tokenizer_service/renderer.py
```diff
@@ -0,0 +1,136 @@
...
+        from vllm.entrypoints.serve.render.serving import OpenAIServingRender
+        from vllm.entrypoints.openai.models.serving import OpenAIModelRegistry
+        from vllm.entrypoints.openai.models.protocol import BaseModelPath
+        from vllm.renderers import renderer_from_config
+        from vllm.plugins.io_processors import get_io_processor
+
+        engine_args = AsyncEngineArgs(model=model_name)
+        model_config = engine_args.create_model_config()
+        vllm_config = VllmConfig(
+            model_config=model_config, device_config=DeviceConfig(device="cpu")
```
sagearc, line 73: workaround until stable cpu wheel

tests/e2e/uds_tokenizer/uds_e2e_test.go
```diff
@@ -63,16 +63,14 @@ func (s *UDSTokenizerSuite) TestTokenizeWithSpecialTokens() {
 	s.Require().NoError(err)
 	s.Require().NotEmpty(tokensWithoutSpecial)
 
-	// BERT adds [CLS] at the start and [SEP] at the end when add_special_tokens=true.
+	// TinyLlama adds <s> (BOS) at the start when add_special_tokens=true.
```
sagearc, line 66: bert is encoder only and thus is not compatible with vllm renderer

api/tokenizerpb/tokenizer.proto
```diff
@@ -110,6 +110,56 @@ message InitializeTokenizerResponse {
...
+  string request_id = 1;            // Request ID from the render response
+  repeated uint32 token_ids = 2;    // Token IDs for the rendered input
+  MultiModalFeatures features = 3;  // Multimodal features (populated for MM inputs)
+  bool success = 4;                 // Whether the request was successful
+  string error_message = 5;         // Error message if the request failed
+}
+
+// RenderCompletionRequest wraps an OpenAI completion request for rendering
+message RenderCompletionRequest {
+  string request_json = 1; // JSON-serialized OpenAI CompletionRequest
```
delavet, line 152: In the newly added proto definition, what is actually transmitted is JSON-serialized data. This significantly diminishes the value of gRPC—we still incur unnecessary overhead from two rounds of JSON serialization and deserialization.

To my understanding, this should only be a temporary workaround. Once vLLM natively supports gRPC `.proto` definitions for the renderer, we can directly adopt those official protocol definitions.

Given the current state, we could—in principle—define our own `.proto` files based on the `ChatCompletionRequest` and `CompletionRequest` class definitions to temporarily eliminate the redundant JSON (de)serialization. However, if we expect to replace the current solution with vLLM’s official `.proto` definitions in the near term, investing effort into a custom `.proto` may not be worthwhile.

Regardless, I believe we should explicitly track this issue and plan to replace the current approach at a well-defined future milestone.

services/uds_tokenizer/tests/test_renderer.py
```diff
@@ -0,0 +1,143 @@
...
+import tokenizerpb.tokenizer_pb2 as tokenizer_pb2
+from tokenizer_service.renderer import RendererService
+
+
+def _chat_request_json(model: str, messages: list[dict]) -> str:
+    """Build a minimal OpenAI ChatCompletionRequest JSON string."""
+    return json.dumps({"model": model, "messages": messages})
+
+
+class TestRenderChatCompletion:
```
delavet, line 40: The implementation of `render` supports multimodal content, but there is no test coverage for actual multimodal input. Will we address this in an MM-related issue in the future?

sagearc: Good point, I'll add relevant tests. Thanks!

api/tokenizerpb/tokenizer.proto
sagearc: Agreed — since this is being discussed in [llm-d#disaggregated-tokenization thread](https://llm-d.slack.com/archives/C09TNUMEXEW/p1773926573865559), let's merge this as-is and track the proper proto definitions as a follow-up PR.

services/uds_tokenizer/tokenizer_service/renderer.py
```diff
@@ -0,0 +1,135 @@
...
+        )
+
+    def _get_renderer(self, model_name: str):
+        if not self.load_renderer(model_name):
+            raise RendererError(f"Failed to load renderer for model: {model_name}")
+        return self._renderers[model_name]
+
+    async def render_chat(self, request_json: str, model_name: str):
+        """Render an OpenAI chat completion request, returning a vLLM GenerateRequest."""
+        from vllm.entrypoints.openai.chat_completion.protocol import (
```
gyliu513: Instead of lazy loading in each function, how about put those import to `__init__`?

The imports are inside method bodies now, executed on every call. While Python caches modules in sys.modules, the lookup still involves dict access and lock acquisition on every request. Dependencies are scattered across methods, making it hard to see what this module depends on.

services/uds_tokenizer/tokenizer_service/renderer.py
```diff
@@ -0,0 +1,135 @@
...
+
+        if isinstance(result, ErrorResponse):
+            raise RendererError(f"Render failed: {result.message}")
+
+        return result
+
+    async def render_completion(self, request_json: str, model_name: str):
+        """Render an OpenAI completion request, returning a list of vLLM GenerateRequests."""
+        from vllm.entrypoints.openai.completion.protocol import CompletionRequest
+        from vllm.entrypoints.openai.engine.protocol import ErrorResponse
```
gyliu513: ditto

services/uds_tokenizer/tests/test_renderer.py
```diff
@@ -0,0 +1,143 @@
...
+        ]
+        request_json = _chat_request_json(test_model, messages)
+        grpc_resp = grpc_stub.RenderChatCompletion(
+            tokenizer_pb2.RenderChatCompletionRequest(
+                request_json=request_json,
+                model_name=test_model,
+            )
+        )
+        assert grpc_resp.request_id
+        direct = asyncio.run(RendererService().render_chat(request_json, test_model))
```
gyliu513, line 97: The grpc_server fixture in conftest.py already creates a RendererService instance with the model loaded, can you exposed as a fixture for tests to reuse.

```python
@pytest.fixture(scope="session")
def renderer_service() -> RendererService:
    """Provide a shared RendererService instance."""
    return RendererService()
```

services/uds_tokenizer/tokenizer_grpc_service.py
```diff
@@ -95,46 +113,120 @@ def RenderChatTemplate(self, request, context):
...
+        try:
+            self.tokenizer_service.load_tokenizer(
                 request.model_name,
                 request.enable_thinking,
                 request.add_generation_prompt,
             )
+        except Exception as e:
+            logging.warning("Tokenizer load failed (non-critical): %s", e)
+
+        return tokenizer_pb2.InitializeTokenizerResponse(success=True)
```
gyliu513, line 140: Even exception finished still return tokenizer_pb2.InitializeTokenizerResponse(success=True)?  Can you share some detail why this error is not critical?

services/uds_tokenizer/tokenizer_service/renderer.py
sagearc: Good point, they didn't public macos compatible wheel so it is an artifact I should had removed. Thanks!

services/uds_tokenizer/tokenizer_grpc_service.py
sagearc: The related transformer-based RPCs will be removed from the main tokenization path right after this PR is merged. They are currently kept mainly for backward compatibility until they can be safely removed.

services/uds_tokenizer/tests/test_renderer.py
sagearc: The direct usages are all in `_matches_direct` tests which need a fresh instance by design to validate two independent paths, I think it over complicates without real gain since we load only the tokenizer for each model name

PR #429, opened 2026-03-16
build: Move fs-connector to use vllm v0.17.1

kfirtoledo:
## Summary
- Bump `torch` from `2.9.1` to `2.10.0` in `pyproject.toml` (build and runtime dependencies)
- Bump `vllm` dev dependency from `0.16.0` to `0.17.1`
- Add pre-built wheel for `llmd_fs_connector-0.17.1-cp312-cp312-linux_x86_64.whl`
- Update README to reference vLLM 0.17.1 and the new wheel URL


zdtsw:
we do not want to keep 0.14.1, but override with 0.17.1 ?

PR #431, opened 2026-03-17
Replace C based ZMQ with pure Go implementation

elevran:
Replace the CGO-based `pebbe/zmq4` ZMQ dependency with the pure-Go `go-zeromq/zmq4` library.
This eliminates the `libzmq/zeromq-devel` system package requirement at build and runtime, enables CGO_ENABLED=0, and simplifies cross-compilation. The corresponding `llm-d-inference-scheduler` changes (Dockerfile, Makefile, CI) will be in a separate PR once this is merged.

The pure-Go library supports the same ZMTP wire protocol so there's no compatibility impact with existing vLLM publishers.
Aside from minor API changes, the main code change is in `zmq_subscriber.go` which drops the polling loop (which existed solely to check for context cancellation every 250ms) and replace it with a blocking sub.Recv() call that the context cancels directly.

### DIFF
--- a/examples/helper/events.go
+++ b/examples/helper/events.go
@@ -37,7 +37,7 @@ func SimulateProduceEvent(ctx context.Context, publisher *Publisher) error {
 	// Use enough tokens to fill 4 blocks (matching PromptHashes count) at blockSize=16.
 	tokenIds := make([]uint32, 64)
 	for i := range tokenIds {
-		tokenIds[i] = uint32(i + 1)
+		tokenIds[i] = uint32(i + 1) //nolint:gosec // i is bounded by len(tokenIds)=64, no overflow
 	}
 
 	// Create event in vLLM msgpack array format: [tag, hashes, parent, tokens, blockSize, loraID, medium, loraName]


go.mod
```diff
@@ -62,6 +62,7 @@ require (
 	github.com/go-openapi/jsonpointer v0.21.0 // indirect
 	github.com/go-openapi/jsonreference v0.20.2 // indirect
 	github.com/go-openapi/swag v0.23.0 // indirect
+	github.com/go-zeromq/goczmq/v4 v4.2.2 // indirect
```
vMaroon, line 65: FYI I verified that this is ineffective - it is used by files in `go-zeromq/zmq4` that are guarded by `//go:build czmq4` tag. Building without CGO works as expected.

vMaroon:
The CI failure in the offline example is a behavioral difference between C libzmq and go-zeromq/zmq4. C libzmq's zmq_connect() is non-blocking and retries indefinitely in the background. go-zeromq's Dial() is synchronous, few retries then hard failure.

This affects two places:

  1. `examples/helper/publisher.go` - the offline example starts subscriber and publisher concurrently. If the subscriber hasn't bound within 2.5s, the publisher gives up. This is what broke CI.
  2. `pkg/kvevents/zmq_subscriber.go` (the remote=true / sub.Dial path) - in production, when a subscriber connects to a vLLM engine that isn't ready yet or restarts mid-session, Dial fails after 2.5s. The outer Start loop retries after 5s so it eventually recovers, but that's ~7.5s per attempt vs instant with C libzmq.

 Minimum fix: pass `zmq4.WithDialerMaxRetries(-1)` to both `NewSub` and `NewPub` to restore C libzmq's connect-and-retry-forever semantics. Consider also `zmq4.WithAutomaticReconnect(true)` on the kvevents subscriber for faster mid-session recovery.

PR #432, opened 2026-03-17
feat(tokenization): replace RenderChat with RenderChatCompletion RPC

### LINKED ISSUE #425: [MM] update sidecar to use `OpenAIServingRender` and parse `GenerateRequest`

sagearc:
Closes #425. Rebased on top of #461.

**Changes Overview**

Switches `Render` and `RenderChat` in `UdsTokenizer` to use the new `RenderCompletion` and `RenderChatCompletion` RPCs introduced in #461, replacing the old `RenderChatTemplate` flow.

On the Go side, `RenderChat` now builds a native `RenderChatCompletionRequest` proto (messages, tools, chat_template_kwargs) and returns token IDs directly instead of calling `Encode` on a rendered prompt string. `Render` calls `RenderCompletion` with the prompt list and returns token IDs directly too — neither returns character offsets anymore since the renderer service doesn't produce them.

**Protocol Design**

Tools and `chat_template_kwargs` are both serialized as JSON strings in the proto (`tools_json`, `chat_template_kwargs`). This avoids building a typed proto structure for fields that are already arbitrary JSON at the call site (`[]interface{}` in GIE's `ChatCompletionsRequest`), and lets Python deserialize them directly without field renaming or special-casing.

On the Python side, the gRPC servicer and renderer are updated to match: the renderer service methods now accept typed `ChatCompletionRequest`/`CompletionRequest` objects directly instead of going through a JSON round-trip. Proto-to-request conversion uses `MessageToDict` and `json.loads` for the JSON string fields.

Also excludes generated pb.go and pb2 files from golangci-lint and ruff in CI.


services/uds_tokenizer/tokenizer_grpc_service.py
```diff
@@ -198,16 +227,22 @@ def RenderCompletion(
...
-            items: list[tokenizer_pb2.RenderChatCompletionResponse] = [
-                self._generate_request_to_proto(r) for r in results
-            ]
-            return tokenizer_pb2.RenderCompletionResponse(items=items, success=True)
+            result = results[0]
+            return tokenizer_pb2.RenderCompletionResponse(
+                request_id=result.request_id,
+                token_ids=list(result.token_ids),
+                success=True,
+            )
```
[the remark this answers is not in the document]
sagearc: We do not batch requests

services/uds_tokenizer/tokenizer_grpc_service.py
```diff
@@ -180,10 +191,28 @@ def RenderChatCompletion(
     ) -> tokenizer_pb2.RenderChatCompletionResponse:
         """Render an OpenAI chat completion request via OpenAIServingRender."""
         try:
+            d = MessageToDict(request, preserving_proto_field_name=True)
```
gyliu513: `MessageToDict` by default omits proto3 fields that equal their default value. Since `add_generation_prompt` is a non-optional bool (proto3 default = false), when Go sends false, `MessageToDict` drops the field from the dict. 

The Python fallback `d.get("add_generation_prompt", True)` then returns True, silently overriding the caller's intent.
```
Go sets AddGenerationPrompt = false
    → proto sends false (== proto3 default)
    → MessageToDict omits the field
    → d.get("add_generation_prompt", True) returns True
```

This means there is no way for a Go caller to disable the generation prompt.

Suggested fix: pass `including_default_value_fields=True` to MessageToDict so that explicit false values are preserved:

```python
d = MessageToDict(request, preserving_proto_field_name=True, including_default_value_fields=True)
```

pkg/tokenization/uds_tokenizer_test.go
```diff
@@ -317,7 +324,7 @@ func (s *UdsTokenizerTestSuite) TestUdsTokenizer_TokenizeError() {
 
 	_, _, err := s.tokenizer.Render("test")
 	s.Assert().Error(err)
-	s.Assert().Contains(err.Error(), "tokenization failed")
+	s.Assert().Contains(err.Error(), "render completion failed")
 }
 
 func (s *UdsTokenizerTestSuite) TestUdsTokenizer_ChatTemplateError() {
```
gyliu513: s/TestUdsTokenizer_ChatTemplateError/TestUdsTokenizer_RenderChatTemplateError

services/uds_tokenizer/tokenizer_grpc_service.py
```diff
@@ -180,10 +191,28 @@ def RenderChatCompletion(
...
+                continue_final_message=d.get("continue_final_message", False),
+                chat_template_kwargs=d.get("chat_template_kwargs"),
+            )
             result = asyncio.run_coroutine_threadsafe(
-                self.renderer_service.render_chat(
-                    request.request_json, request.model_name
-                ),
+                self.renderer_service.render_chat(chat_request, request.model_name),
                 self._loop,
             ).result()
```
gyliu513, line 232: Not yours, but I think we need a timeout in order not to block the whole grpc server.

```python
from concurrent.futures import TimeoutError as FuturesTimeoutError

try:
    result = asyncio.run_coroutine_threadsafe(
        self.renderer_service.render_chat(chat_request, request.model_name),
        self._loop,
    ).result(timeout=30)
except FuturesTimeoutError:
    context.abort(grpc.StatusCode.DEADLINE_EXCEEDED, "render_chat timed out")
```

pkg/tokenization/uds_tokenizer.go
```diff
@@ -256,90 +275,54 @@ func (u *UdsTokenizer) RenderChat(
...
+		b, err := json.Marshal(renderReq.ChatTemplateKWArgs)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to marshal chat_template_kwargs: %w", err)
+		}
+		s := string(b)
+		chatTemplateKwargsJSON = &s
 	}
 
-	resp, err := u.client.RenderChatTemplate(ctx, req)
+	resp, err := u.client.RenderChatCompletion(ctx, &tokenizerpb.RenderChatCompletionRequest{
```
gyliu513, line 308: This will be calling `OpenAIServingRender.render_chat_request()` and it return `GenerateRequest`, `GenerateRequest ` just contain `token_ids`, I think for this function, there is no need to return `offset`, as it always nil.

pkg/tokenization/uds_tokenizer.go
```diff
@@ -228,7 +245,9 @@ func (u *UdsTokenizer) Encode(prompt string, addSpecialTokens bool) ([]uint32, [
 	return resp.InputIds, tokenizersOffsets, nil
 }
 
-// RenderChat renders a chat template using the UDS tokenizer service.
+// RenderChat renders a chat completion request using the UDS renderer service.
+// It calls the RenderChatCompletion RPC which runs vLLM's OpenAIServingRender
+// on the CPU, returning token IDs directly.
 func (u *UdsTokenizer) RenderChat(
```
gyliu513, line 251: offset is always nil, does this still needed?

sagearc: Good catch, opened an issue for this https://github.com/llm-d/llm-d-kv-cache/issues/462

pkg/tokenization/uds_tokenizer.go
sagearc: Good catch, opened an issue for this https://github.com/llm-d/llm-d-kv-cache/issues/462

services/uds_tokenizer/tokenizer_grpc_service.py
sagearc: Since we're already migrating to async [here](https://github.com/llm-d/llm-d-kv-cache/pull/447), would it make more sense to add it there?

pkg/tokenization/uds_tokenizer_test.go
sagearc: done https://github.com/llm-d/llm-d-kv-cache/pull/432/changes/f6a411ee1434c4e03ad3ca88a3d0f0881d045c85

services/uds_tokenizer/tokenizer_grpc_service.py
sagearc: Fixed, made `add_generation_prompt` optional bool in the proto so `HasField` can distinguish "not set" (defaults to True, matching vLLM) from an explicit false. Thanks

sagearc: Also validated against all other default values in vLLM's ChatCompletion request.

services/uds_tokenizer/tokenizer_grpc_service.py
vMaroon: Let's follow-up separately if needed.

Issue #396, opened 2026-03-05
Add support for externally invalidate/reset the KV cache


ezrasilvera:
There are several use cases that require the ability to invalidate the KV-Cache. For example:
1. RL post training - after we update the LLM weights (usually in-place) we probably need to invalidate the references 
2. When working with evolve techniques (e.g., ADRS) - we sometimes just want to perform many loops in which we restart the function we evolve (e.g., routing, scoring, etc..) but don't want to restart the system as we need those loops to be very fast. After each such evolve the KV cache must be invalidated.

vMaroon:
Thanks @ezrasilvera - this effectively means we need to handle `AllBlocksClearedEvent`, and make sure that vLLM emits it properly.

yash9263:
Hi @kapiljain1989, I’ve started working on this and have been going through the details.

PR #437, opened 2026-03-18
feat:add support to invalidate KV cache via AllBlocksCleared event

yash9263:
Issue: #396 

This PR adds support to Invalidate KV cache via the `AllBlocksCleared` event.
- Updated event processing in vllm_adapter.go and pool.go to parse and handle the `AllBlocksCleared` event.
- Added Clear method to the `Index` interface and implemented it for all the index types (CostAwareMemoryIndex, InMemoryIndex, RedisIndex).
- Introduced podToRequestKeys reverse index: podIdentifier -> []engineKeyToRequestKeyMapping for all the indexes.
  - Clear: Looks up only the request keys for the target pod via podToRequestKeys
  - Iterates those keys, and removes matching entries, prunes empty hashes + their engine-key mappings
  - Updates the reverse index: drops the pod entirely if DeviceTier is empty, or trims to still-alive request keys if a specific tier was targeted

Testing
- `pkg/kvcache/kvblock/index_test.go`

Examples:
- Updated the `kv_events/offline` and `valkey_example` to simulate the Clear event.

### DIFF
--- a/pkg/kvcache/kvblock/cost_aware_memory.go
+++ b/pkg/kvcache/kvblock/cost_aware_memory.go
@@ -80,6 +80,7 @@ func NewCostAwareMemoryIndex(cfg *CostAwareMemoryIndexConfig) (*CostAwareMemoryI
 	return &CostAwareMemoryIndex{
 		data:        cache,
 		requestKeys: requestKeys,
+		sweepCh:     make(chan struct{}, 1),
 	}, nil
 }
 
@@ -442,9 +443,7 @@ func (m *CostAwareMemoryIndex) StartSweeper(ctx context.Context, debounce time.D
 	if debounce <= 0 {
 		debounce = 100 * time.Millisecond
 	}
-	if m.sweepCh == nil {
-		m.sweepCh = make(chan struct{}, 1)
-	}
+
 	for {
 		select {
 		case <-ctx.Done():
--- a/pkg/kvcache/kvblock/cost_aware_memory.go
+++ b/pkg/kvcache/kvblock/cost_aware_memory.go
@@ -365,9 +365,8 @@ func (m *CostAwareMemoryIndex) evictPodsFromRequestKey(
 	}
 }
 
-// Clear invalidates all entries for the given podEntry by bumping the pod's
-// generation counter. Stale entries are filtered at Lookup time and reclaimed
-// lazily by ristretto's cost-based eviction (or eagerly via Sweep). O(1).
+// Clear bumps the pod's generation counter, invalidating all prior entries for
+// that pod lazily. Reclaimed by Sweep or by ristretto's cost-based eviction. O(1).
 func (m *CostAwareMemoryIndex) Clear(ctx context.Context, podEntry PodEntry) error {
 	m.gen.bump(podEntry.PodIdentifier)
 	log.FromContext(ctx).V(logging.TRACE).WithName("kvblock.CostAwareMemoryIndex.Clear").
@@ -381,9 +380,8 @@ func (m *CostAwareMemoryIndex) Clear(ctx context.Context, podEntry PodEntry) err
 	return nil
 }
 
-// Sweep performs an immediate scan over all live request-keys, removing entries
-// whose stamped generation is below their pod's current generation. Returns the
-// count of removed entries.
+// Sweep removes entries whose stamped generation is below their pod's current
+// generation. Returns the count removed. O(N) over the index; off the hot path.
 func (m *CostAwareMemoryIndex) Sweep(ctx context.Context) int {
 	traceLogger := log.FromContext(ctx).V(logging.TRACE).WithName("kvblock.CostAwareMemoryIndex.Sweep")
 	if m.gen.anyClears() == 0 {
@@ -438,8 +436,8 @@ func (m *CostAwareMemoryIndex) Sweep(ctx context.Context) int {
 	return removed
 }
 
-// StartSweeper runs Sweep on every Clear, debounced by `debounce`. Returns when
-// ctx is cancelled.
+// StartSweeper runs Sweep on every Clear, debounced and coalesced.
+// Returns when ctx is cancelled. NewIndex starts this by default.
 func (m *CostAwareMemoryIndex) StartSweeper(ctx context.Context, debounce time.Duration) {
 	if debounce <= 0 {
 		debounce = 100 * time.Millisecond
--- a/pkg/kvcache/kvblock/cost_aware_memory.go
+++ b/pkg/kvcache/kvblock/cost_aware_memory.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -94,10 +95,16 @@ type CostAwareMemoryIndex struct {
 	data *ristretto.Cache[string, *CostPodCache]
 	// requestKeys holds the mapping of engine keys to request keys.
 	requestKeys *lru.Cache[BlockHash, []BlockHash]
+	// keyIndex tracks the set of live request-key strings so Sweep can iterate.
+	// Ristretto does not expose iteration; we maintain this set on Add and
+	// prune it lazily during Sweep.
+	keyIndex sync.Map // map[string]struct{}
 	// mu protects concurrent access to the index operations
 	mu sync.RWMutex
 	// gen tracks per-pod generation counters for O(1) Clear via lazy invalidation.
 	gen podGenTracker
+	// sweepCh is signalled by Clear when a background sweeper is running.
+	sweepCh chan struct{}
 }
 
 func (m *CostAwareMemoryIndex) MaxCost() int64 {
@@ -218,6 +225,7 @@ func (m *CostAwareMemoryIndex) Add(ctx context.Context, engineKeys, requestKeys 
 		// Calculate the actual cost for this cache entry
 		cost := podCache.CalculateByteSize(keyStr)
 		m.data.Set(keyStr, podCache, cost)
+		m.keyIndex.Store(keyStr, struct{}{})
 		traceLogger.Info("added pods to key", "requestKey", requestKey, "pods", entries, "cost-bytes", cost)
 	}
 	m.data.Wait()
@@ -239,6 +247,14 @@ func (m *CostAwareMemoryIndex) Lookup(ctx context.Context, requestKeys []BlockHa
 	podsPerKey := make(map[BlockHash][]PodEntry)
 	highestHitIdx := 0
 
+	// Fast-path predicates evaluated once per call.
+	filterPodSet := podIdentifierSet.Len() != 0
+	needGenFilter := m.gen.anyClears() != 0
+	var gens *genCache
+	if needGenFilter {
+		gens = m.gen.snapshot()
+	}
+
 	for idx, key := range requestKeys {
 		keyStr := key.String()
 		if pods, found := m.data.Get(keyStr); found { //nolint:nestif // TODO: can this be optimized?
@@ -249,10 +265,6 @@ func (m *CostAwareMemoryIndex) Lookup(ctx context.Context, requestKeys []BlockHa
 
 			highestHitIdx = idx
 
-			// Filter pods based on the provided pod identifiers AND the per-pod generation
-			// counter. An entry whose stamped generation is below the current generation
-			// for its pod is considered cleared.
-			filterPodSet := podIdentifierSet.Len() != 0
 			pods.cache.Range(func(k, value interface{}) bool {
 				pod, ok := k.(PodEntry)
 				if !ok {
@@ -261,9 +273,11 @@ func (m *CostAwareMemoryIndex) Lookup(ctx context.Context, requestKeys []BlockHa
 				if filterPodSet && !podIdentifierSet.Has(pod.PodIdentifier) {
 					return true
 				}
-				stampedGen, _ := value.(uint64)
-				if stampedGen < m.gen.current(pod.PodIdentifier) {
-					return true
+				if needGenFilter {
+					stampedGen, _ := value.(uint64)
+					if stampedGen < gens.current(pod.PodIdentifier) {
+						return true
+					}
 				}
 				podsPerKey[key] = append(podsPerKey[key], pod)
 				return true
@@ -343,6 +357,7 @@ func (m *CostAwareMemoryIndex) evictPodsFromRequestKey(
 
 	if podCache.Len() == 0 {
 		m.data.Del(keyStr)
+		m.keyIndex.Delete(keyStr)
 		traceLogger.Info("removed requestKey from index as no pods remain", "requestKey", requestKey)
 	} else if podCacheLenBefore != podCache.Len() {
 		m.data.Set(keyStr, podCache, podCache.CalculateByteSize(keyStr))
@@ -352,12 +367,106 @@ func (m *CostAwareMemoryIndex) evictPodsFromRequestKey(
 
 // Clear invalidates all entries for the given podEntry by bumping the pod's
 // generation counter. Stale entries are filtered at Lookup time and reclaimed
-// lazily by ristretto's cost-based eviction. O(1).
+// lazily by ristretto's cost-based eviction (or eagerly via Sweep). O(1).
 func (m *CostAwareMemoryIndex) Clear(ctx context.Context, podEntry PodEntry) error {
 	m.gen.bump(podEntry.PodIdentifier)
 	log.FromContext(ctx).V(logging.TRACE).WithName("kvblock.CostAwareMemoryIndex.Clear").
 		Info("bumped pod generation", "pod", podEntry.PodIdentifier)
+	if m.sweepCh != nil {
+		select {
+		case m.sweepCh <- struct{}{}:
+		default:
+		}
+	}
 	return nil
+}
+
+// Sweep performs an immediate scan over all live request-keys, removing entries
+// whose stamped generation is below their pod's current generation. Returns the
+// count of removed entries.
+func (m *CostAwareMemoryIndex) Sweep(ctx context.Context) int {
+	traceLogger := log.FromContext(ctx).V(logging.TRACE).WithName("kvblock.CostAwareMemoryIndex.Sweep")
+	if m.gen.anyClears() == 0 {
+		return 0
+	}
+	gens := m.gen.snapshot()
+
+	removed := 0
+	m.keyIndex.Range(func(k, _ any) bool {
+		keyStr, ok := k.(string)
+		if !ok {
+			return true
+		}
+		m.mu.Lock()
+		defer m.mu.Unlock()
+
+		podCache, found := m.data.Get(keyStr)
+		if !found || podCache == nil {
+			m.keyIndex.Delete(keyStr)
+			return true
+		}
+
+		var toDelete []PodEntry
+		podCache.cache.Range(func(ek, ev any) bool {
+			entry, ok := ek.(PodEntry)
+			if !ok {
+				return true
+			}
+			stamped, _ := ev.(uint64)
+			if stamped < gens.current(entry.PodIdentifier) {
+				toDelete = append(toDelete, entry)
+			}
+			return true
+		})
+		for _, e := range toDelete {
+			podCache.Delete(e)
+		}
+		removed += len(toDelete)
+		if podCache.Len() == 0 {
+			m.data.Del(keyStr)
+			m.keyIndex.Delete(keyStr)
+		} else if len(toDelete) > 0 {
+			m.data.Set(keyStr, podCache, podCache.CalculateByteSize(keyStr))
+		}
+		return true
+	})
+
+	m.data.Wait()
+	if removed > 0 {
+		traceLogger.Info("sweep removed stale entries", "removed", removed)
+	}
+	return removed
+}
+
+// StartSweeper runs Sweep on every Clear, debounced by `debounce`. Returns when
+// ctx is cancelled.
+func (m *CostAwareMemoryIndex) StartSweeper(ctx context.Context, debounce time.Duration) {
+	if debounce <= 0 {
+		debounce = 100 * time.Millisecond
+	}
+	if m.sweepCh == nil {
+		m.sweepCh = make(chan struct{}, 1)
+	}
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-m.sweepCh:
+			timer := time.NewTimer(debounce)
+			drained := false
+			for !drained {
+				select {
+				case <-m.sweepCh:
+				case <-timer.C:
+					drained = true
+				case <-ctx.Done():
+					timer.Stop()
+					return
+				}
+			}
+			m.Sweep(ctx)
+		}
+	}
 }
 
 // GetRequestKey returns the last request key (highest index in the chain) associated with the given engineKey.
--- a/pkg/kvcache/kvblock/cost_aware_memory.go
+++ b/pkg/kvcache/kvblock/cost_aware_memory.go
@@ -96,6 +96,8 @@ type CostAwareMemoryIndex struct {
 	requestKeys *lru.Cache[BlockHash, []BlockHash]
 	// mu protects concurrent access to the index operations
 	mu sync.RWMutex
+	// gen tracks per-pod generation counters for O(1) Clear via lazy invalidation.
+	gen podGenTracker
 }
 
 func (m *CostAwareMemoryIndex) MaxCost() int64 {
@@ -103,15 +105,18 @@ func (m *CostAwareMemoryIndex) MaxCost() int64 {
 }
 
 // CostPodCache wraps a sync.Map of PodEntry and provides cost calculation for memory usage estimation.
+// The map value is the generation at which the entry was admitted (see podGenTracker).
 type CostPodCache struct {
-	cache sync.Map // map[PodEntry]struct{}
+	cache sync.Map // map[PodEntry]uint64
 	// size tracks the number of entries in cache for O(1) Len().
 	size atomic.Int64
 }
 
-// Add adds a PodEntry to the cache.
-func (c *CostPodCache) Add(entry PodEntry) {
-	if _, loaded := c.cache.LoadOrStore(entry, struct{}{}); !loaded {
+// Add adds (or refreshes) a PodEntry in the cache, stamped with the supplied generation.
+// On re-add of an existing entry, the stored generation is overwritten so post-Clear
+// re-admissions become visible at Lookup time.
+func (c *CostPodCache) Add(entry PodEntry, gen uint64) {
+	if _, loaded := c.cache.Swap(entry, gen); !loaded {
 		c.size.Add(1)
 	}
 }
@@ -207,7 +212,7 @@ func (m *CostAwareMemoryIndex) Add(ctx context.Context, engineKeys, requestKeys 
 		}
 
 		for _, entry := range entries {
-			podCache.Add(entry)
+			podCache.Add(entry, m.gen.current(entry.PodIdentifier))
 		}
 
 		// Calculate the actual cost for this cache entry
@@ -244,25 +249,25 @@ func (m *CostAwareMemoryIndex) Lookup(ctx context.Context, requestKeys []BlockHa
 
 			highestHitIdx = idx
 
-			if podIdentifierSet.Len() == 0 {
-				// If no pod identifiers are provided, return all pods
-				pods.cache.Range(func(k, value interface{}) bool {
-					if pod, ok := k.(PodEntry); ok {
-						podsPerKey[key] = append(podsPerKey[key], pod)
-					}
+			// Filter pods based on the provided pod identifiers AND the per-pod generation
+			// counter. An entry whose stamped generation is below the current generation
+			// for its pod is considered cleared.
+			filterPodSet := podIdentifierSet.Len() != 0
+			pods.cache.Range(func(k, value interface{}) bool {
+				pod, ok := k.(PodEntry)
+				if !ok {
 					return true
-				})
-			} else {
-				// Filter pods based on the provided pod identifiers
-				pods.cache.Range(func(k, value interface{}) bool {
-					if pod, ok := k.(PodEntry); ok {
-						if podIdentifierSet.Has(pod.PodIdentifier) {
-							podsPerKey[key] = append(podsPerKey[key], pod)
-						}
-					}
+				}
+				if filterPodSet && !podIdentifierSet.Has(pod.PodIdentifier) {
 					return true
-				})
-			}
+				}
+				stampedGen, _ := value.(uint64)
+				if stampedGen < m.gen.current(pod.PodIdentifier) {
+					return true
+				}
+				podsPerKey[key] = append(podsPerKey[key], pod)
+				return true
+			})
 		} else {
 			traceLogger.Info("key not found in index", "key", key)
 		}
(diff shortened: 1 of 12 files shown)


pkg/kvcache/kvblock/in_memory.go
```diff
@@ -161,6 +169,7 @@ func (m *InMemoryIndex) Add(ctx context.Context, engineKeys, requestKeys []Block
 
 	for i, requestKey := range requestKeys {
 		// 1. Store engineKey -> requestKey mapping (only if engineKeys provided)
+		curEngineKey := EmptyBlockHash
```
gyliu513: `curEngineKey` will always be empty for your logic. I think it should be set below

```python
curEngineKey := EmptyBlockHash
if engineKeys != nil {
    curEngineKey = engineKeys[i] // << add this line
    m.engineToRequestKeys.Add(engineKeys[i], requestKey)
}
```

pkg/kvcache/kvblock/in_memory.go
```diff
@@ -304,3 +343,62 @@ func podsPerKeyPrintHelper(ks map[BlockHash][]PodEntry) string {
...
+		cache.mu.Unlock()
+
+		if isEmpty {
+			m.data.Remove(entry.requestKey)
+			if entry.engineKey != EmptyBlockHash {
+				if _, exists := m.engineToRequestKeys.Peek(entry.engineKey); exists {
+					m.engineToRequestKeys.Remove(entry.engineKey)
+				}
+			}
+		}
```
gyliu513: there maybe a race condition here as `Add` goroutine may add add another entry after `cache.mu.Unlock()`. If you call `m.data.Remove(entry.requestKey)` directly, lookup will not able to get this entry and this will low down the prefix cache hit ratio.

```python
if isEmpty {
      currentCache, stillExists := m.data.Get(entry.requestKey)
       if !stillExists || currentCache == nil {
                continue
        }
        currentCache.mu.Lock()
        if currentCache.cache.Len() == 0 {
              m.data.Remove(entry.requestKey)
               if entry.engineKey != EmptyBlockHash {
                    m.engineToRequestKeys.Remove(entry.engineKey)
               }
         }
         currentCache.mu.Unlock()
}
```

pkg/kvcache/kvblock/cost_aware_memory.go
```diff
@@ -333,3 +370,50 @@ func (m *CostAwareMemoryIndex) GetRequestKey(ctx context.Context, engineKey Bloc
...
+			if entry, ok := key.(PodEntry); ok {
+				if podEntry.DeviceTier != "" && entry.DeviceTier != podEntry.DeviceTier {
+					return true
+				}
+				if entry.PodIdentifier == podEntry.PodIdentifier {
+					pod.Delete(entry)
+				}
+			}
+			return true
+		})
```
gyliu513: https://pkg.go.dev/github.com/zijiren233/gencontainer/rwmap#:~:text=Range%20does%20not%20necessarily%20correspond,call%20any%20method%20on%20m.

Range does not necessarily correspond to any consistent snapshot of the Map's contents: no key will be visited more than once, but if the value for any key is stored or deleted concurrently (including by f), Range may reflect any mapping for that key from any point during the Range call. Range does not block other methods on the receiver; even f itself may call any method on m. I suggest collect first and then delete.

```python
var toDelete []PodEntry
pod.cache.Range(func(key, value any) bool {
    if entry, ok := key.(PodEntry); ok {
        if podEntry.DeviceTier != "" && entry.DeviceTier != podEntry.DeviceTier {
            return true
        }
        if entry.PodIdentifier == podEntry.PodIdentifier {
            toDelete = append(toDelete, entry)
        }
    }
    return true
})
 for _, entry := range toDelete {
    pod.Delete(entry)
}
```

pkg/kvcache/kvblock/cost_aware_memory.go
```diff
@@ -75,12 +76,23 @@ func NewCostAwareMemoryIndex(cfg *CostAwareMemoryIndexConfig) (*CostAwareMemoryI
...
+		data:            cache,
+		requestKeys:     requestKeys,
+		podToRequestKey: podToRequestKey,
 	}, nil
 }
 
+type engineKeyToRequestKeyMapping struct {
+	engineKey  BlockHash
+	requestKey BlockHash
+}
```
gyliu513: this is not needed based on my comments for add/evict/clear

pkg/kvcache/kvblock/in_memory.go
```diff
@@ -202,6 +211,22 @@ func (m *InMemoryIndex) Add(ctx context.Context, engineKeys, requestKeys []Block
...
+						found = true
+						break
+					}
+				}
+				if !found {
+					m.podToRequestKeys.Add(entry.PodIdentifier, append(existing, engineKeyToRequestKeyMapping{requestKey: requestKey, engineKey: curEngineKey}))
+				}
+			} else {
+				m.podToRequestKeys.Add(entry.PodIdentifier, []engineKeyToRequestKeyMapping{{requestKey: requestKey, engineKey: curEngineKey}})
+			}
```
gyliu513:
```python
  mappings, ok := m.podToRequestKeys.Peek(entry.PodIdentifier)
  if !ok {
      mappings = make(map[BlockHash]BlockHash)
  }
  mappings[requestKey] = curEngineKey 
  m.podToRequestKeys.Add(entry.PodIdentifier, mappings)
```

pkg/kvcache/kvblock/in_memory.go
```diff
@@ -253,6 +278,20 @@ func (m *InMemoryIndex) Evict(ctx context.Context, key BlockHash, keyType KeyTyp
...
+					updated = append(updated, rk)
+				}
+			}
+			if len(updated) == 0 {
+				m.podToRequestKeys.Remove(entry.PodIdentifier)
+			} else {
+				m.podToRequestKeys.Add(entry.PodIdentifier, updated)
+			}
+		}
 	}
```
gyliu513:
```python
if mappings, ok := m.podToRequestKeys.Peek(entry.PodIdentifier); ok {
      delete(mappings, requestKey)
      if len(mappings) == 0 {
          m.podToRequestKeys.Remove(entry.PodIdentifier)
      }
  }
```

pkg/kvcache/kvblock/cost_aware_memory.go
```diff
@@ -75,12 +76,23 @@ func NewCostAwareMemoryIndex(cfg *CostAwareMemoryIndexConfig) (*CostAwareMemoryI
 		return nil, fmt.Errorf("failed to initialize in-memory engine key map: %w", err)
 	}
 
+	podToRequestKey, err := lru.New[string, []engineKeyToRequestKeyMapping](defaultNumCounters)
```
gyliu513:
```python
podToRequestKey, err := lru.New[string, map[BlockHash]BlockHash](defaultNumCounters)
```

pkg/kvcache/kvblock/cost_aware_memory.go
```diff
@@ -193,8 +208,25 @@ func (m *CostAwareMemoryIndex) Add(ctx context.Context, engineKeys, requestKeys
...
+				engineKey:  curEngineKey,
+				requestKey: requestKey,
+			}
+			podIdentifier := entry.PodIdentifier
+			mappings, found := m.podToRequestKey.Peek(podIdentifier)
+			if !found {
+				m.podToRequestKey.Add(podIdentifier, []engineKeyToRequestKeyMapping{mapping})
+			} else {
+				m.podToRequestKey.Add(podIdentifier, append(mappings, mapping))
+			}
```
gyliu513: You may have duplicate entries for one pod, like pod-A report same block after vLLM restart.

1st: 
```
Add(requestKey=100, pod-A):
    mappings = [{requestKey:100, engineKey:50}]
```

vLLM restart:
```
Add(requestKey=100, pod-A):
    mappings = [{requestKey:100, engineKey:50}, {requestKey:100, engineKey:50}]
```


```python
mappings, found := m.podToRequestKey.Peek(entry.PodIdentifier)
if !found {
    mappings = make(map[BlockHash]BlockHash)
}
mappings[requestKey] = curEngineKey
m.podToRequestKey.Add(entry.PodIdentifier, mappings)
```

pkg/kvcache/kvblock/cost_aware_memory.go
```diff
@@ -308,6 +341,10 @@ func (m *CostAwareMemoryIndex) Evict(ctx context.Context, key BlockHash, keyType
 
 	for _, entry := range entries {
 		podCache.Delete(entry)
+
+		if found := m.podToRequestKey.Contains(entry.PodIdentifier); found {
+			m.podToRequestKey.Remove(entry.PodIdentifier)
+		}
 	}
```
gyliu513:
```python
if mappings, found := m.podToRequestKey.Peek(entry.PodIdentifier); found {
    delete(mappings, requestKey)
    if len(mappings) == 0 {
        m.podToRequestKey.Remove(entry.PodIdentifier)
    }
}
```

pkg/kvcache/kvblock/cost_aware_memory.go
gyliu513: Ditto for in_memory.go

pkg/kvcache/kvblock/in_memory.go
gyliu513: @yash9263 I also found your test did not catch this error, I think the test was not covering this case, pleas also consider your test should cover this code path as well.

pkg/kvcache/kvblock/in_memory.go
```diff
@@ -202,6 +212,14 @@ func (m *InMemoryIndex) Add(ctx context.Context, engineKeys, requestKeys []Block
...
 		for _, entry := range entries {
 			podCache.cache.Add(entry, struct{}{})
+
+			// 3. Maintain reverse index: podIdentifier -> [requestKey]: engineKey
+			mappings, ok := m.podToRequestKeys.Peek(entry.PodIdentifier)
+			if !ok {
+				mappings = make(map[BlockHash]BlockHash)
+			}
+			mappings[requestKey] = curEngineKey
+			m.podToRequestKeys.Add(entry.PodIdentifier, mappings)
```
gyliu513: there will be race condition here if two go routine add requestkey for one pod. I think you may want to make a structure change for pod mapping as below:

```python
// podMapping protects the per-pod reverse index map from concurrent access.
// podCache.mu is per-requestKey, so two goroutines handling different requestKeys
// for the same pod would race on the shared map without this lock.
type podMapping struct {
      mu       sync.Mutex
      mappings map[BlockHash]BlockHash // requestKey -> engineKey
}

type InMemoryIndex struct {
      data                *lru.Cache[BlockHash, *PodCache]
      engineToRequestKeys *lru.Cache[BlockHash, BlockHash]
      podToRequestKeys    *lru.Cache[string, *podMapping]  // changed
      podCacheSize        int
}
```

Then for Add in here, the logic will be as below:

```python
podCache.mu.Lock()
for _, entry := range entries {
    podCache.cache.Add(entry, struct{}{})

    pm, ok := m.podToRequestKeys.Get(entry.PodIdentifier)
    if !ok {
        pm = &podMapping{mappings: make(map[BlockHash]BlockHash)}
        if contained, _ := m.podToRequestKeys.ContainsOrAdd(entry.PodIdentifier, pm); contained {
            pm, _ = m.podToRequestKeys.Get(entry.PodIdentifier)
        }
    }
    pm.mu.Lock()
    pm.mappings[requestKey] = curEngineKey
    pm.mu.Unlock()
}
podCache.mu.Unlock()
```

Ditto for evict and clear.

gyliu513: there is no issue for cost_aware_memory.go as it is using a global `sync.RWMutex`.

pkg/kvcache/kvblock/index_test.go
gyliu513: I think `clear` will never be triggered, here should be 
```python
operationIndex % 4
```

pkg/kvcache/kvblock/in_memory.go
```diff
@@ -304,3 +343,87 @@ func podsPerKeyPrintHelper(ks map[BlockHash][]PodEntry) string {
 	}
 	return b.String()
 }
+
+// Clear removes all entries for the given podEntry from the index.
+func (m *InMemoryIndex) Clear(ctx context.Context, podEntry PodEntry) error {
```
gyliu513: For Add and Evict, the lock sequence is 

```python
podCache.mu.Lock()
pm.mu.Lock()
pm.mu.UnLock()
podCache.mu.UnLock()
```
But for clear, it is 

```python
pm.mu.Lock()
podCache.mu.Lock()
podCache.mu.UnLock()
pm.mu.UnLock()
```
This will introduce a dead lock here:

Goroutine A: Add
1. Acquire podCache.mu, hold it
2. Attempt to acquire pm.mu, wait

Goroutine B: Clear
1. Acquire pm.mu, hold it
2. Attempt to acquire podCache.mu, wait

Locked!!

The fix is adjust the lock sequence here, clear should unlock pm.mu first before acquire the podCache.mu.

My suggestion is as following:

```python
pm.mu.Lock()
snapshot := make(map[BlockHash]BlockHash, len(pm.mappings))
for k, v := range pm.mappings {
    snapshot[k] = v
}
pm.mu.Unlock()
for requestKey, engineKey := range snapshot {
    cache, exists := m.data.Peek(requestKey)
    if !exists || cache == nil {
        continue
    }
    cache.mu.Lock()
    // ... delete logic ...
    cache.mu.Unlock()
}
```

pkg/kvcache/kvblock/in_memory.go
```diff
@@ -253,6 +282,16 @@ func (m *InMemoryIndex) Evict(ctx context.Context, key BlockHash, keyType KeyTyp
 	podCache.mu.Lock()
 	for _, entry := range entries {
 		podCache.cache.Remove(entry)
+
+		if pm, ok := m.podToRequestKeys.Peek(entry.PodIdentifier); ok {
+			pm.mu.Lock()
+			delete(pm.mappings, requestKey)
+			pm.mu.Unlock()
+
+			if len(pm.mappings) == 0 {
```
gyliu513: suggest add `shouldRemove := len(pm.mappings) == 0` to pm.mu lock section, as go do not support concurrency read write, here is reading the pm.mapping, you may get panic if there is another goroutine writing the mapping.

Ditto for others if they have same logic.

pkg/kvevents/pool.go
```diff
@@ -296,11 +296,13 @@ func (p *Pool) processEventBatch(ctx context.Context, batch *EventBatch, podIden
...
 
 		case *AllBlocksClearedEvent:
-			debugLogger.Info("All blocks cleared event received",
-				"podIdentifier", podIdentifier,
-				"deviceTier", ev.DeviceTier,
-				"modelName", modelName)
-
+			if err := p.index.Clear(ctx, kvblock.PodEntry{PodIdentifier: podIdentifier, DeviceTier: ev.DeviceTier}); err != nil {
+				debugLogger.Error(err, "Failed to clear all blocks",
+					"deviceTier", ev.DeviceTier,
```
gyliu513: strings.ToLower(ev.DeviceTier)

pkg/kvcache/kvblock/cost_aware_memory.go
```diff
@@ -333,3 +362,71 @@ func (m *CostAwareMemoryIndex) GetRequestKey(ctx context.Context, engineKey Bloc
...
+		}
+		if len(remaining) == 0 {
+			m.podToRequestKeys.Remove(podEntry.PodIdentifier)
+		} else {
+			m.podToRequestKeys.Add(podEntry.PodIdentifier, remaining)
+		}
+	}
+
+	m.data.Wait()
+	traceLogger.Info("Cleared pod entries from InMemoryIndex", "podEntry", podEntry)
```
gyliu513: not `InMemoryIndex` but `CostAwareMemoryIndex`

pkg/kvcache/kvblock/redis.go
```diff
@@ -324,3 +332,100 @@ func (r *RedisIndex) GetRequestKey(ctx context.Context, engineKey BlockHash) (Bl
...
+
+	logger.Info("Cleared pod entries from Redis index", "podEntry", podEntry)
+	return nil
+}
+
+// mustParseUint64 parses a uint64 string, returning 0 on failure.
+func mustParseUint64(s string) uint64 {
+	v, err := strconv.ParseUint(s, 10, 64)
+	if err != nil {
+		return 0
```
gyliu513: why return 0? Does this will be treated as EmptyBlockHash?

gyliu513 (commented):
Good progress, thanks @yash9263 !

pkg/kvcache/kvblock/in_memory.go
yash9263: moved the pm.mu.Unlock() after `m.podToRequestKeys.Remove`, a concurrent `Add` may acquire the lock in the window and add another entry to the pod mapping, which will be dropped by `Remove`.

pkg/kvcache/kvblock/redis.go
yash9263: the `mustParseUint64` shouldn't error, since it's only getting called after checking if the engine key isn't empty, but was getting a lint error for ignoring the error

updated the `redisEngineKey` to return an empty string for the `EmptyBlockHash`.

pkg/kvcache/kvblock/in_memory.go
```diff
@@ -304,3 +343,93 @@ func podsPerKeyPrintHelper(ks map[BlockHash][]PodEntry) string {
...
+		for requestKey, engineKey := range pm.mappings {
+			if exists := m.data.Contains(requestKey); exists {
+				remaining[requestKey] = engineKey
+			}
+		}
+
+		if len(remaining) == 0 {
+			m.podToRequestKeys.Remove(podEntry.PodIdentifier)
+		} else {
+			pm := &podMapping{mappings: remaining}
```
gyliu513: `pm := &podMapping{mappings: remaining}` shadows the outer pm. After the LRU stores the new pointer, concurrent Add goroutines still holding the old pm reference write to a detached object, those writes are silently lost.

```python
if podEntry.DeviceTier == "" {
      m.podToRequestKeys.Remove(podEntry.PodIdentifier)
  } else {
      remaining := make(map[BlockHash]BlockHash)
      pm.mu.Lock()
      for requestKey, engineKey := range pm.mappings {
          if exists := m.data.Contains(requestKey); exists {
              remaining[requestKey] = engineKey
          }
      }

      if len(remaining) == 0 {
          m.podToRequestKeys.Remove(podEntry.PodIdentifier)
      } else {
          pm.mappings = remaining
      }
      pm.mu.Unlock() 
  }
```

pkg/kvcache/kvblock/redis.go
```diff
@@ -322,5 +330,105 @@ func (r *RedisIndex) GetRequestKey(ctx context.Context, engineKey BlockHash) (Bl
...
+		requestKeyStr := meta
+		engineKeyRedisKey := ""
+		if sep >= 0 {
+			requestKeyStr = meta[:sep]
+			if ek := meta[sep+1:]; ek != "" {
+				engineKeyRedisKey = redisEngineKey(BlockHash(mustParseUint64(ek)))
+			}
+		}
+
+		if err := clearPodEntryScript.Run(
```
gyliu513: Each entry triggers a separate clearPodEntryScript.Run(), meaning N round trips for N entries.

How about use pipeline here?

```python
pipe := r.RedisClient.Pipeline()
for .... {
  pipe.EvalSha(ctx, ...)
}
```

pkg/kvcache/kvblock/cost_aware_memory.go
```diff
@@ -333,3 +362,71 @@ func (m *CostAwareMemoryIndex) GetRequestKey(ctx context.Context, engineKey Bloc
...
+	traceLogger := log.FromContext(ctx).V(logging.TRACE).WithName("kvblock.CostAwareMemoryIndex.Clear")
+
+	// Remove all entries for the given pod identifier
+	mappings, found := m.podToRequestKeys.Get(podEntry.PodIdentifier)
+	if !found {
+		traceLogger.Info("pod not found in reverse index, nothing to clear", "podEntry", podEntry)
+		return nil
+	}
+
+	for requestKey, engineKey := range mappings {
```
gyliu513: `mappings` is iterated once to delete entries, then iterated again to build remaining. Can be merged into one pass.

```python
remaining := make(map[BlockHash]BlockHash)
for requestKey, engineKey := range mappings {
    // ... delete matching pod entries ...
    if pod.Len() == 0 {
        m.data.Del(requestKey.String())
        if _, hasEngineKey := m.requestKeys.Get(engineKey); hasEngineKey {
            m.requestKeys.Remove(engineKey)
        }
    } else if podCacheLenBefore != pod.Len() {
        m.data.Set(requestKey.String(), pod, pod.CalculateByteSize(requestKey.String()))
        remaining[requestKey] = engineKey
    } else {
        remaining[requestKey] = engineKey
    }
}

if podEntry.DeviceTier == "" || len(remaining) == 0 {
    m.podToRequestKeys.Remove(podEntry.PodIdentifier)
} else {
    m.podToRequestKeys.Add(podEntry.PodIdentifier, remaining)
}
```

gyliu513:
@vMaroon @sagearc @yankay I think this is in good shape, it is great if you guys can check if this can be merged, thanks!

pkg/kvcache/kvblock/cost_aware_memory.go
```diff
@@ -333,3 +362,65 @@ func (m *CostAwareMemoryIndex) GetRequestKey(ctx context.Context, engineKey Bloc
...
+
+		for _, entry := range toDelete {
+			pod.Delete(entry)
+		}
+
+		if pod.Len() == 0 {
+			m.data.Del(requestKey.String())
+			if _, hasEngineKey := m.requestKeys.Get(engineKey); hasEngineKey {
+				m.requestKeys.Remove(engineKey)
+			}
```
gyliu513: add a checking here `if engineKey != EmptyBlockHash`

gyliu513 (commented):
/lgtm

Thanks @yash9263 !

vMaroon:
Sorry for neglecting this, tough couple of weeks, will take a look ASAP.

yash9263:
I'll resolve the conflicts

vMaroon:
@yash9263 can you profile the performance/memory-footprint of the changes?

vMaroon:
I wonder if we can take a generation counter approach instead:
- Keep a counter per discovered server
- Increment on event
- Attach a generation to the index entries
  - Either switch to the cost-aware index by default, and have its LRU take care of things
  - Add a background routine that evicts outdated entries on generation increase
 -  `kvblock.Scorer` ignores outdated entries

My concerns with this approach are that we're ALWAYS paying a toll for an event that could be infrequent. Even in RL case, it adds overhead to the hot-path. And the added complexity.

yash9263:
> can you profile the performance/memory-footprint of the changes?

@vMaroon  I’ve attached the profile results here using 
`go test -bench=. -benchmem -count=5 ./tests/profiling/kv_cache_index/...`
goos: darwin
goarch: arm64
pkg: github.com/llm-d/llm-d-kv-cache/tests/profiling/kv_cache_index

| Benchmark | ns/op | B/op | allocs/op |
|---|---:|---:|---:|
| InMemory_Add | 13,523,185 | 16,228,293 | 150,408 |
| InMemory_Lookup | 6,097,144 | 4,687,805 | 90,118 |
| InMemory_Clear | 15,193,804 | 1,911,680 | 20,069 |
| Redis_Add | 247,372,395 | 28,880,476 | 828,210 |
| Redis_Lookup | 87,520,894 | 7,577,659 | 230,114 |
| Redis_Clear | 1,512,841,925 | 2,467,301,786 | 8,399,831 |
| CostAware_Add | 31,097,593 | 13,790,711 | 132,468 |
| CostAware_Lookup | 8,879,893 | 4,994,726 | 100,243 |
| CostAware_Clear | 19,402,680 | 2,156,030 | 50,310 |

yash9263:
@vMaroon, I agree with the concerns in the generational counter approach: the generational approach incurs a cost on every Lookup request, and the added complexity of the background sweeper. 

The current Clear keeps the hot path clean and runs only when required, unless pod churn is frequent enough to make the bulk delete a bottleneck.
Happy to pivot to the alternative if that’s the preferred direction. Let me know your thoughts.

vMaroon:
@yash9263 - I prototyped the generation-counter approach with Claude on a branch so we can compare side by side: https://github.com/vMaroon/llm-d-kv-cache/tree/gen-counter-clear and re-ran your benchmark on both.

---
<details>
<summary>(click to expand) full benchmark numbers (M1 Pro, miniredis, -count=5 -benchtime=3x)</summary>

### Latency (ns/op)

| Op | Baseline | PR #437 | gen-counter |
|---|---:|---:|---:|
| InMemory Add     | 10,442,944  | 12,714,625  | 8,919,614   |
| InMemory Lookup  | 6,515,686   | 6,686,664   | 6,034,500   |
| InMemory Clear   | —           | 10,889,311  | 10,533      |
| CostAware Add    | 13,892,919  | 16,316,139  | 13,867,208  |
| CostAware Lookup | 8,700,453   | 8,338,892   | 8,433,686   |
| CostAware Clear  | —           | 16,337,095  | 7,900       |
| Redis Add        | 410,020,461 | 614,371,936 | 430,755,800 |
| Redis Lookup     | 224,152,520 | 221,457,747 | 214,777,083 |
| Redis Clear      | —           | 988,462,758 | 25,497      |

### Memory (B/op)

| Op | Baseline | PR #437 | gen-counter |
|---|---:|---:|---:|
| InMemory Add     | 13,451,606 | 16,228,598    | 13,611,933 |
| InMemory Lookup  | 4,688,251  | 4,687,915     | 4,687,516  |
| InMemory Clear   | —          | 1,911,992     | 875        |
| CostAware Add    | 11,006,722 | 13,784,873    | 12,264,921 |
| CostAware Lookup | 4,932,981  | 4,933,577     | 4,934,339  |
| CostAware Clear  | —          | 2,160,442     | 96         |
| Redis Add        | 21,421,645 | 31,077,666    | 21,578,376 |
| Redis Lookup     | 7,577,526  | 7,577,342     | 7,018,618  |
| Redis Clear      | —          | 2,467,289,346 | 695        |

### Allocations (allocs/op)

| Op | Baseline | PR #437 | gen-counter |
|---|---:|---:|---:|
| InMemory Add     | 130,245 | 150,409   | 130,248 |
| InMemory Lookup  | 90,121  | 90,120    | 90,118  |
| InMemory Clear   | —       | 20,075    | 20      |
| CostAware Add    | 112,297 | 132,462   | 136,190 |
| CostAware Lookup | 100,129 | 100,130   | 100,132 |
| CostAware Clear  | —       | 50,320    | 4       |
| Redis Add        | 523,572 | 833,634   | 533,565 |
| Redis Lookup     | 230,120 | 230,119   | 210,115 |
| Redis Clear      | —       | 8,399,604 | 11      |

</details>

---

- Add (runs on every kv-event): this reverse-index approach adds 17–50% overhead vs baseline; gen-counter stays at baseline.
- Clear (the rollout path): this PR is 10 ms / 16 ms / 988 ms (InMemory / CostAware / Redis); gen-counter is 10 µs / 8 µs / 25 µs.
- Lookup: parity in both.

On the two concerns you raised:

- Lookup overhead - addressed. Fast path when no Clear has happened (single atomic load), gens memoized per call after the first Clear. Hence Lookup parity above.
- Sweeper complexity - fair point, it's net-new code that needs maintenance. Contained though: gen tracker in its own file, Sweep / StartSweeper isolated to each backend, no entanglement with Add/Lookup/Evict semantics.

One caveat on the Redis Clear number: miniredis runs `EVALSHA` through gopher-lua, which is much slower than real Redis, so the absolute 988 ms is inflated. The point isn't the absolute - it's that the proposed approach issues O(pod_entries) script calls per Clear and gen-counter issues one. Real Redis reduces both numbers but the gap stays large I think

Could you check the branch and run the bench on your end? Want to confirm the numbers reproduce before picking a direction. Let me know what you think.

cc @sagearc @gyliu513 @bongwoobak your inputs are welcome.

yash9263:
@vMaroon thanks for prototyping the gen-counter approach, I reproduced the benchmark on my end (M2 Pro, miniredis, -count=5 -benchtime=3x) and the numbers align with yours:
<details>
<summary>benchmarks</summary>

### InMemory Add

| Metric | gen-counter | baseline | PR #437 |
|--------|-------------|----------|---------|
| ns/op | 5,499,639 | 6,971,889 | 8,436,527 |
| B/op | 13,611,936 | 13,451,573 | 16,228,224 |
| allocs/op | 130,248 | 130,245 | 150,408 |

### InMemory Lookup

| Metric | gen-counter | baseline | PR #437 |
|--------|-------------|----------|---------|
| ns/op | 3970347 | 4237083 | 4430917 |
| B/op | 4687989 | 4687984 | 4687890 |
| allocs/op | 90,118 | 90,120 | 90119 |

### InMemory Clear

| Metric | gen-counter | PR #437 |
|--------|-------------|---------|
| ns/op | 2,278 | 5,175,222 |
| B/op | 408 | 1,911,992 |
| allocs/op | 10 | 20,075 |


### CostAware Add

| Metric | gen-counter | baseline | PR #437 |
|--------|-------------|----------|---------|
| ns/op | 14,169,447 | 12,070,305 | 12,573,236 |
| B/op | 12,259,640 | 11,007,474 | 13,781,490 |
| allocs/op | 136,169 | 112,297 | 132,461 |

### CostAware Lookup

| Metric | gen-counter | baseline | PR #437 |
|--------|-------------|----------|---------|
| ns/op | 5,439,528 | 5,403,097 | 5,327,153 |
| B/op | 4,933,373 | 4,932,648 | 4,933,272 |
| allocs/op | 100,129 | 100,127 | 100,130 |

### CostAware Clear

| Metric | gen-counter | PR #437 |
|--------|-------------|---------|
| ns/op | 6,084 | 8,312,014 |
| B/op | 408 | 2,160,965 |
| allocs/op | 10 | 50,323 |

### Redis Add

| Metric | gen-counter | baseline | PR #437 |
|--------|-------------|----------|---------|
| ns/op | 115,743,264 | 120,146,069 | 170,985,264 |
| B/op | 21,578,277 | 21,424,192 | 31,080,216 |
| allocs/op | 533,565 | 523,581 | 833,642 |

### Redis Lookup

| Metric | gen-counter | baseline | PR #437 |
|--------|-------------|----------|---------|
| ns/op | 59,179,472 | 60,174,708 | 60,292,542 |
| B/op | 7,018,744 | 7,578,504 | 7,577,546 |
| allocs/op | 210,116 | 230,124 | 230,120 |

### Redis Clear

| Metric | gen-counter | PR #437 |
|--------|-------------|---------|
| ns/op | 50,208 | 922,452,167 |
| B/op | 61,328 | 2,467,272,621 |
| allocs/op | 259 | 8,399,604 |

</details>

The gen-counter approach clearly wins on `Clear`, and the Add/Lookup is at baseline. I think the gen-counter design is the better direction.
Let me know how you'd like to proceed.

vMaroon:
> @vMaroon thanks for prototyping the gen-counter approach, I reproduced the benchmark on my end (M2 Pro, miniredis, -count=5 -benchtime=3x) and the numbers align with yours:
> 
> benchmarks
> The gen-counter approach clearly wins on `Clear`, and the Add/Lookup is at baseline. I think the gen-counter design is the better direction. Let me know how you'd like to proceed.

@yash9263 I think repurposing this PR and pivoting to that direction can work out - the branch shared is just a reference, I'm not sure it is completely bug free and thorough evaluation was not made. Implementing the approach and owning it is best. Can add me as a co-author if makes sense.

Does that sound ok?

bongwoobak:
Agree with the gen-counter direction. 
This also complements the dynamic speculative TTL work we're planning: when an AllBlocksCleared event fires, the gen-counter instantly invalidates all entries for that pod including stale speculative entries, without waiting for individual TTL expiry.

One more thing, we're planning to add KV events health monitor that tracks per-endpoint state. When it detects a pod has disappeared from endpointList, it can bump the gen-counter to instantly invalidate all stale entries for that pod, including speculative ones.

vMaroon:
hi @yash9263 - any update on this?

yash9263:
Hi @vMaroon, sorry for the delay, I wasn't around. I've updated this branch with the gen-counter commits and added a few tests on top of it. Let me know your thoughts.

PR #461, opened 2026-03-25
feat(tokenization): replace request_json with native proto fields in render RPCs

### LINKED ISSUE #449: Switch render RPCs from `request_json` to vLLM native gRPC API
Currently `RenderChatCompletion` / `RenderCompletion` pass the OpenAI request as a JSON string in the proto. We should switch to vLLM's native gRPC API once the direction is settled.

ref:
https://github.com/llm-d/llm-d-kv-cache/pull/428#discussion_r2963762313

sagearc:
`RenderChatCompletionRequest` and `RenderCompletionRequest` previously used a `request_json` string field, requiring callers to manually marshal OpenAI types to JSON. This replaces those fields with native proto types so callers work directly with structured proto messages.

`RenderChatCompletionRequest` now has `messages`, `tools`, `chat_template`, `add_generation_prompt`, `continue_final_message`, and `chat_template_kwargs`. `RenderCompletionRequest` now has `prompt` and `model_name`. `ChatMessage` is also extended to support multimodal content via `content_parts` (text + image_url blocks), consistent with the multimodal work in #219.

Ref: https://github.com/vllm-project/vllm/pull/36102
Closes #449

### DIFF
--- a/pkg/tokenization/uds_tokenizer.go
+++ b/pkg/tokenization/uds_tokenizer.go
@@ -23,12 +23,14 @@ import (
 	"strings"
 	"time"
 
-	tokenizerpb "github.com/llm-d/llm-d-kv-cache/api/tokenizerpb"
-	types "github.com/llm-d/llm-d-kv-cache/pkg/tokenization/types"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/keepalive"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	tokenizerpb "github.com/llm-d/llm-d-kv-cache/api/tokenizerpb"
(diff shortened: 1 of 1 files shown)


pkg/tokenization/uds_tokenizer.go
```diff
@@ -235,10 +235,28 @@ func (u *UdsTokenizer) RenderChat(
...
+			for _, block := range msg.Content.Structured {
+				part := &tokenizerpb.ContentPart{Type: block.Type}
+				switch block.Type {
+				case "text":
+					text := block.Text
+					part.Text = &text
+				case "image_url":
+					part.ImageUrl = &tokenizerpb.ImageUrl{Url: block.ImageURL.URL}
+				default:
+					continue
```
vMaroon, line 255: We should have a trace log here.

sagearc: Done

Issue #304, opened 2026-02-11
Update examples to support UDS-only builds (no embedded tokenizers)


sagearc:
**Description**

As part of the effort to isolate and deprecate embedded tokenizers (#295), all example binaries are now gated behind the `embedded_tokenizers` build tag (#303). This means every example requires Python/CGO to build, and there is no reference example demonstrating the recommended UDS tokenizer architecture.

Parent issue: #295
Build tag isolation PR: #303
Related: #298 (E2E tests for UDS tokenizer)

PR #471, opened 2026-03-29
docs: update examples to use UDS tokenizer Docker image

sagearc:
## Summary

- Switch examples to use the UDS tokenizer Docker image instead of the embedded tokenizer
- CI builds the tokenizer image with GHA layer caching and runs it as a container during example verification
- `run-example` Makefile target starts the tokenizer container, waits for health, runs the example with `TOKENIZER_ENDPOINT`, then tears it down
- Clean up example code to align with current API surface
- Use `Qwen/Qwen2-VL-7B-Instruct` as the test model (replace encoder-only `bert-base-uncased`)

## Test plan
- [x] CI examples workflow passes
- [x] `make run-example` works end to end locally

closes https://github.com/llm-d/llm-d-kv-cache/issues/304

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -678,7 +678,7 @@ start-tokenizer: check-container-tool ## Start the UDS tokenizer container; requ
 	done
 
 .PHONY: stop-tokenizer
-stop-tokenizer: check-container-tool ## Stop and remove the UDS tokenizer container
+stop-tokenizer: ## Stop and remove the UDS tokenizer container
 	@$(CONTAINER_TOOL) stop uds-tokenizer-example 2>/dev/null || true
 	@$(CONTAINER_TOOL) rm -f uds-tokenizer-example 2>/dev/null || true
 
@@ -688,8 +688,8 @@ run-example-only: $(EXAMPLE) ## Run the example binary only (tokenizer must alre
 	@TOKENIZER_ENDPOINT=localhost:$(UDS_TOKENIZER_GRPC_PORT) ./$(EXAMPLE)
 
 .PHONY: run-example
-run-example: check-container-tool $(EXAMPLE) ## Run the example with UDS tokenizer in Docker (e.g., make run-example offline); requires image-build-uds to have been run first
+run-example: ## Run the example with UDS tokenizer in Docker (e.g., make run-example offline); requires image-build-uds to have been run first
 	@$(MAKE) --no-print-directory start-tokenizer
-	@TOKENIZER_ENDPOINT=localhost:$(UDS_TOKENIZER_GRPC_PORT) ./$(EXAMPLE); status=$$?; \
+	@$(MAKE) --no-print-directory run-example-only; status=$$?; \
 		$(MAKE) --no-print-directory stop-tokenizer; \
 		exit $$status


Makefile
```diff
@@ -655,7 +658,25 @@ EXAMPLE_SHORTS := offline online valkey kv_cache_index kv_cache_index_service
 $(EXAMPLE_SHORTS):
 
 .PHONY: run-example
-run-example: $(EXAMPLE) ## Run the example locally (e.g., make run-example offline)
+run-example: $(EXAMPLE) ## Run the example with UDS tokenizer in Docker (e.g., make run-example offline); requires image-build-uds to have been run first
+	@printf "\033[33;1m==== Starting UDS tokenizer container ====\033[0m\n"
+	@$(CONTAINER_TOOL) run -d --rm --name uds-tokenizer-example --network host \
+		-e GRPC_PORT=$(UDS_TOKENIZER_GRPC_PORT) \
+		-e PROBE_PORT=$(UDS_TOKENIZER_HEALTH_PORT) \
+		$(UDS_TOKENIZER_IMAGE)
```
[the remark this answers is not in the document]
sagearc: @copilot apply changes based on [this feedback](https://github.com/llm-d/llm-d-kv-cache/pull/471#discussion_r3005986411)

hack/verify-examples.sh
```diff
@@ -15,88 +23,58 @@ fi
...
-    kill -INT $pid || true
-    wait $pid || true
-  else
-    kill -INT $pid || true
-    wait $pid || true
-    cat offline.log
-    fail "offline example did not complete successfully."
+echo "[INFO] Running offline example..."
+timeout 30s make run-example offline >offline.log 2>&1 &
+pid=$!
```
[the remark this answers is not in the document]
sagearc: Unrelated to this PR

examples/helper/indexer.go
```diff
@@ -38,12 +42,13 @@ func getKVCacheIndexerConfig() (*kvcache.Config, error) {
 
 	config.TokenizersPoolConfig.ModelName = testdata.ModelName
 
-	huggingFaceToken := os.Getenv(envHFToken)
-	if huggingFaceToken != "" {
-		config.TokenizersPoolConfig.HFTokenizerConfig.HuggingFaceToken = huggingFaceToken
+	if endpoint := os.Getenv(envTokenizerEndpoint); endpoint != "" {
+		config.TokenizersPoolConfig.UdsTokenizerConfig = &tokenization.UdsTokenizerConfig{
+			SocketFile: endpoint,
+			UseTCP:     !strings.HasPrefix(endpoint, "/"),
```
gyliu513: this may not be right, as `./tokenizer.socket` will be treated as TCP, how about update the logic as 

```python
func isTCPAddr(s string) bool {
    host, port, err := net.SplitHostPort(s)
    return err == nil && host != "" && port != ""
}
```

hack/verify-examples.sh
```diff
@@ -7,6 +7,14 @@ fail() {
   exit 1
 }
 
+stop_tokenizer() {
+  container_tool="${CONTAINER_TOOL:-docker}"
+  "${container_tool}" rm -f uds-tokenizer-example 2>/dev/null || true
+}
```
gyliu513: nit: stop container first before remove it

```
stop_tokenizer() {
  container_tool="${CONTAINER_TOOL:-docker}"
  "${container_tool}" stop uds-tokenizer-example 2>/dev/null || true
  "${container_tool}" rm -f uds-tokenizer-example 2>/dev/null || true
}
```

Makefile
```diff
@@ -655,7 +658,25 @@ EXAMPLE_SHORTS := offline online valkey kv_cache_index kv_cache_index_service
 $(EXAMPLE_SHORTS):
 
 .PHONY: run-example
-run-example: $(EXAMPLE) ## Run the example locally (e.g., make run-example offline)
+run-example: check-container-tool $(EXAMPLE) ## Run the example with UDS tokenizer in Docker (e.g., make run-example offline); requires image-build-uds to have been run first
```
gyliu513: How about have a fine-graind pattern here?

- run-example-only
- start-tokenizer
- stop-tokenizer

Then in verify-examples.sh:

- make start-tokenizer
- make run-example-only offline
- make run-example-only online
- make run-example-only kv_cache_index
- make stop-tokenizer

With that, CI do not need to start/stop container frequently.

examples/helper/indexer.go
```diff
@@ -19,15 +17,21 @@ package helper
...
 const (
-	envHFToken = "HF_TOKEN"
+	// envTokenizerEndpoint overrides the UDS tokenizer socket path or TCP address.
+	// Use a path (e.g. /tmp/tokenizer/tokenizer-uds.socket) for UDS mode (default),
+	// or host:port (e.g. localhost:50051) for TCP mode (useful when running the
+	// tokenizer as a Docker container).
+	envTokenizerEndpoint = "TOKENIZER_ENDPOINT" //nolint:gosec // env var name, not a credential
 )
 
 func getKVCacheIndexerConfig() (*kvcache.Config, error) {
```
gyliu513, line 53: Can getKVCacheIndexerConfig be unified to cover the use case for kv_cache_index/main.go, online/main.go, valkey_example/main.go? So that those files do not need to maintain their getKVCacheIndexerConfig

examples/helper/indexer.go
```diff
@@ -38,12 +42,13 @@ func getKVCacheIndexerConfig() (*kvcache.Config, error) {
 
 	config.TokenizersPoolConfig.ModelName = testdata.ModelName
 
-	huggingFaceToken := os.Getenv(envHFToken)
-	if huggingFaceToken != "" {
-		config.TokenizersPoolConfig.HFTokenizerConfig.HuggingFaceToken = huggingFaceToken
+	if endpoint := os.Getenv(envTokenizerEndpoint); endpoint != "" {
```
gyliu513: Can this be moved to helper so tha others can call `helper.ApplyTokenizerEndpoint(config)`?

```python
const (
      // EnvTokenizerEndpoint is the env var for the UDS tokenizer socket path or TCP address.
      // Use a path (e.g. /tmp/tokenizer/tokenizer-uds.socket) for UDS mode,
      // or host:port (e.g. localhost:50051) for TCP mode.
      EnvTokenizerEndpoint = "TOKENIZER_ENDPOINT" //nolint:gosec // env var name, not a credential
)

// ApplyTokenizerEndpoint reads TOKENIZER_ENDPOINT and sets UDS config on the given config.
func ApplyTokenizerEndpoint(config *kvcache.Config) {
      endpoint := os.Getenv(EnvTokenizerEndpoint)
      if endpoint == "" {
              return
      }
      config.TokenizersPoolConfig.UdsTokenizerConfig = &tokenization.UdsTokenizerConfig{
              SocketFile: endpoint,
              UseTCP:     !strings.HasPrefix(endpoint, "/"),
      }
}
```

examples/helper/indexer.go
sagearc: Fixed

hack/verify-examples.sh
sagearc: Fixed

Makefile
sagearc: Yeah this is cleaner, done

examples/helper/indexer.go
sagearc: Done with helper.ApplyTokenizerEndpoint

examples/helper/indexer.go
sagearc: Done

PR #473, opened 2026-03-29
fix: deprecate embedded tokenizers

sagearc:
## Summary

Removes the embedded tokenizer code (python/cgo) that has been behind build tags since #303. The scheduler dropped cgo entirely in [scheduler#609](https://github.com/llm-d/llm-d-inference-scheduler/pull/609), and the embedded path was never updated for multimodal support — so there's no realistic way to use it and no plan to maintain it.

## Changes

- Delete `pkg/preprocessing/chat_completions` and `pkg/tokenization/pool_embedded*` / `embedded_tokenizer*`
- Drop the `build-embedded` CI job and its apt/pip/ZMQ setup
- Remove `!embedded_tokenizers` build tag from `pool_uds.go` — UDS is now the only pool

## Ref

- [kv-cache#303](https://github.com/llm-d/llm-d-kv-cache/pull/303) (isolated behind build tags)
- [llm-d-inference-scheduler#609](https://github.com/llm-d/llm-d-inference-scheduler/pull/609) (scheduler dropped cgo)
- [kv-cache#295](https://github.com/llm-d/llm-d-kv-cache/issues/295) (RFC tracking issue)


gyliu513 (commented):
Nice cleanup, just one comment, please remove `build-embedded` from `ci-test.yaml`. It is weird that CI did not fail...

yankay (commented):
Thanks @sagearc — nice cleanup!

One issue: `make build` (`build-uds`) now only runs `go build ./pkg/...` which doesn't produce a binary. This breaks:

1. **`make run`** — `./bin/llm-d-kv-cache` doesn't exist
2. **Dockerfile** — `COPY --from=builder /workspace/bin/llm-d-kv-cache` fails

The old `build-embedded` had `-o bin/$(PROJECT_NAME)` but that was removed without a replacement. Suggest adding `-o` to `build-uds` or updating the Dockerfile accordingly.

sagearc:
@yankay  Thanks for catching this! Sorry it took me a while to get to it, fixed now.

yankay (commented):
Thanks @sagearc — the build fix looks good! `build-uds` now correctly produces the binary in `bin/`, so `make run` and Dockerfile both work.

/lgtm

yankay:
Nit: there are 15 commits on this branch — it would be nice to squash them into one before merging for a cleaner history.

vMaroon (approved):

PR #476, opened 2026-03-29
feat: Add HMA support to FS connector

### LINKED ISSUE #472: Add HMA support to FS connector via vllm offloading connector HMA support
## Summary

Add HMA (Hybrid Memory Architecture) support to the llmd-fs-backend by adopting the new vllm offloading connector HMA interfaces.

This enables KV cache offloading for models with multiple KV cache groups, such as full attention + sliding window + Mamba (e.g., Jamba, Zamba, Command-A).

Related to vllm-project/vllm#33689 (KV Offloading Roadmap — HMA Support).

## What needs to happen

- Adopt new vllm HMA interfaces: `CanonicalKVCaches`, `OffloadKey`, updated `OffloadingManager` API
- Update FS connector worker, manager, spec, and mediums modules
- Update tests

kfirtoledo:
## Summary

Add Hybrid Memory Architecture (HMA) support to the llmd-fs-backend by adopting the new vllm offloading connector HMA interfaces.

HMA models use multiple, heterogeneous attention mechanisms across layers (e.g., full attention + sliding window + Mamba/SSM). This introduces multiple KV cache groups, each with potentially different block sizes and lookup strategies.

Key vllm interface changes adopted:
- `CanonicalKVCaches` replaces `dict[str, torch.Tensor]` + `AttentionBackend` — vllm now canonicalizes all KV tensors to `(num_blocks, page_size_bytes)` before handing them to connectors, removing the need for per-backend layout detection. This simplifies **worker.py** by removing `_get_tensors()` (~75 lines of layout detection).
- `OffloadKey` replaces `BlockHash` — encodes both `block_hash` and `group_idx` so blocks from different groups are distinguishable. Updated across **manager.py**, **mediums.py**, and **worker.py**.
- `OffloadingManager.lookup()` changed from `Iterable[BlockHash] -> int` to single `OffloadKey -> int | None`
- `PrepareStoreOutput` fields renamed (`keys_to_store`, `evicted_keys`)

- **spec.py** now resolves `gpu_block_size` from attention groups only, since non-attention groups (Mamba/SSM) may have incompatible block sizes. The fs connector correctly handles standard models, HMA hybrids (e.g., Qwen3.5-27B), and pure Mamba models.
- README updated with HMA capability.

## Test Plan

- [x] Updated existing tests for new interfaces (`test_fs_backend.py`, `test_priority_queue.py`)
- [x] Added multi-group HMA roundtrip test (`test_hma.py`)
- [x] End-to-end pass with `state-spaces/mamba-130m-hf`
- [x] End-to-end validation with `Qwen/Qwen3.5-27B`


kv_connectors/llmd_fs_backend/llmd_fs_backend/spec.py
```diff
@@ -67,15 +67,12 @@ def __init__(self, vllm_config: VllmConfig, kv_cache_config: KVCacheConfig):
...
-        assert len(self.gpu_block_size) == 1, (
-            f"Expected exactly one KV cache group, got {len(self.gpu_block_size)}"
-        )
-
         hash_block_size = vllm_config.cache_config.block_size
         assert self.offloaded_block_size % hash_block_size == 0, (
             "offloaded_block_size must be a multiple of hash_block_size"
         )
         self.gpu_blocks_per_file = self.offloaded_block_size // hash_block_size
+        self._hash_block_size = hash_block_size
```
orozery: This won't work with DSv4.
You should use `OffloadingSpec.hash_block_size`.

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
```diff
@@ -367,22 +455,27 @@ def __init__(
...
         """
-        bytes_per_gpu_block = sum(
-            tensor.stride(0) * tensor.element_size() for tensor in tensors
-        )
-        file_size_in_bytes = bytes_per_gpu_block * gpu_blocks_per_file
+        max_group_bytes = 0
+        for indices in group_tensor_indices:
+            group_bytes = sum(
+                tensors[idx].stride(0) * tensors[idx].element_size() for idx in indices
+            )
```
orozery: You should get the group_bytes directly from `CanonicalKVCacheRef.page_size_bytes` instead of doing tensor strides/element_size.

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
```diff
@@ -277,9 +346,18 @@ def __init__(
...
         assert tensors
 
+        # Per-group tensor indices into the flat `tensors` list.
+        # For single-group models this is a single list covering all tensors
+        # used by that group. For HMA models each group has its own subset.
+        group_tensor_indices: list[list[int]] = [
+            [ref.tensor_idx for ref in group_refs]
+            for group_refs in kv_caches.group_data_refs
+        ]
+        assert group_tensor_indices, "CanonicalKVCaches has no groups"
```
orozery: You should pass `group_data_refs: list[list[CanonicalKVCacheRef]]` downwards instead of `group_tensor_indices: list[list[int]]`.

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
```diff
@@ -325,17 +406,24 @@ def __init__(
...
 
-        # Compute per-GPU-block size in bytes for metrics across all tensors.
-        per_block_bytes = sum(t.stride(0) * t.element_size() for t in tensors)
+        # Per-block bytes for throughput metrics. Use the largest group
+        # as an upper bound (exact for single-group, tight over-estimate
+        # for HMA with multiple groups).
+        per_block_bytes = max(
+            sum(tensors[idx].stride(0) * tensors[idx].element_size() for idx in indices)
+            for indices in group_tensor_indices
+        )
```
orozery: This will yield inaccurate metrics.
You won't to return the exact number of bytes transferred per each job.

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -165,17 +168,24 @@ int StorageOffloadEngine::get_device_id() {
...
-  size_t block_size_in_bytes = 0;
-  for (const auto& tensor : tensors) {
-    block_size_in_bytes += static_cast<size_t>(tensor.stride(0)) *
+    const std::vector<std::vector<int64_t>>& group_tensor_indices) {
+  size_t max_group_bytes = 0;
+  for (const auto& indices : group_tensor_indices) {
+    size_t group_block_bytes = 0;
+    for (int64_t idx : indices) {
+      const auto& tensor = tensors[idx];
+      group_block_bytes += static_cast<size_t>(tensor.stride(0)) *
```
orozery: same comment as `_compute_buffer_size_mb`

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
```diff
@@ -191,37 +208,92 @@ def _build_file_block_mapping(
...
+        all_files: list[str] = []
+        all_block_ids: list[list[int]] = []
+
+        block_offset = 0
+        key_offset = 0
+        for group_idx, group_size in enumerate(group_sizes):
+            if group_size == 0:
+                continue
+
+            num_files = math.ceil(group_size / self.gpu_blocks_per_file)
```
orozery: What if `group_size == self.gpu_blocks_per_file` but this group is not aligned to `self.gpu_blocks_per_file`?
You will set `num_files = 1`, instead of `num_files = 3`.
You should use `gpu_spec.block_indices` to determine the group alignment.

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
```diff
@@ -191,37 +208,92 @@ def _build_file_block_mapping(
...
+            group_keys = storage_spec.keys[key_offset : key_offset + num_files]
+
+            # Sanity check: key's encoded group_idx matches expected group.
+            if group_keys:
+                assert get_offload_group_idx(group_keys[0]) == group_idx, (
+                    f"Expected group_idx={group_idx} but key encodes "
+                    f"{get_offload_group_idx(group_keys[0])}"
+                )
+
+            group_files, group_per_file_block_ids = self._build_file_block_mapping(
```
orozery: You need to pass in this group `block_indices` to correctly map to files.

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
```diff
@@ -191,37 +208,92 @@ def _build_file_block_mapping(
...
+                keys=group_keys,
+                block_ids=group_block_ids,
+            )
+
+            all_group_indices.extend([group_idx] * len(group_files))
+            all_files.extend(group_files)
+            all_block_ids.extend(group_per_file_block_ids)
+
```
orozery, line 319: I think you want to add `all_sizes` (the size in bytes per each block).

orozery (changes requested):

kv_connectors/llmd_fs_backend/llmd_fs_backend/spec.py
kfirtoledo: Done

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
kfirtoledo: Done

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
kfirtoledo: Done

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: Done

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
kfirtoledo: Done

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
kfirtoledo: Exact bytes per (group, file) are computed at the PUT/GET call sites via len(ids) * per_group_block_bytes[g] (dbaad02) — same data, not materialized as a separate list. (so i think it's fine)

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
kfirtoledo: 1. Kept `group_data_refs` as the Python source-of-truth.
2. Derived two int views from it: `group_tensor_indices` (identity → TensorCopier) and `per_group_block_bytes` (bytes → engine sizing + handler metrics).
3. C++ pybind can't accept `CanonicalKVCacheRef` without a wrapper struct, and the two C++ consumers (TensorCopier vs. staging-buffer sizing) each need a different field — so passing two flat int arrays is simpler than threading the full ref.

kfirtoledo:
@orozery PTAL

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
```diff
@@ -158,105 +178,202 @@ def _build_file_block_mapping(
...
+        for group_idx, group_size in enumerate(group_sizes):
+            if group_size == 0:
+                continue
+
+            # Number of files this group spans, computed from the group's
+            # logical start index (block_indices[group_idx]) so unaligned
+            # heads/tails are counted correctly.
+            start_block_idx = group_block_indices[group_idx]
+            end_block_idx = start_block_idx + group_size  # exclusive
+            num_files = (end_block_idx - 1) // gpb - start_block_idx // gpb + 1
```
orozery: This is confusing to understand.
Can you compute start end indexes in (gpu_blocks_per_file) granularity, end then `num_files = end_idx - start_idx` ?

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
```diff
@@ -325,17 +462,23 @@ def __init__(
...
             gds_mode=gds_mode,
         )
 
-        # Compute per-GPU-block size in bytes for metrics across all tensors.
-        per_block_bytes = sum(t.stride(0) * t.element_size() for t in tensors)
+        # Per-block bytes for throughput metrics. Handlers select the exact
+        # value for the (group, blocks) they are transferring via
+        # per_group_block_bytes; this float is only used for the global
+        # init-time log line below.
+        per_block_bytes = max(per_group_block_bytes)
```
orozery: I don't see it used anywhere

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
```diff
@@ -325,17 +462,23 @@ def __init__(
             io_threads=threads_per_gpu,
             gpu_blocks_per_file=gpu_blocks_per_file,
             tensors=tensors,
+            group_tensor_indices=group_tensor_indices,
+            per_group_block_bytes=per_group_block_bytes,
```
orozery: What about the data-refs?
Why not just pass the canonical kv caches instead of breaking it down to multiple complex data dtypes (e.g. list[list[int]])?

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
```diff
@@ -158,105 +178,202 @@ def _build_file_block_mapping(
...
+        )
+
         files = []
         per_file_block_ids = []
+        block_offset = 0
+        for f_idx in range(num_files):
+            file_logical_lo = (start_file_idx + f_idx) * gpb
+            file_logical_hi = file_logical_lo + gpb  # exclusive
+            # Slice of this group that lives in file f_idx.
+            slice_lo = max(start_block_idx, file_logical_lo)
```
orozery, line 241: file_logical_lo is always >= than start_block_idx

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
```diff
@@ -158,105 +178,202 @@ def _build_file_block_mapping(
...
+                block_ids=group_block_ids,
+                start_block_idx=start_block_idx,
+            )
 
-            # Build file path for this group of blocks
-            files.append(self.file_mapper.get_file_name(key))
-            per_file_block_ids.append(block_ids_chunk)
+            all_group_indices.extend([group_idx] * len(group_files))
+            all_files.extend(group_files)
+            all_block_ids.extend(group_per_file_block_ids)
```
orozery, line 319: You don't propagate the file offset starting position for the transfer.

orozery (changes requested):

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
kfirtoledo: Done

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
kfirtoledo: Removed

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
kfirtoledo: Done, I wasn't aware of partial write

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
kfirtoledo: Done,

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
kfirtoledo: True for every file after the first one, but for the first file file_logical_lo = start_file_idx * gpu_blocks_per_file rounds start_block_idx down to the file boundary — so it can be < start_block_idx when the group's start is not aligned to gpu_blocks_per_file.

kfirtoledo:
@orozery PTAL

orozery (approved):

PR #485, opened 2026-03-30
feat: Add golden test case for multi-modal

gyliu513:
This is a follow up for https://github.com/llm-d/llm-d-kv-cache/pull/475#issuecomment-4154962501


yankay (approved):
Thanks @gyliu513! Solid extension of the golden test pattern to cover the multimodal pipeline. LGTM.

/lgtm

PR #492, opened 2026-04-01
fix: validate ZMQ sequence frame length before decoding

wenhug:
## Summary
- Add a length guard for the ZMQ sequence frame before calling `binary.BigEndian.Uint64()` in `zmq_subscriber.go`
- Malformed messages with truncated sequence frames (< 8 bytes) are now logged and skipped instead of panicking the process
- Add `TestZMQSubscriber_ShortSequenceFrameSkipped` to verify the fix

Fixes #491

## Test plan
- [ ] `go test -race ./pkg/kvevents/...` passes
- [ ] New test `TestZMQSubscriber_ShortSequenceFrameSkipped` sends a 3-byte sequence frame and verifies no panic

### DIFF
--- a/pkg/kvevents/zmq_subscriber.go
+++ b/pkg/kvevents/zmq_subscriber.go
@@ -127,6 +127,10 @@ func (z *zmqSubscriber) runSubscriber(ctx context.Context) {
 		seqBytes := parts[1]
 		payload := parts[2]
 
+		if len(seqBytes) < 8 {
+			debugLogger.Error(nil, "Sequence frame too short", "got", len(seqBytes), "want", 8, "topic", topic, "endpoint", z.endpoint)
+			continue
+		}
 		seq := binary.BigEndian.Uint64(seqBytes)
 
 		debugLogger.V(logging.TRACE).Info("Received message from zmq subscriber",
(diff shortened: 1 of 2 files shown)


yankay (approved):
Thanks @wenhug — clean fix! LGTM.

Nit (non-blocking): the test only asserts "no panic" — a positive signal (e.g. log/metric assertion) confirming the malformed message was received would be stronger.

/lgtm

pkg/kvevents/zmq_subscriber.go
```diff
@@ -127,6 +127,10 @@ func (z *zmqSubscriber) runSubscriber(ctx context.Context) {
 		seqBytes := parts[1]
 		payload := parts[2]
 
+		if len(seqBytes) < 8 {
+			debugLogger.Error(nil, "Sequence frame too short", "got", len(seqBytes), "want", 8, "topic", topic, "endpoint", z.endpoint)
```
gyliu513, line 131: nit: How about `logger.Error`? For production env, if they are not enabling debug log, this message will not be displayed.

wenhug: V() only affects Info() in logr, Error() always fires regardless of verbosity. So `debugLogger.Error` and `logger.Error` are equivalent here.
Kept debugLogger to match the frame-count check right above it, but happy to switch if you prefer

gyliu513: I see, thanks @wenhug . LGTM

PR #497, opened 2026-04-05
fix: test imports and missing gds_mode arg

kfirtoledo:
## Summary
- Add tests directory to `sys.path` in `conftest.py` so test modules can import each other (fixes `ModuleNotFoundError: No module named 'test_fs_backend'`)
- Add missing `gds_mode` argument to `StorageOffloadingHandlers` in priority queue tests

## Test plan
- [x] `make test` — all tests pass


yankay (approved):
Thanks @kfirtoledo! Clean fix — the missing `gds_mode` arg and the sys.path patch both make sense.

/lgtm

PR #498, opened 2026-04-06
fix: statically link libstdc++ in fs_connector wheel

### LINKED ISSUE #445: llmd fs backend: python wheel built with too-recent gcc for vllm 0.15.1 CUDA container
Environment:
- Container: ghcr.io/llm-d/llm-d-cuda:v0.5.1 (RHEL 9 base)
- vLLM: 0.15.1
- llmd_fs_connector: 0.15.1-cp312-cp312-linux_x86_64.whl
- System libstdc++: libstdc++-11.5.0-11.el9.x86_64 (GCC 11, provides up to GLIBCXX_3.4.29)

Error:
File "/tmp/llmd_packages/llmd_fs_backend/spec.py", line 29, in <module>
    from llmd_fs_backend.worker import (
  File "/tmp/llmd_packages/llmd_fs_backend/worker.py", line 18, in <module>
    import storage_offload
ImportError: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.30' not found
    (required by /tmp/llmd_packages/storage_offload.cpython-312-x86_64-linux-gnu.so)

Root cause: storage_offload.so (the CUDA extension bundled in the wheel) was compiled with GCC
12+, which requires GLIBCXX_3.4.30. The llm-d-cuda:v0.5.1 image is RHEL 9-based and ships GCC 11
(libstdc++-11.5.0-11.el9), which only provides up to GLIBCXX_3.4.29.

natoscott:
The wheel is built on nvidia/cuda:12.9.1-devel-ubuntu22.04 which uses GCC 12, producing a .so that requires GLIBCXX_3.4.30. The target runtime (llm-d-cuda:v0.6.0, RHEL 9) only provides GLIBCXX_3.4.29 (GCC 11), causing an ImportError at load time.

A previous workaround (LD_PRELOAD of Nsight Compute's libstdc++.so.6, documented in #445) is no longer viable in llm-d-cuda:v0.6.0 since Nsight Compute was removed from that image.

Fix: pass -static-libstdc++ as a linker flag so the C++ runtime is embedded directly in storage_offload.so. This eliminates the GLIBCXX version dependency at runtime entirely, with no source code changes.

Fixes #445

### DIFF
--- a/kv_connectors/llmd_fs_backend/setup.py
+++ b/kv_connectors/llmd_fs_backend/setup.py
@@ -60,6 +60,7 @@ setup(
             include_dirs=include_dirs,
             libraries=libraries,
             extra_compile_args={"cxx": cxx_args, "nvcc": nvcc_args},
+            extra_link_args=["-static-libstdc++"],
         ),
     ],
     cmdclass={"build_ext": BuildExtension},


kfirtoledo:
Looks good. Also, the increase in wheel size is minor, less than 1MB. @natoscott, can you fix the commit signature? (that the CI will pass)

natoscott:
@kfirtoledo thanks, seems to be successfully signed now.

PR #502, opened 2026-04-07
Fix: derive Indexer blockSize from injected TokenProcessor

yankay:
## Summary

- **Adds `BlockSize() int` to the `TokenProcessor` interface** so the `Indexer` can query the block size from the single source of truth — the injected `TokenProcessor` instance.
- **Changes `Indexer.blockSize()` to call `tokenProcessor.BlockSize()`** instead of reading from its own `Config.TokenProcessorConfig`, which was never propagated by external callers (e.g. the inference scheduler).
- **Removes the now-redundant `TokenProcessorConfig` field from `kvcache.Config`**, eliminating the dual-config surface that caused the mismatch.

### Problem

`TokenProcessorConfig` was added back to `kvcache.Config` at [#444](https://github.com/llm-d/llm-d-kv-cache/pull/444), but the scheduler never sets `IndexerConfig.TokenProcessorConfig` when building the indexer — it only sets the top-level `TokenProcessorConfig` used to construct the `TokenProcessor`. This left `indexer.blockSize()` falling back to the default of 16, while `tokenProcessor.chunkTokens` used the actual configured block size (e.g. 64), producing different counts for `extraFeatures` and token chunks in `GetPodScores` / `ComputeBlockKeys`.

Related issue: [#490](https://github.com/llm-d/llm-d-kv-cache/issues/490)

## Test plan

- [x] `make build` passes
- [x] `go test ./pkg/...` — all unit tests pass
- [ ] CI (`ci-pr-checks`) should pass


yankay:
Closing: branch was incorrectly based on `deprecate-embedded-tokenizers` instead of `main`. Will re-submit from the correct base.

PR #507, opened 2026-04-10
enable toolcalling from renderer

### LINKED ISSUE #506: Broken tokenization for tool calls using ```--tool-call-parser qwen3_xml```
EPP logs show the following error:
```
[pod/qwen-qwe-60b1c5b9-uct-2507-gaie-epp-6bdf64bf7c-kbs9v/epp] {"level":"error","ts":"2026-04-10T01:31:41Z","caller":"tokenization/pool.go:118","msg":"Dropping tokenization task after max retries","prompt":"","retries":3,"error":"gRPC RenderChatCompletion request failed: rpc error: code = Internal desc = Render failed: \"auto\" tool choice requires --enable-auto-tool-choice and --tool-call-parser to be set","stacktrace":"github.com/llm-d/llm-d-kv-cache/pkg/tokenization.(*Pool).workerLoop\n\t/go/pkg/mod/github.com/llm-d/llm-d-kv-cache@v0.7.1/pkg/tokenization/pool.go:118"}
```

The vLLM serve command includes ```--enable-auto-tool-choice``` and ```--tool-call-parser qwen3_xml```

Scenario file: https://gist.github.com/shashwatj07/16cdda06e4e623587dbb046fe30728c2

Gregory-Pereira:
When tools are provided without an explicit tool_choice, vLLM defaults tool_choice to "auto", which then fails validation because our renderer was created with enable_auto_tools=False. Setting enable_auto_tools=True bypasses this validation and is safe because the tokenizer service only renders prompts — it never parses tool calls from model output, so no tool parsing logic is activated.

Fixes: #506

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ PROD_VERSION ?= 0.0.0
 IMAGE_TAG_BASE ?= ghcr.io/llm-d/$(PROJECT_NAME)
 IMG = $(IMAGE_TAG_BASE):$(DEV_VERSION)
 NAMESPACE ?= hc4ai-operator
-VLLM_VERSION := 0.14.0
+VLLM_VERSION := 0.18.0
 
 TARGETOS ?= $(shell go env GOOS)
 TARGETARCH ?= $(shell go env GOARCH)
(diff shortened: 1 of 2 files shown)


vMaroon:
Note that this only bypasses the error reported in slack - as tool call parsing is not a concern of input processing, but vLLM validates it preflight AFAIK. Expecting no change across the tokenization path.

/lgtm
/approve

PR #509, opened 2026-04-10
fix: register MaxPodHitCount metric in Collectors()

### LINKED ISSUE #508: fix: MaxPodHitCount metric not registered in Collectors(), invisible to Prometheus
## Bug

`MaxPodHitCount` is defined in `pkg/kvcache/metrics/collector.go:44` and actively used in `pkg/kvcache/kvblock/instrumented_index.go:90`, but it is **not included** in the `Collectors()` slice at `collector.go:78-84`.

Since `Register()` calls `metrics.Registry.MustRegister(Collectors()...)`, `MaxPodHitCount` is never registered with the Prometheus registry. This means the metric is silently dropped — no scrape error, no panic, but the data is never exposed.

## Impact

- `kvcache_index_max_pod_hit_count_total` is invisible to Prometheus/Grafana
- Operators cannot monitor per-pod cache hit distribution
- The `recordHitMetrics` goroutine in `instrumented_index.go:71` runs on every `Lookup` call but its output is discarded

## Root Cause

`MaxPodHitCount` was likely added after the initial `Collectors()` list was written and was not appended to the slice.

Additionally, `logMetrics()` does not log `MaxPodHitCount`, so the periodic metrics beat also omits this value.

## Fix

1. Add `MaxPodHitCount` to the `Collectors()` return slice
2. Add `MaxPodHitCount` to `logMetrics()` output
3. Add test coverage for `MaxPodHitCount` in `collector_test.go`

wenhug:
## Summary

- Add `MaxPodHitCount` to the `Collectors()` return slice so it is registered with the Prometheus registry
- Add `MaxPodHitCount` to the periodic `logMetrics()` output
- Add `TestCollectorsIncludesAllMetrics` to verify all defined metrics are included in `Collectors()`
- Add `max_pod_hit_count_logged` subtest to verify `logMetrics()` outputs the new field

## Problem

`MaxPodHitCount` is defined in `collector.go:44` and actively used in `instrumented_index.go:90` via `recordHitMetrics()`, but it was not included in the `Collectors()` slice. Since `Register()` calls `metrics.Registry.MustRegister(Collectors()...)`, the metric was never registered — making `kvcache_index_max_pod_hit_count_total` invisible to Prometheus/Grafana.

## Test plan

- [x] `TestCollectorsIncludesAllMetrics` — verifies every defined metric is in `Collectors()`
- [x] `TestLogMetrics/max_pod_hit_count_logged` — verifies `logMetrics()` includes the field
- [x] All existing tests pass: `go test ./pkg/kvcache/metrics/ -v`

Fixes #508

### DIFF
--- a/pkg/kvcache/metrics/collector.go
+++ b/pkg/kvcache/metrics/collector.go
@@ -78,7 +78,7 @@ var (
 func Collectors() []prometheus.Collector {
 	return []prometheus.Collector{
 		Admissions, Evictions,
-		LookupRequests, LookupHits, LookupLatency,
+		LookupRequests, LookupHits, LookupLatency, MaxPodHitCount,
 		RenderChatTemplateLatency, TokenizationLatency, TokenizedTokensCount,
 	}
 }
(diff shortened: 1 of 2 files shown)


vMaroon:
Thanks @wenhug 

/lgtm
/approve

PR #511, opened 2026-04-11
feat: Heterogeneous BlockSize support

albertoperdomo2:
## Summary

Add heterogeneous block size support to the indexer via **canonical block size normalization**. Different vLLM engines can now use different block sizes (16, 32, 64, 256...) and the indexer normalizes all events to a single internal resolution (default: 16 tokens).

The `TokenProcessor` is the single source of truth for the canonical block size. Both the `Indexer` (read path) and `Pool` (write path) query it via `CanonicalSize()`. The `Pool` remains stateless, so all engine-to-request key mappings are owned by the `Index` implementations, which now support `many:many` relationships via `AddEngineMapping`.

This is no silver bullet and it comes with the following tradeoffs:

- **Trailing token loss**: Tokens that don't fill a complete canonical block are dropped at index time. A 300 token prompt at `canonicalBlockSize=64` indexes 256 tokens (4 blocks); the trailing 44 are not indexed. This is by design, as sub-block trailing tokens provide negligible routing signal.
- **Eviction approximation**: Eviction is eager and conservative. In the `many:1` case (`engineBlockSize=16`, `canonicalBlockSize=64`), evicting *any* of the 4 engine blocks removes the entire canonical block from the index, even if the other 3 engine blocks are still warm. This can cause false misses (pod has cache but indexer doesn't know), but false misses only cause suboptimal routing; the real issue here would be false hits since they cause recomputation on the critical path, which is worse.
- **Scoring granularity for short prompts**: If the canonical block size is large relative to prompt length, short prompts produce few indexable blocks. A 100 token prompt at canonical=64 produces 1 block, and a cold pod and a pod with 90 cached tokens look identical. However, this is tunable via the `canonicalBlockSize` config knob.

The core insight: read-path simplicity is more valuable than write-path simplicity. The read path is in the request-serving critical path; the write path is async event processing. This feature absorbs all complexity on the write side.

## Test Plan

The following test cases are included:

- **Token processor** (`token_processor_test.go`): 6 subtests covering backward compat, hash chain independence across block sizes, correct key counts, partial block handling, and parent key propagation
- **Pool write path** (`pool_test.go`): Many:1 mapping, 1:many mapping, partial block drop, cross-engine scoring (two engines with different block sizes produce identical canonical keys), and legacy fallback when canonical == engine block size
- **Pool eviction** (`pool_test.go`): Eager eviction for both many:1 and 1:many cases, unknown engine key handling
- **Indexer** (`indexer_test.go`): All existing scoring tests pass unchanged

## Test Results

All existing tests related to the modified components passing, as well as the new test cases.

### Benchmark

**TODO**: Need to rerun after the main logic is fine.

## Related Issues

- N/A

### DIFF
--- a/pkg/kvcache/kvblock/in_memory.go
+++ b/pkg/kvcache/kvblock/in_memory.go
@@ -150,23 +150,27 @@ func (m *InMemoryIndex) Lookup(ctx context.Context, requestKeys []BlockHash,
 // Add adds a set of engineKeys/requestKeys and their associated pod entries to the index backend.
 // If engineKeys is nil, only requestKey -> PodEntry mappings are created (no engineKey -> requestKey mapping).
 // This is used for speculative entries where engine keys are not yet known.
+// When engineKeys is non-nil, the mapping type is inferred from the ratio of array lengths.
 func (m *InMemoryIndex) Add(ctx context.Context, engineKeys, requestKeys []BlockHash, entries []PodEntry) error {
 	if len(requestKeys) == 0 || len(entries) == 0 {
 		return fmt.Errorf("no keys or entries provided for adding to index")
 	}
-	if engineKeys != nil && len(engineKeys) != len(requestKeys) {
-		return fmt.Errorf("mismatch between engine keys and request keys length")
-	}
 
 	traceLogger := log.FromContext(ctx).V(logging.TRACE).WithName("kvblock.InMemoryIndex.Add")
 
-	for i, requestKey := range requestKeys {
-		// 1. Store engineKey -> requestKey mapping (only if engineKeys provided)
-		if engineKeys != nil {
-			m.engineToRequestKeys.Add(engineKeys[i], []BlockHash{requestKey})
+	// Build engine->request mappings when engine keys are provided.
+	if engineKeys != nil {
+		n := max(len(engineKeys), len(requestKeys))
+		for i := 0; i < n; i++ {
+			ek := engineKeys[i*len(engineKeys)/n]
+			rk := requestKeys[i*len(requestKeys)/n]
+			existing, _ := m.engineToRequestKeys.Get(ek)
+			m.engineToRequestKeys.Add(ek, append(existing, rk))
 		}
+	}
 
-		// 2. Store requestKey -> PodCache mapping
+	// Store requestKey -> PodCache mappings for all request keys.
+	for _, requestKey := range requestKeys {
 		var podCache *PodCache
 		var found bool
 
@@ -303,11 +307,3 @@ func podsPerKeyPrintHelper(ks map[BlockHash][]PodEntry) string {
 	return b.String()
 }
 
-// AddEngineMapping stores the mapping from an engine key to one or more request keys.
-func (m *InMemoryIndex) AddEngineMapping(_ context.Context, engineKey BlockHash, requestKeys []BlockHash) error {
-	if len(requestKeys) == 0 {
-		return nil
-	}
-	m.engineToRequestKeys.Add(engineKey, requestKeys)
-	return nil
-}
--- a/pkg/kvcache/kvblock/in_memory.go
+++ b/pkg/kvcache/kvblock/in_memory.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/go-logr/logr"
 	lru "github.com/hashicorp/golang-lru/v2"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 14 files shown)


vMaroon:
Thanks @albertoperdomo2 - I think this is the right direction for general robustness.

Two comments on the changes:
1. The `TokenProcessor` can be left as is, and can remain the source of truth for the singular block-size. This includes not requiring it to be configurable through the `kvcache.Indexer`. Both the latter and the `kvevents.Pool` have a reference to it.
2. The `kvevents.Pool` should remain stateless and the mappings should be kept within the `kvblock.Index` and its implementations, extending support for a `many:many` engine-key to request-key enablement. 
    1. In the case of `engine block-size > local block-size` we get `1:many`, in which case the `kvblock.Index`'s expansion is needed.
    2. In the other case of `many:1`, since this mapping is primarily used for eviction handling, it could be feasible to only map the lowest or highest engine-key - making it a 1:1. This works since model-servers evict blocks tail to head.

Though I have not yet thought about the interaction of the above with HMA enablement, not fully sure if it is orthogonal to some of these choices.

pkg/kvcache/kvblock/in_memory.go
```diff
@@ -149,23 +150,27 @@ func (m *InMemoryIndex) Lookup(ctx context.Context, requestKeys []BlockHash,
...
-		return fmt.Errorf("mismatch between engine keys and request keys length")
-	}
 
 	traceLogger := log.FromContext(ctx).V(logging.TRACE).WithName("kvblock.InMemoryIndex.Add")
 
-	for i, requestKey := range requestKeys {
-		// 1. Store engineKey -> requestKey mapping (only if engineKeys provided)
-		if engineKeys != nil {
-			m.engineToRequestKeys.Add(engineKeys[i], requestKey)
+	// Build engine->request mappings when engine keys are provided.
```
vMaroon, line 161: nit: an example / comment on intuition can make this easier to understand.

pkg/kvcache/kvblock/in_memory.go
```diff
@@ -149,23 +150,27 @@ func (m *InMemoryIndex) Lookup(ctx context.Context, requestKeys []BlockHash,
...
+	if engineKeys != nil {
+		n := max(len(engineKeys), len(requestKeys))
+		for i := 0; i < n; i++ {
+			ek := engineKeys[i*len(engineKeys)/n]
+			rk := requestKeys[i*len(requestKeys)/n]
+			existing, _ := m.engineToRequestKeys.Get(ek)
+			m.engineToRequestKeys.Add(ek, append(existing, rk))
```
vMaroon: If different pods generate the same engine-hashes (e.g., `PYTHONHASHSEED` is the same across multiple replicas, leading to the same root-hashes and thus the same hashes for identical token sequences), this would duplicate entries. We can solve this by switching to a set instead of array. Or to calculate all and only then set instead of append.

vMaroon: The comment was originally meant for the `Add` call line, moved due to changes while writing. This applies to all implementations.

albertoperdomo2: Redis is fine since `SADD` deduplicates naturally, but I'll fix the others!

albertoperdomo2:
`SADD`/`SMEMBERS` stores engine-to-request key mappings as unordered sets. In the 1:many case (engine block size > canonical block size), `GetRequestKey` needs the *last* request key in the chain. `SMEMBERS` returns members in arbitrary order, so `rks[len(rks)-1]` is non-deterministic, Pool resolves the wrong parent hash, producing request keys that are never found on lookup.

The fix is to replace `SADD`/`SMEMBERS` with `ZADD`/`ZRANGE`, using the loop index as the score to preserve insertion order. The sorted set size per engine key is bounded by the ratio (i.e. `engineBlockSize / canonicalBlockSize`), a small constant determined by configuration, not by prompt length. However, there is an overhead vs SADD of **8 bytes per member** for the float64 score.

albertoperdomo2:
Performance report can be downloaded [here](https://github.com/user-attachments/files/26707547/het-block-report.html).

vMaroon:
I also ran the classic llm-d precise benchmark but decreased QPS stages:

| Experiment | Output toks/s | Requests/s | Success Rate | TTFT p90 (s) | TTFT mean (s) | ITL mean (s) | ITL p50/ p90 (s) |
|---|---:|---:|---:|---:|---:|---:|---:|
| het:64:128 | 11618.4 | 27.720 | 99.97% | 1.466 | 0.752 | 0.038 | 0.0350/0.045 |
| het:64:32 | 11982.0 | 28.101 | 100.00% | 2.803 | 1.553 | 0.037 | 0.0349/0.044 |
| het:64:64 | 12328.8 | 28.287 | 100.00% | 3.219 | 1.724 | 0.038 | 0.0353/0.045 |
| v0.7.1 | 10494.9 | 28.047 | 100.00% | 7.297 | 2.994 | 0.039 | 0.0360/0.046 |

latency_vs_qps
throughput_vs_qps


Where in the experiment name, it is engine-block-size:canonical-block-size. Looks great, thanks @albertoperdomo2!

PR #512, opened 2026-04-13
fix: prevent write queue deadlock under high concurrency

### LINKED ISSUE #457: llmd fs backend: EngineCore deadlock under SharedStorageOffloadingSpec + mp executor at high concurrency
I've been encountering issues running GuideLLM to measure the fs kvcache offload.  Following report with possible root cause is AI-generated with human guidance - let me know if I (said human) can assist with further testing or to provide additional information.

## Summary

When using `SharedStorageOffloadingSpec` (from `llmd_fs_connector`) with
`--distributed-executor-backend mp`, the vLLM EngineCore deadlocks under
sustained high concurrency. The API server remains alive and `/health` returns
200 OK, but the engine stops dispatching requests, causing the gateway to return
HTTP 503 for all subsequent requests.

## Environment

| Component | Version |
|-----------|---------|
| vLLM | 0.15.1 |
| llm-d | v0.5.1 (`ghcr.io/llm-d/llm-d-cuda:v0.5.1`) |
| llmd_fs_connector | 0.15.1 (`llmd_fs_connector-0.15.1-cp312-cp312-linux_x86_64.whl`) |
| Python | 3.12.12 |
| CUDA | 12.9 |
| GPUs | 2× NVIDIA L40S (tensor-parallel-size=2) |
| OS | RHEL 9 (kernel 5.14) |
| Executor backend | `--distributed-executor-backend mp` | 
| Connector | `OffloadingConnector` / `SharedStorageOffloadingSpec` |

### vLLM command

```bash
pip3.12 install --target /tmp/llmd_packages \
    llmd_fs_connector-0.15.1-cp312-cp312-linux_x86_64.whl

PYTHONPATH=/tmp/llmd_packages \
LD_PRELOAD=/opt/nvidia/nsight-compute/.../libstdc++.so.6 \
PYTHONHASHSEED=42 \
exec vllm serve Qwen/Qwen3-0.6B \
    --tensor-parallel-size 2 \
    --port 8000 \
    --max-num-seq 1024 \
    --distributed-executor-backend mp \
    --kv-transfer-config '{
      "kv_role": "kv_both",
        "shared_storage_path": "/kvcache/kv-cache/",
        "block_size": 256,
        "threads_per_gpu": 64,
        "spec_module_path": "llmd_fs_backend.spec"
      }
    }'
```

Send ≥300 concurrent requests with a large shared prefix (10,000 tokens),
5 conversation turns, 128 prompt + 128 output tokens per turn:

```bash
guidellm benchmark run \
    --target http://localhost:8000 \
    --rate-type concurrent \
    --rate 300 \
    --max-seconds 120 \
    --data '{"prompt_tokens":128,"output_tokens":128,"prefix_tokens":10000,"turns":5,"prefix_count":600}'
```

The deadlock manifests within ~8 minutes of the high-concurrency load starting.
At `rate=150` the first ~39 requests succeed before the deadlock sets in.
At `rate=300` and `rate=500` zero requests succeed.

The issue does **not** reproduce with:
- `--distributed-executor-backend ray` (not tested but likely avoids the shm path)
- native-offload-20k (`CPUOffloadingSpec`) under identical concurrency
- no-offload under identical concurrency
- Larger models (Qwen3-8B, Qwen3-14B, Qwen3-32B-AWQ) under identical concurrency
  (slower token generation rate reduces KV write pressure sufficiently)

### 1. vLLM log: shm_broadcast timeout

Appears in the vLLM log approximately 8 minutes after load begins, then repeats
every 60 seconds until the process is restarted:

```
[EngineCore_DP0 pid=288] INFO shm_broadcast.py:542 No available shared memory
broadcast block found in 60 seconds. This typically happens when some processes
are hanging or doing some time-consuming work (e.g. compilation, weight/kv
cache quantization).
```

### 2. HTTP 503 from gateway

All client requests receive `HTTP 503 Service Unavailable` once the deadlock
sets in. The `/health` endpoint continues to return 200 OK (the API server
process is alive; only the EngineCore is stalled).

### 3. Request error distribution (from GuideLLM)

At rate=300 over 120 seconds:
- Successful: 0
- Errored: 53,415 (`HTTPStatusError: 503 Service Unavailable`)
- Incomplete: 3 (cancelled at benchmark end)

Error traceback from client:
```
httpx.HTTPStatusError: Server error '503 Service Unavailable' for url
  'http://<gateway>/v1/chat/completions'
```

### 4. Connector initialisation is correct

The startup log confirms `SharedStorageOffloadingSpec` initialises successfully
on all TP workers before the deadlock occurs:

```
[Worker_TP0] INFO factory.py:51 Creating offloading spec with name: SharedStorageOffloadingSpec
[Worker_TP1] INFO factory.py:51 Creating offloading spec with name: SharedStorageOffloadingSpec
[Worker_TP0] INFO worker.py:223 StorageOffloadingHandlers: threads_per_gpu=64,
    offloading block_size=256, staging_buffer_size_mb=14, max_staging_memory_gb=150
[Worker_TP1] INFO worker.py:223 StorageOffloadingHandlers: threads_per_gpu=64,
    offloading block_size=256, staging_buffer_size_mb=14, max_staging_memory_gb=150
```

### Causal chain

```
llmd_fs_connector staging buffer exhausted under high write pressure
  → connector write call blocks the vLLM worker process's main loop
    → worker cannot drain vLLM's shm_broadcast queue  [vLLM: shm_broadcast.py]
      → EngineCore times out waiting for broadcast ack  [vLLM: shm_broadcast.py]
        → engine stalls; gateway returns HTTP 503
```

### Primary: llmd_fs_connector write path blocks the worker main loop

The connector initialises with `staging_buffer_size_mb=14` (only 14 MB of
staging memory across 64 I/O threads per GPU). At high concurrency with a fast
model, the rate of KV block writes exhausts this buffer. When the buffer is
full, the connector's write call into `StorageOffloadingHandlers` blocks the
calling thread. The calling thread is the vLLM worker process's main loop. With
the main loop blocked, the worker cannot drain the `shm_broadcast` queue.

The primary fix belongs in **llmd_fs_connector**: write operations into the
connector should never block the worker's main loop. The correct design is a
non-blocking enqueue with backpressure signalled to the vLLM scheduler via the
connector API, allowing the scheduler to reduce dispatch rate rather than
stalling the worker.

### Secondary: vLLM mp executor is not resilient to slow workers

vLLM's `shm_broadcast.py` implements a circular broadcast where the EngineCore
waits indefinitely (with 60-second logged warnings) for all workers to
acknowledge each slot. This design assumes workers drain the queue promptly.
There is no mechanism for a worker to signal that it is temporarily busy, nor
for the EngineCore to gracefully degrade (e.g. timeout a single worker rather
than stalling the whole engine). A secondary fix in **vLLM** would make the mp
executor resilient to slow broadcast acknowledgements regardless of the cause.

### Why Qwen3-0.6B is affected but larger models are not

Qwen3-0.6B generates tokens at ~636 tok/s (no-offload, rate=50). Under
rate=300 concurrent requests the volume of KV blocks written per second
substantially exceeds what larger models produce — Qwen3-8B ~114 tok/s,
Qwen3-14B ~59 tok/s, Qwen3-32B-AWQ ~51 tok/s. The smaller model's high token
throughput creates disproportionate write pressure, exhausting the 14 MB
staging buffer where larger models do not.

### Storage medium note

The storage path was backed by an IBM cloud VPC block PVC.
During the 120-second benchmark window, disk I/O was negligible (≤0.04 MB/s)
— all writes went to the OS page cache. The deadlock therefore occurs before
any physical I/O, purely from staging buffer exhaustion and the resulting
thread-blocking in the connector.

## Impact

- **Scope**: Single-node deployments using `SharedStorageOffloadingSpec` with
  `--distributed-executor-backend mp` under high request concurrency with
  fast (small) models
- **Effect**: Complete request failure (100% 503 errors) with no vLLM crash or
  restart required to trigger — the process must be restarted to recover
- **Workaround**: None identified within the current vLLM 0.15.1 +
  llmd_fs_connector 0.15.1 combination. Reducing concurrency below the
  threshold (rate < ~150 for Qwen3-0.6B on 2× L40S) avoids the deadlock.

### llmd_fs_connector (primary)

1. **Non-blocking write path**: `StorageOffloadingHandlers` write calls should
   never block the caller. Use a bounded async queue; if the queue is full,
   return immediately with a backpressure signal rather than blocking.
2. **Backpressure API**: expose a connector method that the vLLM scheduler can
   poll to determine whether the connector is under pressure, allowing the
   scheduler to reduce dispatch rate before the staging buffer fills.
3. **Larger default staging buffer**: 14 MB is small for high-throughput models;
   the default should scale with `threads_per_gpu` and expected write rate.

### vLLM mp executor (secondary)

4. **Resilient broadcast acknowledgement**: the EngineCore should not stall the
   entire engine if one worker is slow to acknowledge. Options include draining
   the broadcast in a dedicated worker thread (decoupled from the main loop), or
   treating a broadcast timeout as a worker health failure with graceful recovery.
5. **Broadcast timeout context**: the 60-second timeout is appropriate for
   one-time operations (compilation, quantization) but not for steady-state
   request handling. The timeout or escalation path could be context-aware.

kfirtoledo:
## Summary

Fixes #457 — EngineCore deadlock under high concurrency with `SharedStorageOffloadingSpec`.

Under high request concurrency with fast models (e.g., Qwen3-0.6B at rate=300), the write queue grows unbounded (observed: 1,300-2,500 tasks). When GPU KV cache hits 99.9%, `get_finished()` can't report completions fast enough → scheduler can't free blocks → throughput drops to 0 → `shm_broadcast` timeout.

**1. Dynamic write queue limit** — Caps write queue depth based on observed storage speed:
- Formula: `threads × max_write_queued_seconds / avg_write_duration`
- Fast NVMe (20ms writes): limit ~32,000 (never hit)
- Slow block storage (2,200ms writes): limit ~291 (drops excess writes)
- Dropped writes = cache miss on future read, not data loss
- Config: `max_write_queued_seconds` in `kv_connector_extra_config` (default: 10.0, 0 to disable)

**2. Cancel-on-wait for preemption** — When `wait_job()` is called during request preemption, queued-but-not-started write tasks are cancelled immediately. Only in-flight tasks are waited on. Prevents blocking the worker main loop.

## Test plan

- [x] Reproduced deadlock on L40S (rate=300, slow block storage) — **fixed**: no shm_broadcast timeout, throughput never drops to 0
- [x] Reproduced deadlock on H100 (rate=1000, fast NVMe) — **fixed**: no shm_broadcast timeout
- [x] Verified fast storage doesn't drop writes unnecessarily (dynamic limit adapts)
- [x] `make test` — all tests passed
- [x] E2E with Llama-3.1-8B

### DIFF
--- a/kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
+++ b/kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
@@ -78,6 +78,8 @@ StorageOffloadEngine::StorageOffloadEngine(int io_threads,
 // Update Exponential Moving Average (EMA) of per-file write duration.
 // EMA gives more weight to recent samples: new = old * 0.95 + sample * 0.05.
 void StorageOffloadEngine::update_write_duration(uint64_t duration_us) {
+  // Clamp to 1us so sub-microsecond writes still register a non-zero EMA.
+  if (duration_us == 0) duration_us = 1;
   uint64_t old_val = m_avg_write_duration_us.load();
   uint64_t new_val;
   do {
--- a/kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
+++ b/kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
@@ -171,16 +171,24 @@ std::vector<std::pair<int, bool>> StorageOffloadEngine::get_finished() {
   return results;
 }
 
-// Wait for all tasks in the specified job to complete
+// Wait for all tasks in the specified job to complete.
+// Signals cancellation first so queued-but-not-started tasks bail
+// immediately, preventing long blocking during request preemption.
 void StorageOffloadEngine::wait_job(int job_id) {
+  std::shared_ptr<JobState> job_state;
   std::vector<std::shared_future<bool>> futures;
 
   {
     std::lock_guard<std::mutex> lock(m_jobs_mutex);
     auto it = m_jobs.find(job_id);
     if (it == m_jobs.end()) return;
+    job_state = it->second;
     futures = it->second->futures;
   }
+
+  // Signal cancellation — queued tasks will check this flag and bail early.
+  // In-flight tasks (already past GPU copy) will skip file write.
+  job_state->cancelled = true;
 
   for (auto& fut : futures) {
     fut.wait();
@@ -226,6 +234,13 @@ bool StorageOffloadEngine::async_store_gpu_blocks(
 
     auto future = m_thread_pool.enqueue(
         [this, dst_file, block_ids, job_state, gpu_kvs_ready_event]() -> bool {
+          // Check if job was cancelled (e.g. request preempted) before
+          // starting any work — bail immediately to unblock wait_job().
+          if (job_state->cancelled) {
+            job_state->completed_tasks.fetch_add(1);
+            return true;
+          }
+
           // Check if dst_file file already exists - skip write if it does
           if (std::ifstream(dst_file).good()) {
             FileIO::update_atime(dst_file);
--- a/kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
+++ b/kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
@@ -58,7 +58,8 @@ StorageOffloadEngine::StorageOffloadEngine(int io_threads,
                                            int gpu_blocks_per_file,
                                            std::vector<torch::Tensor>& tensors,
                                            int read_preferring_workers,
-                                           const std::string& gds_mode_str)
+                                           const std::string& gds_mode_str,
+                                           float max_write_queued_seconds)
     : m_tensor_copier(tensors, gpu_blocks_per_file),
       m_gds_mode(parse_gds_mode(gds_mode_str)),
       m_thread_pool(
@@ -66,8 +67,38 @@ StorageOffloadEngine::StorageOffloadEngine(int io_threads,
           calc_staging_bytes(gpu_blocks_per_file, tensors, m_gds_mode),
           get_device_id(),
           read_preferring_workers),
-      m_gpu_blocks_per_file(gpu_blocks_per_file) {
+      m_gpu_blocks_per_file(gpu_blocks_per_file),
+      m_max_write_queued_seconds(max_write_queued_seconds) {
   init_handlers(m_gds_mode, tensors);
+  FS_LOG_INFO("Dynamic write queue limit: max_write_queued_seconds="
+              << m_max_write_queued_seconds
+              << (m_max_write_queued_seconds <= 0 ? " (disabled)" : ""));
+}
+
+// Update Exponential Moving Average (EMA) of per-file write duration.
+// EMA gives more weight to recent samples: new = old * 0.95 + sample * 0.05.
+void StorageOffloadEngine::update_write_duration(uint64_t duration_us) {
+  uint64_t old_val = m_avg_write_duration_us.load();
+  uint64_t new_val;
+  do {
+    if (old_val == 0) {
+      new_val = duration_us;
+    } else {
+      new_val = static_cast<uint64_t>(old_val * 0.95 + duration_us * 0.05);
+    }
+    // Atomic try-update: retry if another thread modified the value first
+  } while (!m_avg_write_duration_us.compare_exchange_weak(old_val, new_val));
+}
+
+// Compute dynamic write queue limit based on avg write duration
+size_t StorageOffloadEngine::get_dynamic_write_queue_limit() const {
+  uint64_t avg_us = m_avg_write_duration_us.load();
+  if (avg_us == 0 || m_max_write_queued_seconds <= 0) {
+    return 0;  // no limit yet (no data or disabled)
+  }
+  double avg_sec = avg_us / 1e6;
+  return static_cast<size_t>(m_thread_pool.num_threads() *
+                             m_max_write_queued_seconds / avg_sec);
 }
 
 // Initialize read/write handlers based on GDS mode.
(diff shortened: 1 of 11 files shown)


kfirtoledo:
@natoscott, can you test if this fix also helps in your environment?

kfirtoledo:
@dannyharnik @effi-ofer PTAL

effi-ofer:
@kfirtoledo would it be possible to add a test for this? I would like to run this with the object backend.

kfirtoledo:
@dannyharnik @effi-ofer PTAL

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
```diff
@@ -58,16 +58,49 @@ StorageOffloadEngine::StorageOffloadEngine(int io_threads,
...
+void StorageOffloadEngine::update_write_duration(uint64_t duration_us) {
+  // Clamp to 1us so sub-microsecond writes still register a non-zero EMA.
+  if (duration_us == 0) duration_us = 1;
+  uint64_t old_val = m_avg_write_duration_us.load();
+  uint64_t new_val;
+  do {
+    if (old_val == 0) {
+      new_val = duration_us;
+    } else {
+      new_val = static_cast<uint64_t>(old_val * 0.95 + duration_us * 0.05);
```
dannyharnik: can make 0.95 a parameter

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
```diff
@@ -252,6 +253,7 @@ def __init__(
         gds_mode: str,
         max_staging_memory_gb: int = DEFAULT_MAX_STAGING_MEMORY_GB,
         read_preferring_ratio: float = DEFAULT_READ_PREFERRING_WORKERS_RATIO,
+        max_write_queued_seconds: float = DEFAULT_MAX_WRITE_QUEUED_SECONDS,
```
effi-ofer, line 256: minor, but why a float?

effi-ofer:
Two main comments: 
1. some of the numbers (0.95, 0.5) are kind of magic.  I am ok with that if we document why those numbers were chosen.  If you think they might need to be adjusted then maybe add parameters?  
2. please open an issue for me to implement the same for object storage.

kv_connectors/llmd_fs_backend/llmd_fs_backend/worker.py
kfirtoledo: Float lets users tune sub-second budgets when needed (e.g. for testing or aggressive throttling), and the cost is essentially zero — the math is already floating-point under the hood (`avg_us / 1e6` in C++). Keeping it as float for flexibility.

kv_connectors/llmd_fs_backend/csrc/storage/storage_offload.cpp
kfirtoledo: Renamed `0.95` to a named `EMA_ALPHA = 0.05` constant with a short comment. Keeping it as an internal constant rather than a runtime parameter — adding it would mean another knob to plumb through, document, and test for marginal benefit. Storage write speeds are typically stable enough that alpha=0.05 (~20-sample window) works well across cases. We already expose the meaningful knob (`max_write_queued_seconds`); alpha is an implementation detail. Easy to promote later if a real workload shows it should be tunable.

kfirtoledo:
@effi-ofer , @dannyharnik PTAL

dannyharnik (approved):
\LGTM
\Approve

natoscott:
> @natoscott, can you test if this fix also helps in your environment?

Sorry for not getting back to you on this earlier, I'm currently traveling - I'll do further testing when I can (wont be for another couple of weeks though).

PR #518, opened 2026-04-14
fix: use Tokenize RPC instead of RenderCompletion in UdsTokenizer.Render

acardace:
The RenderCompletion gRPC method is not implemented by the UDS tokenizer server, causing 'Unimplemented' errors for plain completion requests. Delegate Render() to Encode() which calls the Tokenize RPC that the server supports. This fixes both direct Render() callers and the tokenization pool's processTask path.

### DIFF
--- a/pkg/tokenization/uds_tokenizer.go
+++ b/pkg/tokenization/uds_tokenizer.go
@@ -216,24 +216,9 @@ func (u *UdsTokenizer) warmup(ctx context.Context) {
 
 func strPtr(s string) *string { return &s }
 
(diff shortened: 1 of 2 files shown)


sagearc (commented):
Hi @acardace !
This PR reverts to transformers based tokenization. The vllm renderer service is backwards incompatible with previous llm-d versions. Can you try resolve the dependencies and build flags and let us know if it helps?

acardace:
@sagearc I've just re-tested with latest main and it works fine, so just ignore this, thank you!

PR #522, opened 2026-04-16
fix(fs_backend): suffix offload metrics via vllm.general_plugins

### LINKED ISSUE #515: Multiconnector with Offloading connector for CPU offloading and Storage offloading fails due to duplicate metrics registration
I want to enable both CPU offloading and Storage (Lustre backed) offloading for LLM-D FS connector. For this I am using Multi connector as recommended in https://github.com/llm-d/llm-d/issues/682
I am using llm-d image  ghcr.io/llm-d/llm-d-cuda:v0.6.0 with below VLLM config
```
exec vllm serve \
            meta-llama/Llama-3.3-70B-Instruct \
            --tensor-parallel-size 4 \
            --port 8000 \
            --max-num-seq 1024 \
            --enable_prefix_caching \
            --gpu-memory-utilization 0.65 \
            --kv-transfer-config '{
              "kv_connector":"MultiConnector",
              "kv_role":"kv_both",
                "connectors": [
                  {
                    "kv_role": "kv_both",
                      "cpu_bytes_to_use": 64424509440
                    }
                  },
                  {
                    "kv_role": "kv_both",
                    "kv_connector_extra_config": 
                    {
                      "shared_storage_path": "/mnt/files-storage/",
                      "block_size": 256,
                      "threads_per_gpu": 64,
                      "spec_module_path": "llmd_fs_backend.spec"
                    }
                  }
                ]
              }
            }'

```

As it uses offloading connector twice for CPU and storage offloading my pod keeps crashing with metrics registration error
```
(APIServer pid=1) ValueError: Duplicated timeseries in CollectorRegistry: {'vllm:kv_offload_total_bytes_created', 'vllm:kv_offload_total_bytes', 'vllm:kv_offload_total_bytes_total'}
```

I see both connectors initialized 

Image

kfirtoledo:
## Summary

When vLLM is deployed with `MultiConnector` wrapping two `OffloadingConnector` children (for example a CPU cache + the FS backend), startup crashes with `ValueError: Duplicated timeseries in CollectorRegistry`. Both children try to register the same Prometheus metric names (`vllm:kv_offload_total_bytes`, `..._time`, `..._size`), and the second one aborts the engine.

This PR adds a small plugin that patches `OffloadPromMetrics` so that when a duplicate name is detected, it retries with a suffix derived from `spec_name` (for example `_cpu`, `_sharedstorage`). The result is two cleanly separated metric families — one per offload backend — instead of a startup crash.

The patch is loaded through the `vllm.general_plugins` entry point so it runs during `EngineCore.__init__()`, before any KV connector builds its metrics. Single-connector deployments are untouched: the original metric names are kept, and the suffix only kicks in on an actual collision. A TODO is left to remove the workaround once vLLM upstream disambiguates these names natively.

## Test plan

- [x] Deployed `MultiConnector(CPU + FS)` in k8s with the rebuilt wheel — pod reaches `Application startup complete` without `Duplicated timeseries`.
- [x] `/metrics` exposes both series:
  - `vllm:kv_offload_total_bytes_total` (CPU — first child, keeps upstream name)
  - `vllm:kv_offload_total_bytes_sharedstorage_total` (FS — second child, takes the `_sharedstorage` suffix)
- [x] Prometheus scrape target reports `up=1`; Grafana datasource reads both series.

### DIFF
--- a/kv_connectors/llmd_fs_backend/llmd_fs_backend/__init__.py
+++ b/kv_connectors/llmd_fs_backend/llmd_fs_backend/__init__.py
@@ -17,6 +17,15 @@ import os
 
 from vllm.logger import init_logger
 
+from llmd_fs_backend.metrics import install_offload_metric_suffix_patch
+
+# Apply the OffloadPromMetrics spec-suffix patch before any OffloadingConnector
+# is instantiated. Required so MultiConnector can nest two OffloadingConnector
+# backends without hitting "Duplicated timeseries in CollectorRegistry".
+# TODO: remove once vLLM upstream applies an equivalent spec_name suffix
+# to the OffloadPromMetrics names.
+install_offload_metric_suffix_patch()
+
 _LEVEL_MAP = {
     "TRACE": logging.DEBUG,
     "DEBUG": logging.DEBUG,
--- a/kv_connectors/llmd_fs_backend/llmd_fs_backend/metrics.py
+++ b/kv_connectors/llmd_fs_backend/llmd_fs_backend/metrics.py
@@ -1,0 +1,96 @@
+# Copyright 2026 The llm-d Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import logging
+
+log = logging.getLogger(__name__)
+
+PATCHED_FLAG = "llmd_fs_patched"
+
+
+def install_offload_metric_suffix_patch() -> None:
+    """
+    Patch vLLM OffloadPromMetrics so each OffloadingConnector instance
+    registers Prometheus metrics under a spec_name-derived suffix
+    (vllm:kv_offload_total_bytes_sharedstorage, ..._cpu, ...).
+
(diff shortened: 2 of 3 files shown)


kfirtoledo:
@effi-ofer, @dannyharnik , @omerpaz95, @Sneha-at PTAL

dannyharnik:
overall \lgtm 
One question: 
In the description above you have: 
- vllm:kv_offload_total_bytes_total (CPU — first child, keeps upstream name)
- vllm:kv_offload_total_bytes_sharedstorage_total (FS — second child, takes the _sharedstorage suffix)

Shouldn't the second line be: 
- vllm:kv_offload_total_bytes_total _sharedstorage 
?

kfirtoledo:
Good question @dannyharnik! The naming is actually correct as-is. Here's how it works:

The Prometheus `Counter()` constructor receives the **base** name without `_total` — e.g. `name="vllm:kv_offload_total_bytes"`. The `_total` suffix is auto-appended by the Prometheus client library.

Our patch intercepts at the `Counter()` call, so it appends the suffix to the **base** name:
1. First child (CPU) → `Counter(name="vllm:kv_offload_total_bytes")` → succeeds → exposed as `vllm:kv_offload_total_bytes_total`
2. Second child (FS) → same call → catches `Duplicated timeseries` → retries with `Counter(name="vllm:kv_offload_total_bytes_sharedstorage")` → exposed as `vllm:kv_offload_total_bytes_sharedstorage_total`

If we put the suffix **after** `_total` (i.e. `..._total_sharedstorage`) it would violate the [Prometheus naming convention](https://prometheus.io/docs/practices/naming/#metric-names) that counters must end with `_total`.

Confirmed from the deployed pod's `/metrics` endpoint:
```
vllm:kv_offload_total_bytes_total                    ← CPU (upstream name, unchanged)
vllm:kv_offload_total_bytes_sharedstorage_total      ← FS (suffixed, _total still last)
```

dannyharnik (approved):
Got it. Be sure to remove this once solved at the vLLM level. 

\LGTM
\approve

PR #527, opened 2026-04-16
feat(fs_backend): add performance and stress tests

kfirtoledo:
## Summary

Adds a `tests/performance/` suite for benchmarking the fs_connector and
comparing against the other KV-offload backends (base / gpu / cpu / fs /
multi-cpu-fs). Two complementary entry points share a small `utils.py`.

### `test_throughput.py` — sustained cold → hot → steady-state
Runs N sequential requests of a fixed prompt through the chosen backend
and reports:
- **Cold (req 1):** full prefill + backend populate
- **Hot avg (req 2..N):** reload from offload backend
- **Last-10 avg:** steady-state latency
- **Throughput (KV GB/s):** `(num_tokens × KV_bytes_per_token) / last_10_avg`

### `test_stress.py` — batched hot/cold ratio sweep
Each `llm.generate()` call submits `batch_size` prompts mixed by
`hot_ratio`:
- `ratio = 0.0` → all cold (pure write stress)
- `ratio = 1.0` → all hot (pure read stress)
- in between → mix exercising the scheduler's `read_preferring_ratio`

Default sweep: `hot_ratios = [0.0, 0.1, ..., 1.0]` × 5 iters each. Set
`--num-repeats=N>=2` to repeat the full sweep against a growing FS pool
to surface drift / fragmentation.

Reports per `(repeat, ratio)`: mean / p50 / p99 latency, tokens/s, KV GB/s.

## Configuration

Both tests accept the same knobs (CLI flags + pytest `conftest.py` options):

| Flag | Default | Notes |
|------|---------|-------|
| `--backend` | `fs` | `base` / `gpu` / `cpu` / `fs` / `multi-cpu-fs` |
| `--model` | `Qwen/Qwen2.5-7B-Instruct` | HF model id |
| `--num-tokens` | 31000 / 10000 | per-request token count |
| `--num-requests` / `--num-iterations` | 40 / 5 | sequential (throughput) / per-ratio (stress) |
| `--tp-size` | 1 | tensor parallel |
| `--gpu-block-size` / `--fs-block-size` / `--cpu-block-size` | 16 / 256 / 48 | |
| `--threads-per-gpu` | 24 | fs I/O workers |
| `--log-level` | `info` | `STORAGE_LOG_LEVEL` |
| `--storage-root` / `\$FS_CONNECTOR_STORAGE_ROOT` | `/tmp` | pytest option |
| `--hot-ratios` / `--num-repeats` | sweep / 1 | stress-only |

## Makefile

- `make test` — unit tests (excludes `./tests/performance`)
- `make test-performance` — new: runs the throughput + stress pytest suite

## Test plan

- [x] `make test-performance` passes on an 80GB H100 node (3m19s)
  - throughput across Qwen3-0.6B, Llama-3.2-1B, Qwen2.5-7B @ 31K tokens,
    Llama-3.1-8B @ 128K tokens — all assertions pass (hot < cold, steady
    degradation < 20%, p99 < 10s, tail < 5s)
  - stress sweep on Qwen2.5-7B (batch=32, 3 ratios, 2 iters)
- [x] Sanity: all 5 backends (`base`, `gpu`, `cpu`, `fs`, `multi-cpu-fs`)
  work for both tests on Qwen3-0.6B — no deadlocks, no duplicated
  Prometheus registrations, clean `[CLEANUP]` output

### DIFF
--- a/kv_connectors/llmd_fs_backend/tests/performance/__init__.py
+++ b/kv_connectors/llmd_fs_backend/tests/performance/__init__.py
@@ -1,0 +1,1 @@
+# Performance and stress tests for fs_connector
--- a/kv_connectors/llmd_fs_backend/tests/performance/conftest.py
+++ b/kv_connectors/llmd_fs_backend/tests/performance/conftest.py
@@ -31,7 +31,7 @@ def pytest_addoption(parser):
             "Root directory for fs_connector storage. Any local Linux path "
             "(e.g., /tmp, /mnt/pvc/kv-cache, /dev/shm). "
             "A unique subdirectory is created under this root per test. "
-            "Default: $FS_CONNECTOR_STORAGE_ROOT env var, or system /tmp."
+            "Default: system /tmp."
         ),
     )
     parser.addoption(
@@ -66,51 +66,11 @@ def _apply_storage_log_level(request):
 
 
 @pytest.fixture
-def storage_root(request):
-    """
-    Resolve the storage root directory for fs_connector tests.
-
-    Precedence:
-      1. --storage-root CLI flag (if set)
-      2. FS_CONNECTOR_STORAGE_ROOT env var (if set)
-      3. System /tmp (default)
-    """
-    root = request.config.getoption("--storage-root")
-    if root is None:
-        root = os.environ.get("FS_CONNECTOR_STORAGE_ROOT")
-    if root is None:
-        root = tempfile.gettempdir()  # typically /tmp
+def storage_path(request):
+    """Unique temp storage dir per test; cleaned up after."""
+    root = request.config.getoption("--storage-root") or tempfile.gettempdir()
     os.makedirs(root, exist_ok=True)
-    return root
-
-
-@pytest.fixture
-def storage_path(storage_root):
-    """
-    Provide a unique temporary storage subdirectory for each test.
-
-    Created under `storage_root` and cleaned up after the test.
-    """
-    tmpdir = tempfile.mkdtemp(prefix="fs_connector_perf_", dir=storage_root)
+    tmpdir = tempfile.mkdtemp(prefix="fs_connector_perf_", dir=root)
     yield tmpdir
     if os.path.exists(tmpdir):
         shutil.rmtree(tmpdir, ignore_errors=True)
-
-
-@pytest.fixture
-def fs_connector_config(storage_path):
-    """Base fs_connector configuration for tests."""
-    return {
-        "kv_connector": "OffloadingConnector",
-        "kv_role": "kv_both",
-        "kv_connector_extra_config": {
-            "spec_name": "SharedStorageOffloadingSpec",
-            "spec_module_path": "llmd_fs_backend.spec",
-            "shared_storage_path": storage_path,
-            "threads_per_gpu": 64,
-            "block_size": 256,
-            "max_staging_memory_gb": 150,
-            "gds_mode": "disabled",
-            "read_preferring_ratio": 0.75,
-        },
-    }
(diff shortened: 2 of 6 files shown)


kfirtoledo:
@Etelis, @effi-ofer, @rshavitt PTAL

kv_connectors/llmd_fs_backend/tests/performance/conftest.py
```diff
@@ -0,0 +1,116 @@
...
+
+    Created under `storage_root` and cleaned up after the test.
+    """
+    tmpdir = tempfile.mkdtemp(prefix="fs_connector_perf_", dir=storage_root)
+    yield tmpdir
+    if os.path.exists(tmpdir):
+        shutil.rmtree(tmpdir, ignore_errors=True)
+
+
+@pytest.fixture
```
Etelis: Is that ever used?

kv_connectors/llmd_fs_backend/tests/performance/__init__.py
```diff
@@ -0,0 +1 @@
+# Performance and stress tests for fs_connector
```
Etelis, line 1: `--num-iterations`

I don't see it anywhere mentioned?

Etelis (commented):
Overall LGTM
But I'll continue and re run.

kv_connectors/llmd_fs_backend/tests/performance/__init__.py
kfirtoledo: Removed

kv_connectors/llmd_fs_backend/tests/performance/conftest.py
kfirtoledo: Good catch, removed

effi-ofer:
I like it @kfirtoledo.  We need something like this.  A few comments:

1. I would rather we move away from just gpu/cpu/file and think of the three tiers as gpu/cpu/storage.  That would better accommodate new secondary tiers like object store, key value stores, etc.  
2. Regarding the backend, why base / gpu / cpu / fs / multi-cpu-fs?  What does it mean to have a gpu backend?  
3. how do you support gds? choosing fs for the backend and then setting the appropriate flags in the extra?
4. keep in mind that I created a backend parameter in the object store PR which is very similar to your backend but not identical. There will be some merging fun.
5. Are you planning to remove the performance related tests from kv_connectors/llmd_fs_backend/tests?

kfirtoledo:
Thanks @effi — here's the answer to the comments:

**1.** Done. Renamed tier `fs` → `storage`, `multi-cpu-fs` → `multi-cpu-storage`. Knob `--fs-block-size` → `--storage-block-size`. Future variants slot in cleanly.

**2.** I prefer to keep everything on the single `--backend` axis, including `gpu`. I could run it  `base` + `--enable-prefix-cache`, but one value per run is simpler — a single CLI swap switches baselines, no flag combos to remember.

**3.** Added `--storage-type {fs,gds}` on both tests. Default `fs` (CPU staging → `gds_mode="disabled"`); `gds` → `gds_mode="read_write"`. Only meaningful when `--backend=storage` / `multi-cpu-storage`.

**4.** Once the object store PR lands, we'll add `obj` to `--storage-type` alongside `fs` / `gds`.

**5.** Those (`test_priority_queue.py`, `test_gds_backend.py`) are evaluation tests, not unit tests / CI — different category from the new `performance/` sweeps. Leaving them where they are.

kfirtoledo:
@effi-ofer, @rshavitt PTAL

rshavitt:
Thanks for adding the --with-gpu-prefix-cache flag. LGTM

PR #533, opened 2026-04-19
 Add Hybrid Multi-head Attention (HMA) support for KV-Cache scoring

kapiljain1989:
# Model Config & Group ID Management

## What It Does

Tracks which attention groups are cached per KV-block per pod, enabling group-aware scoring for HMA (Hybrid Multi-head Attention) models like DeepSeek-V3.

## Architecture

There is no registry struct. Model configuration is resolved at startup into two plain maps via pure functions:

- `BuildHMAModels(configs) -> map[string]bool` — which models use HMA (used by Pool for bitmask decisions)
- `BuildAttentionInfo(configs) -> map[string]*ModelAttentionInfo` — precomputed attention metadata (used by HybridPrefixCacheScorer)

The scorer is selected once at construction based on `ModelConfigs`. Nothing is dynamic on the hot path.

```
Startup:
  ModelConfigs ([]*ModelConfig)
    |
    +--> BuildHMAModels()      --> map[string]bool           --> Pool (resolved internally from ModelConfigs)
    |
    +--> BuildAttentionInfo()  --> map[string]*ModelAttentionInfo
    |                                 |
    +--> NewKVBlockScorer()           |
           |                          |
           +-- any HMA model? ------> HybridPrefixCacheScorer
           |                            (carries AttentionInfo + DefaultScorer)
           |
           +-- all simple? ----------> LongestPrefixScorer

Hot path (no config lookups):
  Scorer.Score(keys, keyToPods, model) --> pod scores
```

## Scorer Selection

Scorer type is determined once at construction by `NewKVBlockScorer`:

1. If `ScoringStrategy` is explicitly set to `LongestPrefixMatch` in config, use `LongestPrefixScorer` (regardless of model configs).
2. If no explicit strategy and no HMA models in `ModelConfigs`, use `LongestPrefixScorer`.
3. Otherwise, use `HybridPrefixCacheScorer` with a pre-built `DefaultScorer` (LongestPrefixScorer) for non-HMA model fallback.

```go
func NewKVBlockScorer(config *KVBlockScorerConfig) (KVBlockScorer, error) {
    weightMap := buildWeightMap(config.BackendConfigs)
    defaultScorer := &LongestPrefixScorer{MediumWeights: weightMap}

    if config.ScoringStrategy == LongestPrefixMatch {
        return defaultScorer, nil
    }
    if config.ScoringStrategy == "" && len(BuildHMAModels(config.ModelConfigs)) == 0 {
        return defaultScorer, nil
    }
    return &HybridPrefixCacheScorer{
        MediumWeights: weightMap,
        AttentionInfo: BuildAttentionInfo(config.ModelConfigs),
        DefaultScorer: defaultScorer,
    }, nil
}
```

### Per-request scorer path logging

When `HybridPrefixCacheScorer` is active, it logs at V(1) which path each request takes:
- `"using HybridPrefix scorer"` — model found in AttentionInfo
- `"using LongestPrefix scorer"` — model not in AttentionInfo, falls back to DefaultScorer

### Via Indexer Config (recommended)

```go
config := &kvcache.Config{
    KVBlockIndexConfig: kvblock.DefaultIndexConfig(),
    ModelConfigs: []*kvcache.ModelConfig{
        {
            Name:  "DeepSeek-V3",
            IsHMA: true,
            AttentionGroups: []kvcache.AttentionGroupConfig{
                {GroupID: 0, AttentionType: kvcache.AttentionTypeFull, BlockSize: 64},
                {GroupID: 1, AttentionType: kvcache.AttentionTypeSlidingWindow,
                    BlockSize: 64, SlidingWindowSize: 4096},
            },
        },
        {
            Name:  "Qwen/Qwen3-8B",
            IsHMA: false,
        },
    },
}
// Indexer sees HMA model --> auto-selects HybridPrefixMatch scorer
indexer, _ := kvcache.NewKVCacheIndexer(ctx, config, tokenProcessor)

// Pool resolves HMA models from its own ModelConfigs at construction
poolCfg.ModelConfigs = config.ModelConfigs
pool := kvevents.NewPool(poolCfg, indexer.KVBlockIndex(), tokenProcessor, adapter)
```

### Scorer config with explicit strategy override

```go
scorer, _ := kvcache.NewKVBlockScorer(&kvcache.KVBlockScorerConfig{
    ScoringStrategy: kvcache.LongestPrefixMatch, // overrides auto-detection
    BackendConfigs:  kvcache.DefaultKVCacheBackendConfig(),
    ModelConfigs:    modelConfigs,
})
```

### Default model configs

`DefaultModelConfigs()` ships a built-in HMA model:

```go
func DefaultModelConfigs() []*ModelConfig {
    return []*ModelConfig{
        {
            Name:  "openai/gpt-oss-20b",
            IsHMA: true,
            AttentionGroups: []AttentionGroupConfig{
                {GroupID: 0, AttentionType: AttentionTypeSlidingWindow,
                    BlockSize: 64, SlidingWindowSize: 128},
                {GroupID: 1, AttentionType: AttentionTypeFull, BlockSize: 64},
            },
        },
    }
}
```

Unknown models default to non-HMA (no tracking overhead).

### Problem

HMA models like DeepSeek-V3 have multiple attention groups (e.g., full attention + sliding window). A pod is only useful if it has cached enough blocks for *all* groups. The `LongestPrefixMatch` scorer doesn't understand group-level cache state.

### Algorithm — Single-pass boundary evaluation per pod

1. **Candidate filtering:** Only pods that have the full-attention group cached at block 0 are candidates. Full attention requires all prior tokens — if block 0 is missing, nothing is reusable.

2. **Left-to-right walk:** For each candidate pod, walk blocks `b=0, 1, 2, ...` in order:
   - **Full attention (kill switch):** If the pod is missing the full-attention group at block `b`, stop — `lastSeqFull = b-1`. Full attention is strictly prefix-based, so a gap means everything after is unusable.
   - **SWA groups (contiguous counter):** For each SWA group, track a running count of consecutive hits. On a hit, increment; on a miss, reset to 0. When the count reaches `threshold` (= `ceil((windowSize-1) / blockSize)`), record `swaLastSeq[g] = b` as a "sticky checkpoint" — it only moves forward, never backward.

3. **Final score:** `min(lastSeqFull, swaLastSeq[0], swaLastSeq[1], ...) + 1`
   - This is the number of blocks from the start where *all* groups have sufficient cache coverage.
   - If any SWA group never reaches its threshold, the pod is not scored.

### Walkthrough example

Consider DeepSeek-V3 with `full_attention (group 0)` + `sliding_window (group 1, window=128, blockSize=64)`.
SWA threshold = `ceil((128-1)/64) = 2` blocks.

Requested blocks: `[B0, B1, B2, B3, B4]`

| Block | Pod-A full (g0) | Pod-A SWA (g1) | SWA count | SWA checkpoint |
|-------|-----------------|----------------|-----------|----------------|
| B0    | hit             | hit            | 1         | -1             |
| B1    | hit             | hit            | 2 (>=threshold) | **1**    |
| B2    | hit             | miss           | 0 (reset) | 1 (sticky)     |
| B3    | hit             | hit            | 1         | 1              |
| B4    | hit             | hit            | 2 (>=threshold) | **4**    |

Result: `lastSeqFull=4`, `swaLastSeq=4` -> `score = min(4, 4) + 1 = 5`

Now if Pod-B has same data but full attention misses at B3:

| Block | Pod-B full (g0) | Pod-B SWA (g1) | SWA count | SWA checkpoint |
|-------|-----------------|----------------|-----------|----------------|
| B0    | hit             | hit            | 1         | -1             |
| B1    | hit             | hit            | 2         | **1**          |
| B2    | hit             | miss           | 0         | 1              |
| B3    | **miss -> STOP**|                |           |                |

Result: `lastSeqFull=2`, `swaLastSeq=1` -> `score = min(2, 1) + 1 = 2`

### Key properties

- **Full attention is the kill switch** — a single miss terminates the scan, since prefix continuity is broken
- **SWA checkpoints are sticky** — once the threshold is met, the checkpoint only advances, never regresses. Gaps reset the counter but don't erase prior checkpoints
- **The `min()` ensures correctness** — the score reflects the furthest block where *every* group has coverage, not just one
- **Non-HMA models fall back** to `LongestPrefixScorer` automatically (no attention info -> nil -> DefaultScorer path)

## StoredGroups Bitmask

`PodEntry.StoredGroups` is a `uint32` bitmask tracking which attention groups are cached.

| Model Type | StoredGroups | Behavior |
|------------|-------------|----------|
| Simple | `0` | No tracking, zero overhead |
| HMA | `0b11` | Bit 0 = full attn, Bit 1 = SWA |

### Event Flow (write path)

```
vLLM emits BlockStored(GroupIdx: 1)
  --> Pool: hmaModels[model] == true  (resolved from ModelConfigs at construction)
  --> storedGroups = 1 << 1 = 0b10
  --> Index.Add(): existing 0b01 |= 0b10 = 0b11  (merged)
```

### Eviction (partial group removal)

```
vLLM emits BlockRemoved(GroupIdx: 1)
  --> Pool: storedGroups = 1 << 1 = 0b10
  --> Index.Evict():
        existing 0b11 &^ 0b10 = 0b01  (group 0 still cached, keep entry)

vLLM emits BlockRemoved(GroupIdx: 0)
  --> existing 0b01 &^ 0b01 = 0b00  (no groups left, delete entry)
```

Non-HMA shortcut: `StoredGroups == 0` deletes the entry immediately.

## Index Backend Implementation

All three backends (`InMemoryIndex`, `CostAwareMemoryIndex`, `RedisIndex`) key pod caches by `"podID@tier"` string with `*PodEntry` values, enabling in-place bitmask mutation.

| Backend | Add (merge) | Evict (clear) |
|---------|------------|---------------|
| InMemory | `existing.StoredGroups \|= entry.StoredGroups` | `existing.StoredGroups &^= entry.StoredGroups` |
| CostAware | Same (via `sync.Map` + pointer) | Same + `LoadAndDelete` when zero |
| Redis | Lua `bit.bor` (atomic) | Lua `bit.band` clear (atomic) |


sagearc (changes requested):
Thanks for pushing this forward. The one-index direction and group-aware scoring work here are useful pieces.

Based on the discussion from https://github.com/vllm-project/vllm/pull/40984 and the goal of keeping the first group-presence change small, I think we should avoid changing the indexing pattern more than needed. The bitmask approach may be a good long-term shape, especially for scoring, but it also changes merge/delete behavior in each backend.

For the group-presence layer, we can preserve correctness with less backend churn by keeping the group id as part of the indexed entry identity and letting add/remove stay exact. Explicit grouped entries may also keep the code easier to follow.

Once that layer is in place, the hybrid scoring work should have a cleaner base to build on.

kapiljain1989:
@sagearc , Got it 
Let me do needful

PR #535, opened 2026-04-23
fix(e2e): fix container image build platform for macOS/non-Linux hosts

### LINKED ISSUE #534: make e2e-test failed on macos
```console
(hermes-agent) gyliu-cary@Mac llm-d-kv-cache % make e2e-test
Checking if ZMQ is already installed...
bash: line 1: pkg-config: command not found
Installing ZMQ dependencies...
==> Auto-updating Homebrew...
Adjust how often this is run with `$HOMEBREW_AUTO_UPDATE_SECS` or disable with
`$HOMEBREW_NO_AUTO_UPDATE=1`. Hide these hints with `$HOMEBREW_NO_ENV_HINTS=1` (see `man brew`).
==> Auto-updated Homebrew!
Updated 2 taps (homebrew/core and homebrew/cask).
==> New Formulae
try: Quickly manage and navigate project directories for experiments
vite-plus: Unified toolchain and entry point for web development
==> New Casks
hop: View and edit HWP documents

You have 13 outdated formulae installed.

==> Fetching downloads for: zeromq and pkgconf
✔︎ Bottle Manifest pkgconf (2.5.1)                                                        Downloaded   12.2KB/ 12.2KB
✔︎ Bottle Manifest zeromq (4.3.5_2)                                                       Downloaded   11.0KB/ 11.0KB
✔︎ Bottle Manifest libsodium (1.0.22)                                                     Downloaded    7.3KB/  7.3KB
✔︎ Bottle libsodium (1.0.22)                                                              Downloaded  408.8KB/408.8KB
✔︎ Bottle pkgconf (2.5.1)                                                                 Downloaded  121.7KB/121.7KB
✔︎ Bottle zeromq (4.3.5_2)                                                                Downloaded    1.7MB/  1.7MB
==> Installing zeromq dependency: libsodium
==> Pouring libsodium--1.0.22.arm64_tahoe.bottle.tar.gz
🍺  /opt/homebrew/Cellar/libsodium/1.0.22: 88 files, 1.1MB
==> Pouring zeromq--4.3.5_2.arm64_tahoe.bottle.tar.gz
🍺  /opt/homebrew/Cellar/zeromq/4.3.5_2: 85 files, 6.7MB
==> Running `brew cleanup zeromq`...
Disable this behaviour by setting `HOMEBREW_NO_INSTALL_CLEANUP=1`.
Hide these hints with `HOMEBREW_NO_ENV_HINTS=1` (see `man brew`).
==> Pouring pkgconf--2.5.1.arm64_tahoe.bottle.tar.gz
🍺  /opt/homebrew/Cellar/pkgconf/2.5.1: 28 files, 532.3KB
==> Running `brew cleanup pkgconf`...
✅ ZMQ dependencies installed.
==== Building UDS tokenizer image llm-d-uds-tokenizer:e2e-test ====
docker build \
		--platform darwin/arm64 \
		--build-arg TARGETOS=darwin \
		--build-arg TARGETARCH=arm64 \
		-t llm-d-uds-tokenizer:e2e-test services/uds_tokenizer
[+] Building 3.7s (2/2) FINISHED                                                               docker:desktop-linux
 => [internal] load build definition from Dockerfile                                                           0.0s
 => => transferring dockerfile: 2.52kB                                                                         0.0s
 => WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 16)                                0.0s
 => WARN: RedundantTargetPlatform: Setting platform to predefined $TARGETPLATFORM in FROM is redundant as thi  0.0s
 => WARN: RedundantTargetPlatform: Setting platform to predefined $TARGETPLATFORM in FROM is redundant as thi  0.0s
 => ERROR [internal] load metadata for docker.io/library/python:3.12-slim                                      3.6s
------
 > [internal] load metadata for docker.io/library/python:3.12-slim:
------

 3 warnings found (use docker --debug to expand):
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 16)
 - RedundantTargetPlatform: Setting platform to predefined $TARGETPLATFORM in FROM is redundant as this is the default behavior (line 16)
 - RedundantTargetPlatform: Setting platform to predefined $TARGETPLATFORM in FROM is redundant as this is the default behavior (line 29)
Dockerfile:16
--------------------
  14 |
  15 |     # Build stage
  16 | >>> FROM --platform=$TARGETPLATFORM python:3.12-slim as builder
  17 |
  18 |     # Set build arguments
--------------------
ERROR: failed to build: failed to solve: python:3.12-slim: failed to resolve source metadata for docker.io/library/python:3.12-slim: no match for platform in manifest: not found
make: *** [image-build-uds] Error 1
```

gyliu513:
Fixed #534 

Fixes the `--platform` flag for container image builds when running on non-Linux hosts (macOS/darwin).

Previously, `TARGETOS` was set via `go env GOOS`, which returns `darwin` on macOS. Passing `--platform darwin/arm64` to Docker is invalid and causes image builds to fail.

This PR introduces `CONTAINER_PLATFORM` that always forces `linux` as the target OS for container builds, regardless of the host OS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,17 @@ VLLM_VERSION := 0.18.0
 TARGETOS ?= $(shell go env GOOS)
 TARGETARCH ?= $(shell go env GOARCH)
 
+# Container images are Linux-based; build them for Linux even on macOS/Windows.
+# (Using $(TARGETOS)/$(TARGETARCH) breaks on macOS, e.g. darwin/arm64 is not a valid
+# container platform for most base images.)
+ifeq ($(TARGETOS),linux)
+CONTAINER_TARGETOS ?= $(TARGETOS)
+else
+CONTAINER_TARGETOS ?= linux
+endif
+CONTAINER_TARGETARCH ?= $(TARGETARCH)
+CONTAINER_PLATFORM ?= $(CONTAINER_TARGETOS)/$(CONTAINER_TARGETARCH)
+
 CONTAINER_TOOL := $(shell { command -v docker >/dev/null 2>&1 && echo docker; } || { command -v podman >/dev/null 2>&1 && echo podman; } || echo "")
 BUILDER := $(shell command -v buildah >/dev/null 2>&1 && echo buildah || echo $(CONTAINER_TOOL))
 UDS_TOKENIZER_IMAGE ?= llm-d-uds-tokenizer:e2e-test
@@ -76,9 +87,9 @@ e2e-test: e2e-test-uds ## Run end-to-end tests
 image-build-uds: check-container-tool ## Build the UDS tokenizer container image
 	@printf "\033[33;1m==== Building UDS tokenizer image $(UDS_TOKENIZER_IMAGE) ====\033[0m\n"
 	$(CONTAINER_TOOL) build \
-		--platform $(TARGETOS)/$(TARGETARCH) \
-		--build-arg TARGETOS=$(TARGETOS) \
-		--build-arg TARGETARCH=$(TARGETARCH) \
+		--platform $(CONTAINER_PLATFORM) \
+		--build-arg TARGETOS=$(CONTAINER_TARGETOS) \
+		--build-arg TARGETARCH=$(CONTAINER_TARGETARCH) \
 		-t $(UDS_TOKENIZER_IMAGE) services/uds_tokenizer
 
 .PHONY: e2e-test-uds
@@ -148,9 +159,9 @@ build-uds: check-go download-zmq ## Build
 image-build: check-container-tool load-version-json ## Build Docker image
 	@printf "\033[33;1m==== Building Docker image $(IMG) ====\033[0m\n"
 	$(CONTAINER_TOOL) build \
-		--platform $(TARGETOS)/$(TARGETARCH) \
-		--build-arg TARGETOS=$(TARGETOS) \
-		--build-arg TARGETARCH=$(TARGETARCH) \
+		--platform $(CONTAINER_PLATFORM) \
+		--build-arg TARGETOS=$(CONTAINER_TARGETOS) \
+		--build-arg TARGETARCH=$(CONTAINER_TARGETARCH) \
 		-t $(IMG) .
 .PHONY: image-push
 image-push: check-container-tool load-version-json ## Push Docker image $(IMG) to registry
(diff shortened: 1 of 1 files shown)


sagearc (changes requested):
Does this fix the full `e2e-test` native windows, or just the image build step?

The `--platform` fix looks good, but `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE=/var/run/docker.sock` is still hardcoded to a unix path, so `testcontainers` might still fail there? There might be more gaps I overlooked though

If native windows is out of scope and WSL is expected (I'd guess that's the common case for windows users), the makefile should already work for windows/WSL users, so the description might be overclaiming on the windows side?

gyliu513:
@sagearc Good catch, thanks for the review!

To clarify the scope: this PR only fixes the `--platform` flag for container image builds on non-Linux hosts (macOS/darwin). The full e2e flow on native Windows is out of scope, we expect Windows users to use WSL, where `go env GOOS` returns `linux` and the Makefile already worked correctly before this PR.

Regarding `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE=/var/run/docker.sock`, this is pre-existing code not introduced by this PR. On native Windows it would indeed fail, which is another reason native Windows is out of scope here.

I've updated the PR title and description to make the scope clearer.

yankay:
Thanks, this looks good to me.

/lgtm

Makefile
```diff
@@ -12,6 +12,17 @@ VLLM_VERSION := 0.18.0
 TARGETOS ?= $(shell go env GOOS)
```
liu-cong, line 12: We should allow the caller to specify the target os and arch, and perhaps default to the os/arch of the build host.

Makefile
```diff
@@ -76,9 +87,9 @@ e2e-test: e2e-test-uds ## Run end-to-end tests
 image-build-uds: check-container-tool ## Build the UDS tokenizer container image
 	@printf "\033[33;1m==== Building UDS tokenizer image $(UDS_TOKENIZER_IMAGE) ====\033[0m\n"
 	$(CONTAINER_TOOL) build \
```
liu-cong, line 89: Shall we use `docker buildx` here for multi-arch support?

gyliu513: The current target already passes `--platform $(CONTAINER_PLATFORM)` to docker build, which handles the single-target cross-platform case (e.g., building a linux/arm64 image on a macOS host). 

docker buildx is needed only if you want to build a single manifest that covers multiple architectures at once. Also, buildx is Docker-specific, podman doesn't use it.

Makefile
gyliu513: I think this is already handled by 
```
TARGETOS ?= $(shell go env GOOS)
TARGETARCH ?= $(shell go env GOARCH) 
```
use `?=`, which allows callers to override
them at invocation time (e.g., make image-build-uds TARGETOS=linux TARGETARCH=amd64) and defaults to the build host's OS/arch. Is that right?

liu-cong: If we already allow callers to specify the targets, IMO we should not override to `linux` here, we should just let the callers specify the right os/arch (and maybe update our dev docs).

gyliu513: Thanks for the feedback. I'd like to keep the current approach because of how make e2e-test works.

The dependency chain is: e2e-test → e2e-test-uds → image-build-uds. When a developer on macOS runs make e2e-test, image-build-uds is triggered automatically as a dependency — there's no opportunity to pass TARGETOS=linux at the command line, and we shouldn't require them to.

The ifeq block ensures that container builds always get a valid platform (linux/arm64) regardless of the host OS, while leaving TARGETOS/TARGETARCH untouched for Go cross-compilation. CONTAINER_PLATFORM is still overridable by the caller if needed. I think this is the right separation of concerns here. Comments?

shmuelk: To specify a TARGETOS of anything other than linux, will require more changes than just the TRAGETOS, the base image for instance comes to mind.

Why not make it easier for the simpler cases?

PR #538, opened 2026-04-24
feat: Add RenderBatchCompletion RPC for multi-prompt tokenization

albertoperdomo2:
## Summary 

With [llm-d-inference-scheduler/pull/860](https://github.com/llm-d/llm-d-inference-scheduler/pull/860/), we fixed the handling of string array prompts. To avoid multiple gRPC roundtrips, this PR adds a `RenderBatchCompletion` gRPC endpoint that accepts multiple prompts in a single call, returning per-prompt token IDs, inspired by [this](https://github.com/llm-d/llm-d-inference-scheduler/pull/860/#discussion_r3128956144) comment. 

It leverages vLLM's native batch support in `OpenAIServingRender`,  the existing render_completion already handles `prompt=["a","b"]` but the current `RenderCompletion` RPC discards all but the first result. Added the `RenderBatch` method to the Go Tokenizer interface with single-prompt optimization (falls back to existing Render to avoid unnecessary overhead). 

> [!NOTE]
> I could not find `protoc 33.2` and the closest I could find is `v33.6`, hence the version bump in the comments. 

## Tests

Added new tests for the RPC endpoint, and mock `RenderBatch` added to `MockTokenizer`:

- TestUdsTokenizer_RenderBatch, multi-prompt happy path
- TestUdsTokenizer_RenderBatchSinglePrompt, single-prompt fallback to Render
- TestUdsTokenizer_RenderBatchEmpty,  nil and empty input
- TestUdsTokenizer_RenderBatchError, server error propagation


albertoperdomo2:
I added a new endpoint instead of updating the existing one because changing `string prompt = 2` to `repeated string prompts = 2` in the proto is wire-compatible but the generated Go field type changes from `string` to `[]string`, so the `RenderCompletion` calls in llm-d-inference-scheduler would fail to compile until updated simultaneously. 

Happy to change it if someone can help me make sure they get merged simultaneously.

albertoperdomo2:
Adding a comment to avoid closing the PR.

PR #542, opened 2026-04-26
build(fs_connector): add cu130 wheel variant for vLLM 0.19.x

kfirtoledo:
## Summary

Ship two prebuilt wheels per arch for the v0.19 line:

- `llmd_fs_connector-0.19-cp312-cp312-linux_{amd64,aarch64}.whl` — default CUDA 12.x build
- `llmd_fs_connector-0.19+cu130-cp312-cp312-linux_{amd64,aarch64}.whl` — CUDA 13.0 variant (PEP 440 local version)

Mirrors the `vllm/vllm-openai` tag convention (`v0.19.1` default / `v0.19.1-cu130` non-default), using the wheel-filename-compatible `+cuXYZ` syntax instead of `-cuXYZ`.

### Build-system changes

- **`Dockerfile.wheel`**: parameterize torch version, CUDA index, and wheel local version; cu130 sets `WHEEL_LOCAL_VERSION=+cu130` so filename and METADATA match.
- **`ci-wheels.yaml`**: `arch × variant` matrix; cu12 uses Dockerfile defaults, cu130 overrides only the three CUDA-related build args.
- **`Dockerfile.dev`**: `VLLM_IMAGE` and `CUDA_TOOLKIT_PKG` as build args.

### DIFF
--- a/.github/workflows/ci-wheels.yaml
+++ b/.github/workflows/ci-wheels.yaml
@@ -15,11 +15,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        arch: [amd64, arm64]
+        variant: [cu12, cu130]
         include:
-          - os: ubuntu-24.04
-            arch: amd64
-          - os: ubuntu-24.04-arm
-            arch: arm64
+          - arch: amd64
+            os: ubuntu-24.04
+          - arch: arm64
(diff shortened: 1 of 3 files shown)


Gregory-Pereira (approved):
Im not certain it works / have not been able to test this but generally it looks good to me

Issue #517, opened 2026-04-14
Deprecate tokenization on the lookup path


vMaroon:
## Background

Before llm-d v0.6, `kvcache.Indexer` accepted a textual prompt and handled rendering, tokenization, and KV-block-key computation in one pass. v0.6 introduced a tokenization plugin in `llm-d-inference-scheduler` that took over the first stage; the indexer began exposing `ScoreTokens` for the tokens-in path.

The goal of this issue is to finish the move: scheduler owns tokenization end-to-end, indexer never touches a prompt string. We do this in three phases so existing consumers have a deprecation window.

## Phase 1 — Signal intent, make internal tokenization optional 

Surface-level changes that mark the deprecated path and let new integrations skip it entirely.

- `kvcache.Config.TokenizersPoolConfig` — marked `Deprecated`, made optional (`omitempty`).
- `kvcache.NewDefaultConfig` — no longer auto-populates `TokenizersPoolConfig`.
- `kvcache.NewKVCacheIndexer` — skips pool construction when config is nil; `Run` becomes a `<-ctx.Done()` no-op in that mode.
- `Indexer.GetPodScores` (prompt-string), `Indexer.ComputeBlockKeys` (prompt-string), `Indexer.SetTokenizer`, and `tokenization.Pool` — marked `Deprecated`. Prompt-string calls return a clear error when the pool is absent.
- `Indexer.ComputeBlockKeysFromTokens` — added as the tokens-in counterpart of `ComputeBlockKeys`.
- Examples and e2e suites that still demonstrate the prompt-string flow opt in explicitly via a new `helper.ConfigureInternalTokenizer`.
- `docs/architecture.md` and `docs/configuration.md` — deprecation callouts pointing at `ScoreTokens` / `ComputeBlockKeysFromTokens`.

## Phase 2 — Migrate remaining consumers

Drive the deprecated APIs to zero callers inside this repo and across the immediate ecosystem.

- Add a tokens-in RPC to `api/indexerpb/indexer.proto` (e.g. `ScoreTokens`); mark the prompt-string `GetPodScores` RPC `[deprecated = true]`. Regenerate stubs.
- Migrate `examples/kv_cache_index_service/server/server.go` and the client to the tokens-in RPC.
- Update `examples/kv_cache_aware_scorer/kvcache_aware_scorer.go` to call `ScoreTokens` directly (it currently goes through `GetPodScores` after constructing a `RenderChatRequest`).
[...]

PR #543, opened 2026-04-26
[1/3] deprecating internal tokenization on the lookup path

vMaroon:
## Summary

Surface-level changes that signal the deprecation of the in-process tokenization pool and let new integrations skip it entirely. Behavior for existing callers is preserved as long as they configure `TokenizersPoolConfig` explicitly.

## What changed

- **`kvcache.Config.TokenizersPoolConfig`** — marked `Deprecated`, made optional (`omitempty`).
- **`kvcache.NewDefaultConfig`** — no longer auto-populates `TokenizersPoolConfig`. New integrations are tokenizer-pool-free by default.
- **`kvcache.NewKVCacheIndexer`** — skips pool construction when config is nil. `Indexer.Run` becomes a `<-ctx.Done()` no-op in that mode.
- **Deprecated APIs** — `Indexer.GetPodScores` (prompt-string), `Indexer.ComputeBlockKeys` (prompt-string), `Indexer.SetTokenizer`, and `tokenization.Pool` carry `Deprecated:` godoc tags. Prompt-string calls return an explicit error when the pool is absent, redirecting callers to `ScoreTokens`.
- **`Indexer.ComputeBlockKeysFromTokens`** — added as the tokens-in counterpart of `ComputeBlockKeys`.
- **Examples and e2e suites** — flows that still demonstrate the prompt-string path opt in explicitly via a new `helper.ConfigureInternalTokenizer`. `helper.ApplyTokenizerEndpoint` is now nil-safe.
- **Docs** — `docs/architecture.md` and `docs/configuration.md` carry deprecation callouts pointing at `ScoreTokens` / `ComputeBlockKeysFromTokens`.

## Migration

Existing consumers that still need the prompt-string flow:

```go
cfg, _ := kvcache.NewDefaultConfig()
poolCfg, _ := tokenization.DefaultConfig()
poolCfg.ModelName = "your-model"
cfg.TokenizersPoolConfig = poolCfg
```

New integrations: tokenize externally and call `Indexer.ScoreTokens` directly. No tokenizer pool config required.

## Related issues
- #517 

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./pkg/... ./tests/integration/...`
- [ ] `make e2e-test-uds` (requires `UDS_TOKENIZER_IMAGE`)
- [ ] `make e2e-test` with `embedded_tokenizers` build tag (requires `Python.h` toolchain)

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -14,7 +14,7 @@ The indexer is built from several modules with separated concerns:
 |:---|:---|:---|
 | `kvcache.Indexer` | Orchestrator — coordinates block-key computation, index lookup, and scoring. | — |
 | `kvblock.TokenProcessor` | Converts a token sequence into a deterministic list of block keys. Reproduces each engine's chained FNV-64a over CBOR content-addressing scheme. | — |
-| `tokenization.Pool` | Worker pool for rendering and tokenizing prompts. Sources tokenizers from a UDS sidecar, from local files, or from HuggingFace Hub. | — |
+| `tokenization.Pool` | Deprecated in-process worker pool backing the prompt-string APIs. New integrations tokenize externally and call `ScoreTokens` directly. | — |
 | `kvblock.Scorer` | Computes per-pod scores from block keys and lookup results. | Longest consecutive prefix match, weighted by device tier. |
 | `kvblock.Index` | The block index itself. Pluggable interface storing `requestKey → []PodEntry` plus the auxiliary `engineKey → requestKey` map. | Two-level in-memory LRU. |
 | `kvevents.Pool` | Sharded worker pool that consumes ZMQ messages, orders them per-pod (FNV-1a on pod ID), and applies them to the index. | — |
@@ -26,7 +26,7 @@ flowchart LR
         direction TB
         Indexer["kvcache.Indexer<br/>(orchestrator)"]
         TP["kvblock.TokenProcessor<br/>(tokens → block keys)"]
-        Tokzr["tokenization.Pool<br/>(UDS / embedded)"]
+        Tokzr["tokenization.Pool<br/>(deprecated)"]
         BScorer["kvblock.Scorer<br/>(longest-prefix match)"]
         Index["kvblock.Index<br/>(block key → pods)"]
         Pool["kvevents.Pool<br/>(sharded ZMQ workers)"]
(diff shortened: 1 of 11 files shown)


docs/architecture.md
```diff
@@ -275,16 +275,27 @@ The library exposes the primitive; the policy (when to insert, what TTL) is deci
...
+> callers that still rely on the deprecated path must set it explicitly.
+
+Efficient tokenization matters because the Read Path computes request keys synchronously per request. Historically the indexer shipped an in-process worker pool with a composite tokenizer:
 
 - **`tokenization.Pool`** — supports both synchronous (for scoring requests — complete results required) and asynchronous (fire-and-forget) modes.
 - **Tokenizer backends**:
   - **`CachedLocalTokenizer`** — loads tokenizers from local files. Useful for air-gapped environments, custom tokenizers, or pre-loaded models. Supports manual mapping, auto-discovery of HuggingFace cache layouts (`models--org--model/snapshots/{hash}/tokenizer.json`), and custom directory structures.
   - **`CachedHFTokenizer`** — downloads and caches tokenizers from HuggingFace. Wraps HF's Rust tokenizers and maintains an LRU cache of active instances.
-  - **`CompositeTokenizer`** (default) — tries backends in order, falling back from local to HuggingFace. Best of both worlds: fast local access when available, remote fetching otherwise.
+  - **`CompositeTokenizer`** — tries backends in order, falling back from local to HuggingFace.
```
sagearc: Embedded tokenization was already deprecated and removed in #473. This part should no longer be maintained, you can remove it instead.

docs/architecture.md
```diff
@@ -275,16 +275,27 @@ The library exposes the primitive; the policy (when to insert, what TTL) is deci
...
 - **Caching** — all tokenizer backends maintain an LRU of loaded tokenizer instances to avoid repeated disk loads.
 
-The library can also be driven by an **external tokenizer sidecar** over a Unix domain socket, in which case the internal `tokenization.Pool` is not used. This is the recommended setup for production deployments (isolates tokenizer downloads and Python dependencies from the host process).
+The library can also be driven by an **external tokenizer sidecar** over a Unix domain socket. This is the path forward: it isolates tokenizer downloads and Python dependencies from the host process, and aligns with the tokens-in API direction.
```
sagearc: It is now supported to configure an external tokenizer from both the EPP and this repo. Maybe we should add some context here to distinguish between the two.

pkg/kvcache/indexer.go
```diff
@@ -123,13 +135,17 @@ func (k *Indexer) KVBlockIndex() kvblock.Index {
 	return k.kvBlockIndex
 }
 
+var errInternalTokenizationDisabled = fmt.Errorf(
```
sagearc: why unexported?

sagearc (changes requested):
Well done! added a few minor comments.
Also note that `examples/kv_cache_aware_scorer/kvcache_aware_scorer.go` might be broken now, it is not included in CI examples so it wouldn't have been caught.

PR #550, opened 2026-04-27
build(fs_backend): make -static-libstdc++ opt-in via FS_PORTABLE_WHEEL

kfirtoledo:
## Problem

`-static-libstdc++` (added in #498 for wheel portability) causes a `std::codecvt` segfault when the connector is used in editable installs or pytest. The static-linked `.so` plus the system `libstdc++.so.6` already loaded by torch / gRPC / etc. results in two libstdc++ copies in the same Python process, which can mismatch on identity-based checks for shared C++ runtime state.

vLLM production runs are unaffected (engine is constructed in a worker subprocess); the crash surfaces in pytest or direct-construction scripts.

## Fix

Make `-static-libstdc++` opt-in via `FS_PORTABLE_WHEEL=1`, set by `Dockerfile.wheel` and the `make wheel` target — so the redistributable wheel is unchanged. The development build (`pip install -e .` / `make build`) is **without** the flag, so libstdc++ is dynamically linked and the bug doesn't trigger.

## Test plan
- [x ] `make build && make test` in `kv_connectors/llmd_fs_backend/` — no segfault
- [x] `make wheel` produces a `.so` with `-static-libstdc++` (`ldd` shows no `libstdc++.so.6`)

Refs: #498


kfirtoledo:
@effi-ofer, PTAL

PR #552, opened 2026-04-28
Add redis lookup and improve nixl lookup 

effi-ofer:
1. Add redis lookup in addition to the default nixl_query and dict lookup.  These lookup methods are only available for object store.
2. Improve nixl_query by using batch lookup.
3. document the usage of the lookup_mode parameter.


kv_connectors/llmd_fs_backend/docs/object_store.md
```diff
@@ -38,6 +38,33 @@ Add the object store configuration and specify the object store backend in  `kv_
 }'
 ```
 
+Object store lookup is performed by default using nixl lookup which resolves to an OBJECT_HEAD call to the object store.  Alternatively, you can switch to using Redis or Valkey by setting the lookup_mode extra config as follows:
```
kfirtoledo, line 41: maybe worth splitting into two sections for each type of lookup and when is better to use each one of them

kv_connectors/llmd_fs_backend/llmd_nixl/lookup/dict.py
```diff
@@ -0,0 +1,50 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""In-memory set-backed block existence lookup."""
+
+from collections.abc import Iterable
+
+
+class DictLookup:
```
kfirtoledo, line 20: Consider moving these into a `lookup/` subpackage with a shared `Protocol` (or ABC) that all three inherit from. Makes the contract explicit and groups the three implementations cleanly.

kv_connectors/llmd_fs_backend/llmd_nixl/lookup/redis.py
```diff
@@ -0,0 +1,89 @@
...
+            )
+        self._ttl = int(cfg.get("redis_ttl", 0))
+        self._client = redis.from_url(redis_url)
+        self._client.ping()
+
+    def exists(self, key: str) -> bool:
+        """Return True if the block identified by key has been recorded."""
+        return bool(self._client.exists(key))
+
+    def lookup(self, keys: Iterable[str]) -> int:
```
kfirtoledo, line 62: The first-key-individual + rest-batched pattern makes sense for `NixlLookup` (NIXL parallelizes HEADs internally, so the split saves N-1 HEADs on a cold miss). For Redis, `pipeline.execute()` is one round-trip regardless of command count — so this costs 2 RTTs on the warm-cache hit case where 1 RTT would suffice. Suggest pipelining all keys in a single `execute()`:

```python
with self._client.pipeline(transaction=False) as pipe:
    for key in keys:
        pipe.exists(key)
    results = pipe.execute()
for i, hit in enumerate(results):
    if not hit:
        return i
return len(results)
```

kv_connectors/llmd_fs_backend/llmd_nixl/manager.py
```diff
@@ -40,32 +42,32 @@ def __init__(
...
         cfg = extra_config or {}
-        self.lookup_mode = cfg.get("lookup_mode", LOOKUP_MODE_NIXL_QUERY)
+        lookup_mode = cfg.get("lookup_mode", LOOKUP_MODE_NIXL_QUERY)
 
-        self._stored_keys: set[str] = set()
-        self._nixl_lookup = None
+        if lookup_mode == LOOKUP_MODE_NIXL_QUERY:
+            self._lookup = NixlLookup(cfg)
+        elif lookup_mode == LOOKUP_MODE_REDIS:
+            from llmd_nixl.redis_lookup import RedisLookup  # lazy: avoids redis import
```
kfirtoledo: `redis` is in `[project.dependencies]` so it ships with every install — the lazy import + comment are misleading. Suggest moving this to a top-level import alongside `NixlLookup` / `DictLookup` and dropping the comment. (If the intent was actually to make redis optional, that needs a `[project.optional-dependencies]` change instead — but as it stands the dep is already required.)

kv_connectors/llmd_fs_backend/llmd_nixl/lookup/redis.py
```diff
@@ -0,0 +1,89 @@
...
+# limitations under the License.
+
+"""Redis-backed block existence lookup."""
+
+from collections.abc import Iterable
+
+import redis
+
+
+class RedisLookup:
```
kfirtoledo, line 23: None of the Redis calls (`exists`, `lookup`, `add`, `add_all`, `ping`) are wrapped in error handling. A transient Redis outage will raise `redis.ConnectionError` up through the offloading manager into vLLM and crash the engine. Cache-lookup failures should be **soft** — log a warning and treat as 0 hits / no-op write. The actual S3 write already succeeded, so a Redis index miss is recoverable; an engine crash is not.

kv_connectors/llmd_fs_backend/llmd_nixl/redis_lookup.py
```diff
@@ -0,0 +1,89 @@
...
+        cfg = extra_config or {}
+        redis_url = cfg.get("redis_url")
+        if not redis_url:
+            raise ValueError(
+                "RedisLookup requires 'redis_url' in extra_config "
+                "(e.g. 'redis://localhost:6379/0')"
+            )
+        self._ttl = int(cfg.get("redis_ttl", 0))
+        self._client = redis.from_url(redis_url)
+        self._client.ping()
```
kfirtoledo: Fail-fast on init is the right call here, but consider wrapping `ping()` to raise a clear, actionable error so users see "Redis at <redis_url> is unreachable" instead of a low-level `redis.ConnectionError` traceback:

```python
try:
    self._client.ping()
except redis.RedisError as e:
    raise RuntimeError(
        f"RedisLookup: cannot reach Redis at {redis_url}: {e}"
    ) from e
```

kv_connectors/llmd_fs_backend/llmd_nixl/manager.py
```diff
@@ -40,32 +42,32 @@ def __init__(
...
 
-        if self.lookup_mode == LOOKUP_MODE_NIXL_QUERY:
-            self._nixl_lookup = NixlLookup(cfg)
+            self._lookup = RedisLookup(cfg)
+        else:
```
kfirtoledo, line 51: Silent fallback hides config typos. A user who writes `"redus"` or `"nixl"` instead of `"nixl_query"` will silently get `dict` mode — which is broken under `mp` (the README-recommended setting) because each worker has its own in-process set. Suggest raising `ValueError(f"Unknown lookup_mode {lookup_mode!r}; expected one of {{nixl_query, redis, dict}}")` instead of warning.

kv_connectors/llmd_fs_backend/docs/object_store.md
```diff
@@ -38,6 +38,33 @@ Add the object store configuration and specify the object store backend in  `kv_
...
+    "secret_key": "minioadmin",
+    "ca_bundle": "/root/tls.crt",
+    "backend": "OBJ",
+    "lookup_mode": "redis",
+    "redis_url": "redis://10.43.176.139:6379/0",
+  }
+}'
+```
+
+lookup_mode takes the following values: "nixl_query" (default), "redis", and "dict". Where "dict" is a local python set lookup method primarily intended for testing and single process usage.
```
kfirtoledo: Worth strengthening this — `dict` mode is not just "primarily" for single-process, it is **only** correct in single-process. Under `mp` (the recommended `--distributed_executor_backend` in the main README) each worker has its own in-process set and they do not share state, so cross-worker cache hits silently never happen. Suggest: *"`dict`: in-process set, intended for testing only. Does not share state across workers — incorrect under `--distributed_executor_backend mp`."*

kv_connectors/llmd_fs_backend/llmd_nixl/redis_lookup.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def __init__(self, extra_config: dict):
+        cfg = extra_config or {}
+        redis_url = cfg.get("redis_url")
+        if not redis_url:
+            raise ValueError(
+                "RedisLookup requires 'redis_url' in extra_config "
+                "(e.g. 'redis://localhost:6379/0')"
+            )
+        self._ttl = int(cfg.get("redis_ttl", 0))
```
kfirtoledo: `0` as a sentinel for "no expiry" is implicit. Consider `None` instead — `cfg.get("redis_ttl")` with no default, then `if self._ttl is not None: ... ex=self._ttl`. Reads more clearly than `if self._ttl > 0:` and matches the `redis-py` convention where `ex=None` means no expiry.

kv_connectors/llmd_fs_backend/llmd_nixl/lookup/redis.py
```diff
@@ -0,0 +1,89 @@
...
+            results = pipe.execute()
+        for i, hit in enumerate(results, start=1):
+            if not hit:
+                return i
+        return 1 + len(rest)
+
+    def add(self, key: str) -> None:
+        """Mark a block key as offloaded."""
+        if self._ttl > 0:
+            self._client.set(key, 1, ex=self._ttl)
```
kfirtoledo, line 96: Storing `key → "1"` for pure existence tracking has per-key value overhead at scale. For millions of keys, `SADD <namespace_set> <key>` (one Redis set holding all keys) would be more memory-efficient. Trade-off: TTL semantics get harder (sets do not support per-member TTL without `redis-stack`). If TTL matters, current design is fine; if not, a single set is leaner. Worth considering for large deployments.

kv_connectors/llmd_fs_backend/llmd_nixl/nixl_lookup.py
```diff
@@ -27,6 +29,8 @@ class NixlLookup:
 
     def __init__(self, extra_config: dict):
         cfg = extra_config or {}
+        if not cfg.get("bucket"):
```
kfirtoledo: `bucket` validation also lives in `ObjBackend.__init__` (along with `endpoint_override` / `access_key` / `secret_key`). Two places enforcing the same constraint with different error messages — worth centralizing in a shared config validator so they cannot drift.

kv_connectors/llmd_fs_backend/docs/object_store.md
kfirtoledo, line 41: Nit: `nixl lookup` → `NIXL lookup`, and `OBJECT_HEAD` → `HEAD` (the HTTP verb).

kv_connectors/llmd_fs_backend/llmd_nixl/lookup/redis.py
effi-ofer: The current redis lookup protocol is: test the first key and if it's there then test all the rest (where the rest can be pretty large).  I suggest we stick with this protocol for now, and do more thorough evaluation as part of the multi-tier lookup work.

kv_connectors/llmd_fs_backend/docs/object_store.md
effi-ofer: I think splitting into multiple sections will just make it more complex to read and consume.  I suggest in any case we wait until we have performance numbers for nixl vs redis before we further delve into this in the docs.

kv_connectors/llmd_fs_backend/llmd_nixl/manager.py
effi-ofer: changed to error.  And BTW, the default is nixl lookup not dictionary.

kv_connectors/llmd_fs_backend/llmd_nixl/lookup/redis.py
effi-ofer: Excellent point.  It boils down to what's the eviction policy.  Lets keep the current implementation and revisit this as part of the multi-tier lookup work.

kv_connectors/llmd_fs_backend/docs/object_store.md
effi-ofer: Accept for nixl --> NIXL.  Kept the OBJECT_HEAD which I believe is the actual object store terminology (or HEAD Object depends on which docs you're reading).

kv_connectors/llmd_fs_backend/llmd_nixl/nixl_lookup.py
effi-ofer: Excellent point.  I removed the check in nixl lookup since it's already done as part of the initialization of the nixl object store backend.

PR #554, opened 2026-04-29
Fix: UDS Tokenizer render error

### LINKED ISSUE #553: UDS Tokenizer render error
Image

yyzxw:
fix: #553


yankay:
Nice catch, thanks @yyzxw! This brings the renderer in line with `tokenizer.py`, which already hard-codes `trust_remote_code=True` in several places.

/lgtm

Minor suggestion (non-blocking, fine as a follow-up): it might be worth unifying all these spots behind an env var like `TRUST_REMOTE_CODE` (default `true` to preserve current behavior), matching the style of `LOG_LEVEL` / `USE_MODELSCOPE` etc. That would let security-sensitive deployments opt out without code changes.

Issue #546, opened 2026-04-27
Remove the assert gpu_blocks_per_file == 1 in object store backend


effi-ofer:
PR #499 is pushed with a limitation on block_size == gpu_block_size.  This issue is being opened to remove this limitation.

PR #558, opened 2026-04-29
Support multiple gpu blocks in a single object store block

effi-ofer:
This PR lifts the assert that gpu blocks per file == 1.  It does this by copying all the gpu blocks into a single staging buffer before writing it to the object store.  Reading is performed using range reads when only a subset of the blocks from a single object is requested.

The PR also updates the test_obj_backend test and makes it more inlined with the test_fs_backend test.

### DIFF
--- a/kv_connectors/llmd_fs_backend/llmd_nixl/obj_backend.py
+++ b/kv_connectors/llmd_fs_backend/llmd_nixl/obj_backend.py
@@ -70,10 +70,9 @@ class ObjBackend(_StagedBackend):
 
     def _get_staging_and_copy(self, block_ids: list) -> tuple:
         num_files = len(block_ids)
-        assert self._staging_pool.qsize() >= num_files, (
-            f"Staging pool exhausted: need {num_files} slots, "
-            f"have {self._staging_pool.qsize()} (pool size={self.io_threads * 8})"
-        )
+        shortfall = num_files - self._staging_pool.qsize()
+        if shortfall > 0:
+            self._extend_staging_pool(shortfall)
         stagings, tensors_out = [], []
         with torch.cuda.stream(self._d2h_stream):
             for block_list in block_ids:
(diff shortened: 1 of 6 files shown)


effi-ofer:
Fixes issue #546

kv_connectors/llmd_fs_backend/llmd_nixl/obj_backend.py
```diff
@@ -69,13 +65,75 @@ def _backend_params(self) -> dict:
             params["ca_bundle"] = self._ca_bundle
         return params
 
+    def _staging_bytes_per_slot(self) -> int:
```
kfirtoledo, line 68: Do we still need  in_StagedBackend the _get_staging_and_copy, _get_staging, _get_blocks_data, _complete_read  
or you can remove them?

kv_connectors/llmd_fs_backend/llmd_nixl/obj_backend.py
```diff
@@ -69,13 +65,75 @@ def _backend_params(self) -> dict:
...
+        return self.gpu_blocks_per_file * len(self.tensors) * self._block_size
+
+    def _get_staging_and_copy(self, block_ids: list) -> tuple:
```
kfirtoledo: Naming nit (applies to all three staging methods, not just this one): the WRITE/READ symmetry is hard to see and `_complete_read` reads like "finish a read" rather than "post-read H2D copy." Aligning with the public API verbs would make the call sites read consistently:

```
_get_staging_and_copy  →  _store_gpu_blocks_to_staging
_get_staging           →  _reserve_staging_for_load
_complete_read         →  _load_staging_to_gpu_blocks
```

Matches `async_store_gpu_blocks` / `async_load_gpu_blocks`. Not blocking.

kfirtoledo (commented):
minor comments

kv_connectors/llmd_fs_backend/llmd_nixl/obj_backend.py
effi-ofer: We don't.  Removed.

kv_connectors/llmd_fs_backend/llmd_nixl/obj_backend.py
effi-ofer: Good suggestions.  Done.

kfirtoledo (approved):

PR #561, opened 2026-04-30
Fix CPU offloading for precise-prefix-cache-aware

mengmeiye:
This PR addresses the issue in CPU offloading where EPP removes GPU PodEntry and deletes engineKey→requestKey mapping. The solution is letting EPP remove GPU PodEntry, but see CPU entry still exists → preserve engineKey→requestKey mapping. 

This PR was tested using [`llm-d-benchmark`](https://github.com/llm-d/llm-d-benchmark) with [`llm-d-kv-cache`](https://github.com/llm-d/llm-d-kv-cache/tree/main) upstream (v0.8.0-rc.1) and [`offloading_index (this PR)`](https://github.com/mengmeiye/llm-d-kv-cache/tree/offloading_index).
The only different between these two experiments is the `inferenceScheduler` image. One uses `v0.8.0-rc.1` tag and the other uses `cpu-offloading`. The scenario file is in https://github.com/mengmeiye/llm-d-benchmark/blob/offloading_stress_test/config/scenarios/guides/precise-prefix-cache-aware-offloading.yaml
To reproduce the experiments:
```bash
llmdbenchmark --spec precise-prefix-cache-aware-offloading standup -p <NAMESPACE>
llmdbenchmark --spec precise-prefix-cache-aware-offloading run -p <NAMESPACE> --monitoring
```

In each of the experiments, we have

- 256 groups × 64 blocks/group = 16,384 unique prefix blocks across the system
- Each pod can hold ~5,359 GPU blocks → after ~84 groups, GPU is full and eviction begins
- With OffloadingConnector, evicted GPU blocks move to CPU
    - Here is the key improvement:
        - `Upstream (v0.8.0-rc.1)` EPP removes GPU PodEntry and deletes engineKey→requestKey mapping. Then, EPP has no record that the block exists anywhere
        - `offloading_index (this PR)` EPP removes GPU PodEntry, but sees CPU entry still exists → preserves engineKey→requestKey mapping. Then, EPP still knows this pod holds the block (on CPU tier). 
- With new requests arriving whose prefix matches those blocks, performance starts making differences: 
    - `Upstream (v0.8.0-rc.1)` The EPP finds score=0 for all pods and falls back to load-balanced routing (random pod). The request likely lands on a pod that doesn't have the prefix cached → cold prefill → high TTFT.
    - `offloading_index (this PR)` The prefix scorer finds a match → routes to the correct pod → the pod already has the KV blocks in GPU (kept hot by repeated affinity routing) → prefix cache hit → low TTFT.
 
TTFT comparison on a GPU cluster: 
- Upstream (`v0.8.0-rc.1`):          mean=1.007s, p50=0.178s,  p95=6.116s,  p99=11.435s
- `offloading_index` (this PR):    mean=0.552s,  p50=0.119s,  p95=1.814s,  p99=8.252s
- Improvement:
    - Mean TTFT improvement:  45.2%
    - P95 TTFT improvement:   70.3%

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -102,6 +102,9 @@ When a block is offloaded from GPU to CPU, the engine emits a `BlockStored` with
 On eviction, the indexer only removes the `engineKey → requestKey` mapping if all associated request keys have no remaining pod entries. This preserves the mapping when other tiers still hold the block.
 
 ```
+  E1 = engine key (the engine's internal content-address hash for the block)
+  R1 = request key (the indexer's own hash, computed from tokens)
+
   1. GPU BlockStored (engine_key=E1, tier=gpu, tokens=[...])
      → computes R1 from tokens, stores E1 → R1, stores R1 → {pod, gpu}
 
@@ -111,6 +114,9 @@ On eviction, the indexer only removes the `engineKey → requestKey` mapping if 
   3. GPU BlockRemoved (engine_key=E1, tier=gpu)
      → evicts R1 → {pod, gpu}, R1 still has {pod, cpu} → keeps E1 → R1
 ```
+
+> [!NOTE]
+> **Event ordering guarantee:** Events from the same pod are processed in order because the event pool shards worker queues by pod identifier (FNV-1a hash). Since both the GPU `BlockStored` and the subsequent CPU offload `BlockStored` originate from the same engine on the same pod, the GPU event (which carries tokens and establishes the `engineKey → requestKey` mapping) is always processed before the CPU event that depends on it. As a defensive measure, if a CPU offload event arrives for an engine key with no existing mapping, the indexer treats it as a graceful no-op (logs at debug level and skips).
 
 ### The Dual-Key Design
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -116,7 +116,7 @@ On eviction, the indexer only removes the `engineKey → requestKey` mapping if 
 ```
 
 > [!NOTE]
-> **Event ordering guarantee:** Events from the same pod are processed in order because the event pool shards worker queues by pod identifier (FNV-1a hash). Since both the GPU `BlockStored` and the subsequent CPU offload `BlockStored` originate from the same engine on the same pod, the GPU event (which carries tokens and establishes the `engineKey → requestKey` mapping) is always processed before the CPU event that depends on it. As a defensive measure, if a CPU offload event arrives for an engine key with no existing mapping, the indexer treats it as a graceful no-op (logs at debug level and skips).
+> **Event ordering guarantee:** A block must exist on GPU before it can be offloaded to CPU, so the engine always publishes the GPU `BlockStored` (with tokens) before the CPU `BlockStored` (without tokens). Both events originate from the same pod, and per-pod event ordering is preserved through the processing pipeline. As a defensive measure, if a CPU offload event arrives for an engine key with no existing mapping (e.g. due to message loss), the indexer treats it as a graceful no-op.
 
 ### The Dual-Key Design
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -51,7 +51,7 @@ The library has two primary data flows: the **Write Path** for ingesting cache e
 
 Model servers publish three event types over ZMQ whenever their KV-cache state changes:
 
-- **`BlockStored`** — blocks with the given content hashes have been created on a specific device tier. Payload includes the chained parent hash, the token chunk, any LoRA ID/name, and any multimodal extra keys.
+- **`BlockStored`** — blocks with the given content hashes have been created on a specific device tier. Payload includes the chained parent hash, the token chunk, any LoRA ID/name, and any multimodal extra keys. For CPU offloading, the engine may emit a `BlockStored` with engine keys and a device tier but **no tokens** — see [CPU Offloading](#cpu-offloading-precise-prefix-caching).
 - **`BlockRemoved`** — blocks with the given hashes have been evicted from a specific device tier and/or attention group.
 - **`AllBlocksCleared`** — the pod dropped its entire cache (a reset). This can occur during RL weight rollouts and other scenarios. The indexer drops all entries associated with the pod via a reverse `pod → request keys` index.
 
@@ -72,9 +72,13 @@ sequenceDiagram
     Adapter-->>Worker: podID, modelName, []Event
 
     loop For each event
-        alt BlockStored
+        alt BlockStored (with tokens)
             Worker->>Worker: Compute request keys from tokens<br/>(hashSeed, parent, extra)
             Worker->>Index: Add(engineKeys, requestKeys, podEntry)
+        else BlockStored (offloading, no tokens)
+            Worker->>Index: GetRequestKey(engineKey)
+            Index-->>Worker: resolved requestKeys
+            Worker->>Index: Add(nil, resolvedKeys, podEntry)
         else BlockRemoved
             Worker->>Index: Evict(engineKey, podEntry)
         else AllBlocksCleared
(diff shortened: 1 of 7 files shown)


pkg/kvcache/kvblock/redis.go
```diff
@@ -287,9 +287,26 @@ func (r *RedisIndex) Evict(ctx context.Context, key BlockHash, keyType KeyType,
...
+				if cmd.Val() > 0 {
+					allEmpty = false
+					break
+				}
+			}
+		} else {
+			allEmpty = false
+		}
+		if allEmpty {
+			if err := r.RedisClient.Del(ctx, redisEngineKey(key)).Err(); err != nil {
```
yankay: TOCTOU between the HLEN pipeline and this `Del`: a concurrent offloading `Add` (pool.go:372, engineKeys=nil) can insert into one of `rks` after HLEN=0 but before Del, leaving an orphan CPU entry with no engine→request mapping. Suggest mirroring `pruneRequestKeyScript` (redis.go:147) with a Lua script over `[engineKey, rk1, rk2, ...]` — atomic, one RTT, matches existing style.

pkg/kvcache/kvblock/in_memory.go
```diff
@@ -244,7 +244,16 @@ func (m *InMemoryIndex) Evict(ctx context.Context, key BlockHash, keyType KeyTyp
...
 		}
-		m.engineToRequestKeys.Remove(key)
+		allEmpty := true
+		for _, rk := range rks {
+			if pc, found := m.data.Get(rk); found && pc != nil && pc.cache.Len() > 0 {
+				allEmpty = false
+				break
+			}
+		}
+		if allEmpty {
```
yankay, line 264: Same race as the Redis path, without a top-level lock. A concurrent offloading `Add` can insert into one of `rks` between the loop and `Remove(key)`. `CostAwareMemoryIndex.Evict` takes `m.mu` for this reason (cost_aware_memory.go:281); suggest making InMemoryIndex consistent.

pkg/kvevents/pool_test.go
```diff
@@ -478,6 +478,181 @@ func TestCanonicalWritePath_ExtraKeysManyToOne(t *testing.T) {
...
+		assert.True(t, tiers["cpu"], "cpu entry should be present")
+	}
+}
+
+// TestBlockStoredEvent_OffloadingUnknownEngineKeys verifies that an offloading
+// event with engine keys not yet in the index is a graceful no-op.
+func TestBlockStoredEvent_OffloadingUnknownEngineKeys(t *testing.T) {
+	ctx := logging.NewTestLoggerIntoContext(context.Background())
+	pool, _, _ := newTestPool(t, 16)
+
```
yankay, line 549: This asserts both `"GPU"` and `"cpu"` — surfaces a casing inconsistency: default tier is `"GPU"` (pool.go:32), event-driven tiers go through `strings.ToLower` (pool.go:267). Suggest lower-casing the default so the same logical tier has one key.

yankay (commented):
Nice work on this fix — TTFT numbers are impressive and the cross-impl test coverage is great. Just left a few inline notes above, not blocking the direction.

mengmeiye:
Hi @yankay Thank you for your comments! I believe the new commit should has resolved the issues. Please feel free to let me know if you have further suggested edits. Thanks!

yankay:
Thanks @mengmeiye for the quick turnaround — the Lua script, the InMemoryIndex mutex, and the lowercase tier fix all address the previous comments cleanly. Nice work!

/lgtm

pkg/kvevents/pool.go
```diff
@@ -349,9 +349,35 @@ func (p *Pool) processEventBatch(ctx context.Context, batch *EventBatch, podIden
 			}
 
 			if len(requestKeys) == 0 {
-				debugLogger.Info("no request keys produced, skipping",
-					"podIdentifier", podIdentifier, "tokenCount", len(ev.Tokens),
-					"blockSize", p.tokenProcessor.BlockSize())
+				// Offloading/location-only event (e.g., DeviceTier=CPU with no
```
vMaroon: Can you update the relevant docs on this behavior?

mengmeiye:
Hi @vMaroon I updated doc on `docs/architecture.md`. Please feel free to let me know if further edits are needed. Thank you!

docs/architecture.md
```diff
@@ -91,6 +95,45 @@ sequenceDiagram
 4. **Engine-specific parsing** — a worker calls `EngineAdapter.ParseMessage()` to decode the engine-specific payload into a batch of generic events.
 5. **Index update** — the worker applies each event to the index.
 
+### CPU Offloading (Precise Prefix Caching)
+
+When an inference engine offloads KV-cache blocks from GPU to CPU memory, the indexer must update its records so that the scorer reflects the new device tier. This is the **precise prefix-cache-aware** approach to CPU offloading — the `precise-prefix-cache-scorer` tracks the actual per-block location reported by the engine via KV-Events. It differs from the **tiered prefix cache** approach, which uses a `cpu-prefix-cache-scorer` (an instance of the `prefix-cache-scorer` plugin with a separate LRU capacity for CPU-tier blocks) that relies on scheduling-history estimation rather than real-time engine state.
```
vMaroon: The 2nd part of this section (on approximate) can be dropped - should stick to the architectural design/facts.

docs/architecture.md
```diff
@@ -91,6 +95,45 @@ sequenceDiagram
...
+2. Engine copies the block to CPU → emits `BlockStored(engineKeys, DeviceTier="cpu")` **with no tokens** (content is unchanged; only the location is new).
+3. Engine evicts the block from GPU → emits `BlockRemoved(engineKey, DeviceTier="gpu")`.
+
+The indexer handles this sequence as follows:
+
+**Token-less `BlockStored` resolution.** A standard `BlockStored` event carries tokens, allowing the indexer to compute request keys. An offloading event carries only engine keys and a device tier — no tokens. When the indexer encounters a `BlockStored` with engine keys but zero tokens, it resolves request keys via the existing `engineKey → requestKey` mapping (established during step 1) and adds the new pod entry (with the CPU device tier) to those request keys. This ensures the scorer sees the block's new location.
+
+**Conditional engine→request mapping removal.** On eviction (step 3), the indexer only removes the `engineKey → requestKey` mapping if **all** associated request keys have no remaining pod entries. Since the CPU entry was added in step 2, the mapping is preserved — the CPU entry still depends on it for future eviction.
```
vMaroon: nit: I feel like verbosity can be reduced without sacrificing clarity.

mengmeiye: Hi @vMaroon, two of your comments should have been resolved. Could you please check the updated doc and let me know? Thanks!

docs/architecture.md
```diff
@@ -91,6 +95,23 @@ sequenceDiagram
...
 
+### CPU Offloading
+
+When a block is offloaded from GPU to CPU, the engine emits a `BlockStored` with engine keys and a device tier but **no tokens** (the content is unchanged). The indexer resolves request keys from the existing `engineKey → requestKey` mapping and adds the new pod entry with the CPU tier.
+
+On eviction, the indexer only removes the `engineKey → requestKey` mapping if all associated request keys have no remaining pod entries. This preserves the mapping when other tiers still hold the block.
+
+```
+  1. GPU BlockStored (engine_key=E1, tier=gpu, tokens=[...])
+     → computes R1 from tokens, stores E1 → R1, stores R1 → {pod, gpu}
```
liu-cong, line 109: what is R1 here? can you clarify

docs/architecture.md
```diff
@@ -91,6 +95,23 @@ sequenceDiagram
 4. **Engine-specific parsing** — a worker calls `EngineAdapter.ParseMessage()` to decode the engine-specific payload into a batch of generic events.
 5. **Index update** — the worker applies each event to the index.
 
+### CPU Offloading
+
+When a block is offloaded from GPU to CPU, the engine emits a `BlockStored` with engine keys and a device tier but **no tokens** (the content is unchanged). The indexer resolves request keys from the existing `engineKey → requestKey` mapping and adds the new pod entry with the CPU tier.
```
liu-cong, line 100: Is it guaranteed that the kv events for the GPU tier arrives first so that events of CPU tier (without tokens) can be correctly resolved?

docs/architecture.md
mengmeiye: Good catch! Just added clarification on `E1` and `R1`.

docs/architecture.md
mengmeiye: Good question. I just added a note in lines 188-189 there. To answer your question in details,
the guarantee comes from three layers:

  1. `Engine-side causality` - You can't offload a block to CPU that was never on GPU. The engine must first allocate and store the block on GPU (emitting BlockStored with tokens), and only later decide to offload it to CPU (emitting BlockStored without tokens). This is a physical constraint of the offloading operation itself.
  2. `ZMQ wire ordering` - Both events are published from the same engine process over the same ZMQ PUB socket. ZMQ guarantees messages from a single publisher arrive at the subscriber in the order they were sent.
  3. `Pool sharding` - Both events carry the same pod identifier in the topic, so ShardingKey() routes them to the same worker queue. The worker processes that queue in FIFO order (`pool.go`:`174` — sequential `Get()` loop).

Therefore, the engine publishes GPU first (causality) → the subscriber receives GPU first (ZMQ ordering) → the worker processes GPU first (same shard, FIFO). By the time the CPU event is processed, the engineKey → requestKey mapping already exists.

The defensive no-op handles the one edge case that breaks this: ZMQ PUB/SUB is unreliable, so if the connection drops and the GPU event is lost while the CPU event gets through on reconnect, the lookup simply finds nothing and skips.

Please feel free to let me know if you have further questions/concerns. Thanks!

vMaroon: The above is currently true, but longitivtiy of storage offloading changes the picture. And with tiered offloading, we do get a path of storage -> CPU that may effectively outlive the scheduler/gpu view. This path needs further investigation.

mengmeiye: Ah excellent point!

mengmeiye: Actually let me add a note there for this.

docs/architecture.md
```diff
@@ -91,6 +95,31 @@ sequenceDiagram
...
+     → resolves E1 → R1, adds R1 → {pod, cpu}
+
+  3. GPU BlockRemoved (engine_key=E1, tier=gpu)
+     → evicts R1 → {pod, gpu}, R1 still has {pod, cpu} → keeps E1 → R1
+```
+
+> [!NOTE]
+> **Event ordering guarantee:** A block must exist on GPU before it can be offloaded to CPU, so the engine always publishes the GPU `BlockStored` (with tokens) before the CPU `BlockStored` (without tokens). Both events originate from the same pod, and per-pod event ordering is preserved through the processing pipeline. As a defensive measure, if a CPU offload event arrives for an engine key with no existing mapping (e.g. due to message loss), the indexer treats it as a graceful no-op.
+>
+> This ordering assumption holds for the current GPU→CPU offloading path. Tiered offloading (e.g. storage→CPU) may introduce paths where the original GPU event is no longer present in the indexer, requiring further investigation.
```
vMaroon, line 121: This path is mostly relevant for prefetching or proactive transfers. Effectively these paths introduce a different source of indexing which will be driven by the scheduler similarly to speculative entries.

vMaroon (approved):

PR #563, opened 2026-05-02
fix: evict all index entries on AllBlocksClearedEvent

### LINKED ISSUE #537: When an inference pod is deleted, the kv cache info should be deleted.
For now, It seems that when a pod is deleted, only the subscriber for that pod is deleted, the existing kv cache info on binding to that pod is not deleted.

Btw, for now, the `AllBlocksClearedEvent` event does not clear the existing kv cache info binding to that pod.
https://github.com/llm-d/llm-d-kv-cache/blob/066f89a1c3428e04ace5645136418b93609a1a37/pkg/kvevents/pool.go#L388-L392

Jwrede:
Fixes #537

## Problem

When `AllBlocksClearedEvent` fires (pod deletion or full cache clear), the handler in `pool.go` only logs the event. The pod's entries remain in the index indefinitely, causing:
- Memory leak from accumulated stale entries
- Suboptimal routing (scorer returns hits for pods that no longer hold the data)

## Fix

Add `EvictByPod(ctx, podIdentifier string) error` to the `Index` interface and implement it for all backends:

- **InMemoryIndex**: iterates all request keys, removes entries matching the pod, cleans up empty keys and orphaned engine key mappings
- **CostAwareMemoryIndex**: same logic adapted for Ristretto cache with sync.Map pod caches
- **RedisIndex**: SCAN-based iteration, removes matching hash fields, prunes empty keys
- **tracedIndex / instrumentedIndex**: delegate to inner implementation

Call `EvictByPod` from the `AllBlocksClearedEvent` handler in `pool.go`.

## Testing

Two new tests in `pool_test.go`:
- `TestAllBlocksClearedEvent_EvictsAllPodEntries` -- verifies cleared pod is removed while other pods remain
- `TestAllBlocksClearedEvent_RemovesEngineKeyMappings` -- verifies engine key mappings are cleaned up when the only pod is cleared

All existing tests pass. Race detector clean.

### DIFF
--- a/pkg/kvcache/kvblock/cost_aware_memory.go
+++ b/pkg/kvcache/kvblock/cost_aware_memory.go
@@ -336,6 +336,91 @@ func (m *CostAwareMemoryIndex) evictPodsFromRequestKey(
 	}
 }
 
+// EvictByPod removes all entries associated with the given pod identifier from the index.
+// It iterates all request keys and removes the pod from each CostPodCache. Request keys
+// that become empty are removed. Engine key mappings that point to removed request keys
+// are also cleaned up.
+func (m *CostAwareMemoryIndex) EvictByPod(ctx context.Context, podIdentifier string) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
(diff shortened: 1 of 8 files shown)


Jwrede:
Closing in favor of #437 which implements this with a reverse index approach.

Issue #520, opened 2026-04-14
Add storage KV-cache indexing support


albertoperdomo2:
## Summary

The Go indexer has zero visibility into storage-tier (FS backend) KV blocks. This means routing decisions are made without any awareness of what's cached on disk, and the `pvc_evictor` silently stales any future storage index entries with no cleanup path.

## Problems to solve

**1. Indexer blindness to storage tier**
The FS backend does not emit `BlockStored`/`BlockRemoved` ZMQ events. The Go indexer only knows about GPU/CPU blocks; storage blocks are invisible to `LongestPrefixScorer`.

**2. Block-size heterogeneity**
~~`chunkedTokenDatabase` uses a single configured `blockSize`. Storage offloads at `offloaded_block_size` (default 256 tokens) — a different, independent FNV-64a hash chain. `Pool.processEventBatch()` ignores the `BlockSize` field in `BlockStoredEvent` and always re-derives keys at the GPU block size.~~

- [x] Done in this PR: [#511](https://github.com/llm-d/llm-d-kv-cache/pull/)

**3. Index scale for shared storage**
A shared FS serving dozens of replicas accumulates millions of block files. Indexing every storage block would replicate the GPU index's scalability problem at a coarser resolution. The `pvc_evictor` deletes files via `xargs rm -f` with no event emission, so entries go stale silently.

## Proposal

Initial proposal is being tracked in this document: [[PUBLIC] Proposal: Storage KV-cache indexing for llm-d-kv-cache](https://docs.google.com/document/d/18Y2c4BXJ8E8q8ZZtfV1PSbuVNR0eSI5QjxsP0Z01inE/edit?hl=es&tab=t.0)

zdtsw:
is this still needed after we have switched to vllm's tiered kv-cache?  @albertoperdomo2

albertoperdomo2:
@zdtsw I do think so, since the KVEvents for secondary tiers are fundamentally different (distributed), but I'll need to confirm this. I have most of it already implemented, I just need a bit of time to evaluate the performance and prepare the PR (in the `llm-d-router` now) .

PR #571, opened 2026-05-05
feat: Emit storage KV events from the fs backend

albertoperdomo2:
## Summary

This PR adds storage-tier `BlockStored` event emission for the llm-d fs backend as part of the larger storage-aware KV cache indexing effort. When POSIX or OBJ/NIXL storage offload completes successfully, the connector can publish storage events through a ZMQ publisher configured via `storage_events_endpoint`.

The emitted events use the same vLLM KV event transport contract expected by the Go event subscriber: a three-frame ZMQ message containing topic, sequence, and msgpack payload. The event payload follows the existing vLLM positional `BlockStored` shape, uses `SHARED_STORAGE` as the medium/source, and normalizes block hashes to the lower 64 bits to match the fs backend `FileMapper` behavior.

The publisher is fail-open so storage success is not blocked by event publishing errors, and NIXL/OBJ manager wiring is covered so both backend paths emit events consistently.

## Testing

All existing tests pass, and this PR adds focused storage-event unit coverage.

Validated with:

```
pytest kv_connectors/llmd_fs_backend/tests/test_storage_events.py -q
```

## Related Issues

- #520

### DIFF
--- a/kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
+++ b/kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
@@ -93,12 +93,12 @@ class StorageEventPublisher:
             return
 
         event = [
-            "BlockStored",       # [0] tag
-            hashes,              # [1] block_hashes (all hashes from this complete_store call)
-            0,                   # [2] parent_hash (unknown at storage tier)
-            [],                  # [3] token_ids (empty)
-            0,                   # [4] block_size (unused)
-            None,                # [5] lora_id
+            "BlockStored",  # [0] tag
+            hashes,  # [1] block_hashes (all hashes from this complete_store call)
+            0,  # [2] parent_hash (unknown at storage tier)
+            [],  # [3] token_ids (empty)
+            0,  # [4] block_size (unused)
+            None,  # [5] lora_id
             self._medium.value,  # [6] medium / device tier
         ]
         self._send_batch([msgpack.packb(event, use_bin_type=True)])
@@ -113,9 +113,7 @@ class StorageEventPublisher:
             if self._closed:
                 return
 
-            payload = msgpack.packb(
-                [time.time(), packed_events], use_bin_type=True
-            )
+            payload = msgpack.packb([time.time(), packed_events], use_bin_type=True)
             self._seq += 1
             self._socket.send_multipart(
                 [
--- a/kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
+++ b/kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
@@ -25,6 +25,7 @@ import zmq
 logger = logging.getLogger(__name__)
 
 _UINT64_MASK = (1 << 64) - 1
+DEFAULT_STORAGE_EVENTS_HWM = 100_000  # vLLM's default
 
 
 class StorageMedium(enum.Enum):
@@ -54,6 +55,7 @@ class StorageEventPublisher:
         self,
         endpoint: str,
         model_name: str,
+        sndhwm: int = DEFAULT_STORAGE_EVENTS_HWM,
         medium: StorageMedium = StorageMedium.SHARED_STORAGE,
     ) -> None:
         """Bind a ZMQ PUB socket on *endpoint* and configure the topic prefix.
@@ -62,11 +64,12 @@ class StorageEventPublisher:
             endpoint: ZMQ bind address (e.g. ``tcp://*:5559``).
             model_name: Model identifier included in the topic string.
             medium: Storage backend type embedded in the topic and each event.
+            sndhwm: ZMQ send high-water mark.
         """
         self._ctx = zmq.Context()
         self._socket = self._ctx.socket(zmq.PUB)
         self._socket.setsockopt(zmq.LINGER, 0)
-        self._socket.setsockopt(zmq.SNDHWM, 1000)
+        self._socket.setsockopt(zmq.SNDHWM, sndhwm)
         self._socket.bind(endpoint)
 
         self._model_name = model_name
--- a/kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
+++ b/kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
@@ -1,0 +1,109 @@
+# Copyright 2026 The llm-d Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import struct
+import threading
+import time
+from collections.abc import Iterable
+
+import msgpack
+import zmq
+
+logger = logging.getLogger(__name__)
+
+_UINT64_MASK = (1 << 64) - 1
+
+
+def _hash_to_uint64(block_hash: int | bytes) -> int:
+    """Mask each hash to lower 64 bits, matching the FileMapper."""
+    if isinstance(block_hash, bytes):
+        return int.from_bytes(block_hash, "big") & _UINT64_MASK
+    return int(block_hash) & _UINT64_MASK
+
+
+class StorageEventPublisher:
+    """Publishes storage-tier KV cache events via ZMQ PUB socket.
+
+    Events use the same msgpack positional-array format as vLLM's GPU
+    KV events so the Go vLLMAdapter can parse them without modification.
+    ZMQ messages use the 3-frame format expected by zmq_subscriber.go:
+    [topic, sequence, payload].
+    """
+
+    def __init__(
+        self, endpoint: str, model_name: str, source_id: str = "SHARED_STORAGE"
+    ):
+        self._ctx = zmq.Context()
+        self._socket = self._ctx.socket(zmq.PUB)
+        self._socket.setsockopt(zmq.LINGER, 0)
+        self._socket.setsockopt(zmq.SNDHWM, 1000)
+        self._socket.bind(endpoint)
+
+        self._model_name = model_name
+        self._source_id = source_id
+        self._topic = f"kv@{self._source_id}@{self._model_name}"
+        self._seq: int = 0
+        self._closed = False
+        self._send_lock = threading.Lock()
+        logger.info(
+            "StorageEventPublisher bound to %s (topic: %s)",
+            endpoint,
+            self._topic,
+        )
+
+    def publish_blocks_stored(self, block_hashes: Iterable[int | bytes]) -> None:
+        hashes = [_hash_to_uint64(h) for h in block_hashes]
+        if not hashes:
+            return
+
+        events: list[bytes] = []
+        for h in hashes:
+            event = [
+                "BlockStored",  # [0] tag
+                [h],  # [1] block_hashes (one per file)
+                0,  # [2] parent_hash (unused)
+                [],  # [3] token_ids (empty)
+                0,  # [4] block_size (unused)
+                None,  # [5] lora_id
+                "SHARED_STORAGE",  # [6] medium / device tier
+            ]
+            events.append(msgpack.packb(event, use_bin_type=True))
+
+        self._send_batch(events)
+
+    def _send_batch(self, packed_events: list[bytes]) -> None:
+        batch = [time.time(), packed_events]
+        payload = msgpack.packb(batch, use_bin_type=True)
+
+        with self._send_lock:
+            if self._closed:
+                return
+
+            self._seq += 1
+            self._socket.send_multipart(
+                [
+                    self._topic.encode("utf-8"),
+                    struct.pack(">Q", self._seq),
+                    payload,
+                ]
+            )
+
+    def close(self) -> None:
+        with self._send_lock:
+            if self._closed:
+                return
+            self._closed = True
+            self._socket.close()
+            self._ctx.term()
--- a/kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
+++ b/kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
@@ -92,20 +92,16 @@ class StorageEventPublisher:
         if not hashes:
             return
 
-        events: list[bytes] = []
-        for h in hashes:
-            event = [
-                "BlockStored",  # [0] tag
-                [h],  # [1] block_hashes (one per file)
-                0,  # [2] parent_hash (unused)
-                [],  # [3] token_ids (empty)
-                0,  # [4] block_size (unused)
-                None,  # [5] lora_id
-                self._medium.value,  # [6] medium / device tier
-            ]
-            events.append(msgpack.packb(event, use_bin_type=True))
-
-        self._send_batch(events)
+        event = [
+            "BlockStored",       # [0] tag
+            hashes,              # [1] block_hashes (all hashes from this complete_store call)
+            0,                   # [2] parent_hash (unknown at storage tier)
+            [],                  # [3] token_ids (empty)
+            0,                   # [4] block_size (unused)
+            None,                # [5] lora_id
+            self._medium.value,  # [6] medium / device tier
+        ]
+        self._send_batch([msgpack.packb(event, use_bin_type=True)])
 
     def _send_batch(self, packed_events: list[bytes]) -> None:
         """Send a batch of pre-packed events as a 3-frame ZMQ message.
@@ -113,13 +109,13 @@ class StorageEventPublisher:
         Frames: ``[topic, sequence, payload]``.  Thread-safe; silently
         drops the message if the publisher has been closed.
         """
-        batch = [time.time(), packed_events]
-        payload = msgpack.packb(batch, use_bin_type=True)
-
         with self._send_lock:
             if self._closed:
                 return
 
+            payload = msgpack.packb(
+                [time.time(), packed_events], use_bin_type=True
+            )
             self._seq += 1
             self._socket.send_multipart(
                 [
--- a/kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
+++ b/kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import enum
 import logging
 import struct
 import threading
@@ -26,8 +27,15 @@ logger = logging.getLogger(__name__)
 _UINT64_MASK = (1 << 64) - 1
 
 
+class StorageMedium(enum.Enum):
+    """Storage backend types used as the device-tier label in events."""
+
+    SHARED_STORAGE = "SHARED_STORAGE"
+    OBJECT_STORE = "OBJECT_STORE"
+
+
 def _hash_to_uint64(block_hash: int | bytes) -> int:
-    """Mask each hash to lower 64 bits, matching the FileMapper."""
+    """Mask a block hash to its lower 64 bits to match the FileMapper truncation."""
     if isinstance(block_hash, bytes):
         return int.from_bytes(block_hash, "big") & _UINT64_MASK
     return int(block_hash) & _UINT64_MASK
@@ -43,8 +51,18 @@ class StorageEventPublisher:
     """
 
     def __init__(
-        self, endpoint: str, model_name: str, source_id: str = "SHARED_STORAGE"
-    ):
+        self,
+        endpoint: str,
+        model_name: str,
+        medium: StorageMedium = StorageMedium.SHARED_STORAGE,
+    ) -> None:
+        """Bind a ZMQ PUB socket on *endpoint* and configure the topic prefix.
+
+        Args:
+            endpoint: ZMQ bind address (e.g. ``tcp://*:5559``).
+            model_name: Model identifier included in the topic string.
+            medium: Storage backend type embedded in the topic and each event.
+        """
         self._ctx = zmq.Context()
         self._socket = self._ctx.socket(zmq.PUB)
         self._socket.setsockopt(zmq.LINGER, 0)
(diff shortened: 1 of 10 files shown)


yankay (commented):
Thanks! Verified locally with pytest --noconftest, all 7 storage event tests passed.

/lgtm

vMaroon:
cc @kfirtoledo

kv_connectors/llmd_fs_backend/llmd_fs_backend/spec.py
```diff
@@ -103,20 +105,46 @@ def __init__(self, vllm_config: VllmConfig, kv_cache_config: KVCacheConfig):
             dtype=dtype,
         )
```
kfirtoledo, line 104: lets add an option to disable the publisher  by flag

kv_connectors/llmd_fs_backend/llmd_fs_backend/spec.py
```diff
@@ -103,20 +105,46 @@ def __init__(self, vllm_config: VllmConfig, kv_cache_config: KVCacheConfig):
...
 
+        self._storage_events_endpoint = self.extra_config.get(
+            "storage_events_endpoint", None
+        )
+
+    def _create_event_publisher(self):
```
kfirtoledo: I think this should be part of the manager

kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
```diff
@@ -0,0 +1,109 @@
...
+        events: list[bytes] = []
+        for h in hashes:
+            event = [
+                "BlockStored",  # [0] tag
+                [h],  # [1] block_hashes (one per file)
+                0,  # [2] parent_hash (unused)
+                [],  # [3] token_ids (empty)
+                0,  # [4] block_size (unused)
+                None,  # [5] lora_id
+                "SHARED_STORAGE",  # [6] medium / device tier
```
kfirtoledo: This should be an enum we can have multiple types of backend fs_shared_storage  or object_store, the name probably should be initialized when you create it in the manager, you can use backend fields

kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
```diff
@@ -0,0 +1,109 @@
...
+        self._seq: int = 0
+        self._closed = False
+        self._send_lock = threading.Lock()
+        logger.info(
+            "StorageEventPublisher bound to %s (topic: %s)",
+            endpoint,
+            self._topic,
+        )
+
+    def publish_blocks_stored(self, block_hashes: Iterable[int | bytes]) -> None:
```
kfirtoledo, line 87: Can you description of each function

kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
```diff
@@ -0,0 +1,109 @@
...
+                [],  # [3] token_ids (empty)
+                0,  # [4] block_size (unused)
+                None,  # [5] lora_id
+                "SHARED_STORAGE",  # [6] medium / device tier
+            ]
+            events.append(msgpack.packb(event, use_bin_type=True))
+
+        self._send_batch(events)
+
+    def _send_batch(self, packed_events: list[bytes]) -> None:
```
kfirtoledo, line 109: Is it going to impact the performance?

kfirtoledo (changes requested):
Looks good, minor changes

albertoperdomo2:
I'll address the comments, thanks @kfirtoledo - btw, I just tested it live and starting the server with:

```bash
vllm serve Qwen/Qwen3-0.6B \
    --gpu-memory-utilization 0.85 \
    --enable-chunked-prefill \
    --kv-transfer-config '{
      "kv_role": "kv_both",
        "spec_module_path": "llmd_fs_backend.spec",
        "shared_storage_path": "/mnt/files-storage/kv-cache/",
        "block_size": 256,
        "threads_per_gpu": "64",
        "enable_events": true,
        "storage_events_endpoint": "tcp://*:5559"
      }
    }'
  ```

properly published events:

```bash
Subscribed to tcp://localhost:5559 with filter 'kv@'
Waiting for storage events...

[seq=1] topic=kv@SHARED_STORAGE@Qwen/Qwen3-0.6B
  BlockStored: hashes=[12834380269199539485] tier=SHARED_STORAGE

[seq=2] topic=kv@SHARED_STORAGE@Qwen/Qwen3-0.6B
  BlockStored: hashes=[10573836138397579963] tier=SHARED_STORAGE

[seq=3] topic=kv@SHARED_STORAGE@Qwen/Qwen3-0.6B
  BlockStored: hashes=[10573836138397579963] tier=SHARED_STORAGE
```

kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
albertoperdomo2: Running throughput and stress tests now 👍🏽

albertoperdomo2:
## Throughput Benchmark Results

### No events enabled (baseline)

- Command:
```
python -m tests.performance.test_throughput \
    --backend=storage \
    --model=Qwen/Qwen3-0.6B \
    --num-requests=40 \
    --num-tokens=10000
```

- Results:
```
[RESULTS] backend=storage  model=Qwen/Qwen3-0.6B
  Input tokens: 10000
  Cold (req 1):          0.425s
  Hot avg (req 2..40):   0.077s
  Last-10 avg:           0.076s
  Total (40 reqs):       3.445s
  Throughput (KV):       14.06 GB/s
(EngineCore pid=488) INFO 05-12 10:25:24 [core.py:1210] Shutdown initiated (timeout=0)
(EngineCore pid=488) INFO 05-12 10:25:24 [core.py:1233] Shutdown complete
[CLEANUP] Finish Removed /tmp/fs_connector_perf
```

### Events enabled

- Command:
```
python -m tests.performance.test_throughput \
    --backend=storage \
    --model=Qwen/Qwen3-0.6B \
    --num-requests=40 \
    --num-tokens=10000 \
    --enable-events
```

- Results:
```
[RESULTS] backend=storage  model=Qwen/Qwen3-0.6B
  Input tokens: 10000
  Cold (req 1):          0.406s
  Hot avg (req 2..40):   0.077s
  Last-10 avg:           0.076s
  Total (40 reqs):       3.427s
  Throughput (KV):       14.14 GB/s
(EngineCore pid=967) INFO 05-12 10:30:05 [core.py:1210] Shutdown initiated (timeout=0)
(EngineCore pid=967) INFO 05-12 10:30:05 [core.py:1233] Shutdown complete
[CLEANUP] Finish Removed /tmp/fs_connector_perf
```

## Stress Test Benchmark Results

### No events enabled (baseline)

- Command:
```
python -m tests.performance.test_stress \
    --backend=storage \
    --model=Qwen/Qwen3-0.6B \
    --num-tokens=2000 \
    --batch-size=8 \
    --num-iterations=2 \
    --hot-ratios=0.0,0.5,1.0
```

- Results:
```
[sweep 1/1] ratio=0.00  pool_size=0
  iter  1/2: 0.300s (26.70 req/s)
  iter  2/2: 0.237s (33.81 req/s)

[sweep 1/1] ratio=0.50  pool_size=16
  iter  1/2: 0.156s (51.23 req/s)
  iter  2/2: 0.165s (48.52 req/s)

[sweep 1/1] ratio=1.00  pool_size=24
  iter  1/2: 0.117s (68.50 req/s)
  iter  2/2: 0.116s (68.70 req/s)

==================== Per-ratio summary ====================
ratio   rep1 mean      p50      p99   KV GB/s
---------------------------------------------
 0.00         0.268    0.268    0.300      6.37
 0.50         0.161    0.161    0.165     10.65
 1.00         0.117    0.117    0.117     14.65

[RESULTS] backend=storage  model=Qwen/Qwen3-0.6B
  Total prompts: 48, wall time: 1.09s, pool_size=24
  Steady-state (repeat 1, ratio=1.0): tokens/s=137,197  KV=14.65 GB/s  mean=0.117s p50=0.117s p99=0.117s p100=0.117s
(EngineCore pid=2213) INFO 05-12 11:04:36 [core.py:1210] Shutdown initiated (timeout=0)
(EngineCore pid=2213) INFO 05-12 11:04:36 [core.py:1233] Shutdown complete
[CLEANUP] Finish Removed /tmp/fs_connector_stress
```

### Events enabled

- Command:
```
python -m tests.performance.test_stress \
    --backend=storage \
    --model=Qwen/Qwen3-0.6B \
    --num-tokens=2000 \
    --batch-size=8 \
    --num-iterations=2 \
    --hot-ratios=0.0,0.5,1.0 \
    --enable-events
```

- Results:
```
[sweep 1/1] ratio=0.00  pool_size=0
  iter  1/2: 0.277s (28.86 req/s)
  iter  2/2: 0.256s (31.21 req/s)

[sweep 1/1] ratio=0.50  pool_size=16
  iter  1/2: 0.171s (46.83 req/s)
  iter  2/2: 0.177s (45.22 req/s)

[sweep 1/1] ratio=1.00  pool_size=24
  iter  1/2: 0.116s (68.76 req/s)
  iter  2/2: 0.108s (73.97 req/s)

==================== Per-ratio summary ====================
ratio   rep1 mean      p50      p99   KV GB/s
---------------------------------------------
 0.00         0.267    0.267    0.277      6.41
 0.50         0.174    0.174    0.177      9.83
 1.00         0.112    0.112    0.116     15.23

[RESULTS] backend=storage  model=Qwen/Qwen3-0.6B
  Total prompts: 48, wall time: 1.11s, pool_size=24
  Steady-state (repeat 1, ratio=1.0): tokens/s=142,545  KV=15.23 GB/s  mean=0.112s p50=0.112s p99=0.116s p100=0.116s
(EngineCore pid=2546) INFO 05-12 11:09:24 [core.py:1210] Shutdown initiated (timeout=0)
(EngineCore pid=2546) INFO 05-12 11:09:24 [core.py:1233] Shutdown complete
```

cc: @kfirtoledo

kv_connectors/llmd_fs_backend/llmd_fs_backend/manager.py
```diff
@@ -34,8 +34,63 @@ class SharedStorageOffloadingManager(OffloadingManager):
...
+            return None
+
+        try:
+            from llmd_fs_backend.event_publisher import (
+                StorageEventPublisher,
+                StorageMedium,
+            )
+
+            backend = extra_config.get("backend", "POSIX")
+            medium = (
```
kfirtoledo: maybe also part of the config

kv_connectors/llmd_fs_backend/llmd_fs_backend/spec.py
```diff
@@ -33,6 +34,7 @@
 )
 
 DEFAULT_STORAGE_BLOCK_SIZE = 256
+logger = logging.getLogger(__name__)
```
kfirtoledo: use from llmd_fs_backend import _logger as logger

kv_connectors/llmd_fs_backend/llmd_fs_backend/spec.py
```diff
@@ -106,17 +108,21 @@ def __init__(self, vllm_config: VllmConfig, kv_cache_config: KVCacheConfig):
...
         assert self.vllm_config.parallel_config.rank == 0, "Scheduler rank should be 0"
         if not self._manager:
+            model_name = self.vllm_config.model_config.model
             backend = self.extra_config.get("backend", "POSIX")
             if backend == "OBJ":
                 from llmd_nixl.manager import NixlStorageOffloadingManager
 
                 self._manager = NixlStorageOffloadingManager(
                     file_mapper=self.file_mapper,
+                    model_name=model_name,
```
kfirtoledo: You can get the model name from the file_mapper

kv_connectors/llmd_fs_backend/llmd_fs_backend/manager.py
```diff
@@ -97,6 +152,7 @@ def prepare_store(
 
     def complete_store(self, block_hashes: Iterable[BlockHash], success: bool = True):
         """
-        For shared storage, storing is stateless - no action needed.
+        For shared storage, storing is stateless but we emit events for stored blocks.
         """
-        pass
+        if success:
+            self._publish_blocks_stored(block_hashes)
```
kfirtoledo: Resource leak: `vllm.v1.kv_offload.abstract.OffloadingManager` declares `shutdown()` as the lifecycle teardown hook (default no-op). The publisher binds a ZMQ socket + context at construction and is never closed during normal vLLM teardown. Override it here so the socket is released when the engine tears down:

```suggestion
            self._publish_blocks_stored(block_hashes)

    def shutdown(self) -> None:
        if self._event_publisher is not None:
            self._event_publisher.close()
```

kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
```diff
@@ -0,0 +1,139 @@
...
+                [h],  # [1] block_hashes (one per file)
+                0,  # [2] parent_hash (unused)
+                [],  # [3] token_ids (empty)
+                0,  # [4] block_size (unused)
+                None,  # [5] lora_id
+                self._medium.value,  # [6] medium / device tier
+            ]
+            events.append(msgpack.packb(event, use_bin_type=True))
+
+        self._send_batch(events)
```
kfirtoledo: The vLLM `BlockStored` wire format carries `block_hashes: []hash` (a list), and the Go adapter (`convertBlockStoredEvent`) handles multi-hash batches natively. Creating N single-element-list events per call multiplies msgpack overhead and per-event work on the Go side for no gain — the medium/lora/parent fields are identical across them. Batch them into one event:

```suggestion
        event = [
            "BlockStored",       # [0] tag
            hashes,              # [1] block_hashes (all hashes from this complete_store call)
            0,                   # [2] parent_hash (unknown at storage tier)
            [],                  # [3] token_ids (empty)
            0,                   # [4] block_size (unused)
            None,                # [5] lora_id
            self._medium.value,  # [6] medium / device tier
        ]
        self._send_batch([msgpack.packb(event, use_bin_type=True)])
```

kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
```diff
@@ -0,0 +1,139 @@
...
+                return
+
+            self._seq += 1
+            self._socket.send_multipart(
+                [
+                    self._topic.encode("utf-8"),
+                    struct.pack(">Q", self._seq),
+                    payload,
+                ]
+            )
```
kfirtoledo: `time.time()` is captured before `self._send_lock`, so under concurrent `complete_store` calls the timestamp and sequence number can diverge: thread A captures T1, yields; thread B captures T2 > T1, grabs the lock first and sends with `seq=1, ts=T2`; thread A then sends `seq=2, ts=T1`. Result: timestamps appear non-monotonic relative to sequence. Move the timestamp into the critical section:

```suggestion
        with self._send_lock:
            if self._closed:
                return

            payload = msgpack.packb(
                [time.time(), packed_events], use_bin_type=True
            )
            self._seq += 1
            self._socket.send_multipart(
                [
                    self._topic.encode("utf-8"),
                    struct.pack(">Q", self._seq),
                    payload,
                ]
            )
```

kv_connectors/llmd_fs_backend/llmd_fs_backend/spec.py
albertoperdomo2: yep, I realized that while working on the PVC evitor events a few minutes ago. Thanks, I'll change that!

kv_connectors/llmd_fs_backend/tests/test_storage_events.py
```diff
@@ -0,0 +1,322 @@
...
+    import importlib.util
+
+    spec = importlib.util.spec_from_file_location(name, path)
+    mod = importlib.util.module_from_spec(spec)
+    sys.modules[name] = mod
+    spec.loader.exec_module(mod)
+    return mod
+
+
+def load_storage_event_modules():
```
kfirtoledo, line 85: Suggestion (non-blocking): the `load_storage_event_modules` / `temporary_modules` / `importlib.util.spec_from_file_location` machinery here exists because [`tests/conftest.py`](kv_connectors/llmd_fs_backend/tests/conftest.py) has a session-scoped `autouse=True` `require_cuda` fixture — so without `--noconftest`, the entire test session skips on a GPU-less box. The escape hatch works, but it's ~150 lines of plumbing, it stubs out real `vllm.*` imports (so interface drift won't surface), and the file is ~322 lines for what is fundamentally a small unit-test surface.

A cleaner shape, IMO:

1. Convert `require_cuda` in `conftest.py` from `autouse=True` to a marker — e.g. `@pytest.mark.requires_cuda`, with conftest only skipping tests carrying that marker (and the existing CUDA-heavy tests in `test_fs_backend.py` get the marker).
2. Then this file can drop all the module stubbing, `import` the publisher and managers normally, and use `monkeypatch.setattr(event_publisher_module.zmq, "Context", FakeZMQContext)` for the one place ZMQ actually matters.

Net effect: same coverage, ~80 lines instead of 322, real imports, and the event tests run anywhere with no `pytest --noconftest` workaround.

The gaps I'd close while you're in there: `_create_event_publisher` factory paths (`enable_events=False` → None, missing endpoint → None, `backend="OBJ"` → `OBJECT_STORE` medium, bind failure → fail-open None), and a tiny threaded test that hammers `publish_blocks_stored` to confirm `_send_lock`/`_seq` actually do their job.

kfirtoledo (commented):
Follow-up suggestion on the test structure.

kv_connectors/llmd_fs_backend/tests/test_storage_events.py
albertoperdomo2: I have actually a WIP [PR](https://github.com/llm-d/llm-d-kv-cache/pull/578) exactly for this. It is not the exact same implementation but the end shape is the same. Let me know if I should include that PR here directly, or if we can merge that PR first maybe.

kfirtoledo (commented):
Combining a few more findings from a second pass with my earlier draft comments.

kv_connectors/llmd_fs_backend/llmd_fs_backend/manager.py
```diff
@@ -34,8 +34,63 @@ class SharedStorageOffloadingManager(OffloadingManager):
     SharedStorageOffloadingManager manages KV offloading to a shared storage medium.
     """
 
-    def __init__(self, file_mapper: FileMapper) -> None:
+    def __init__(
+        self,
+        file_mapper: FileMapper,
+        model_name: str = "",
```
kfirtoledo: no need to pass `model_name` — take it from the `file_mapper`.

kv_connectors/llmd_fs_backend/tests/test_storage_events.py
```diff
@@ -0,0 +1,322 @@
...
+
+def _publisher_with_fake_zmq(monkeypatch):
+    ctx = FakeZMQContext()
+    monkeypatch.setattr(event_publisher_module.zmq, "Context", lambda: ctx)
+
+    publisher = StorageEventPublisher("tcp://*:5559", "test-model")
+    return publisher, ctx
+
+
+def test_hash_to_uint64_matches_file_mapper_lower_64_bits():
```
kfirtoledo, line 220: Can you add a comment on each test describing what it covers?

kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
```diff
@@ -0,0 +1,139 @@
...
+        if not hashes:
+            return
+
+        events: list[bytes] = []
+        for h in hashes:
+            event = [
+                "BlockStored",  # [0] tag
+                [h],  # [1] block_hashes (one per file)
+                0,  # [2] parent_hash (unused)
+                [],  # [3] token_ids (empty)
```
kfirtoledo: With `token_ids=[]`, the Go [`Pool.processEventBatch`](pkg/kvevents/pool.go#L351-L356) hits the `no request keys produced, skipping` branch — `TokensToKVBlockKeys` returns no keys when there are no tokens, so the event never reaches the index. Net effect today: events flow on the wire but nothing gets indexed. Either tokens need to flow through here, or call this out as a known limitation pending a follow-up storage-tier path that indexes by `block_hash` alone.

kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
```diff
@@ -0,0 +1,139 @@
...
+
+        Args:
+            endpoint: ZMQ bind address (e.g. ``tcp://*:5559``).
+            model_name: Model identifier included in the topic string.
+            medium: Storage backend type embedded in the topic and each event.
+        """
+        self._ctx = zmq.Context()
+        self._socket = self._ctx.socket(zmq.PUB)
+        self._socket.setsockopt(zmq.LINGER, 0)
+        self._socket.setsockopt(zmq.SNDHWM, 1000)
```
kfirtoledo: vLLM's reference `ZmqEventPublisher` (`vllm/distributed/kv_events.py`) uses `hwm=100_000` and also exposes a replay ROUTER socket + bounded queue, all configurable via `KVEventsConfig`. At 1000, events will get silently dropped under any sustained load — and without a drop counter, you'd never know. Suggest bumping to match vLLM's default and exposing it via `extra_config` (e.g. `storage_events_hwm`).

kv_connectors/llmd_fs_backend/llmd_fs_backend/manager.py
```diff
@@ -34,8 +34,63 @@ class SharedStorageOffloadingManager(OffloadingManager):
...
+                if backend == "OBJ"
+                else StorageMedium.SHARED_STORAGE
+            )
+
+            return StorageEventPublisher(
+                endpoint=endpoint,
+                model_name=model_name,
+                medium=medium,
+            )
+        except Exception:
```
kfirtoledo: Narrow to `zmq.ZMQError`. The catch-all swallows programming errors (typos, bad enum values) which then just get logged and silently disable the publisher. The realistic failure mode here is `zmq.ZMQError` (port in use, bad endpoint) — catch that explicitly.

kv_connectors/llmd_fs_backend/llmd_fs_backend/event_publisher.py
albertoperdomo2: I think this is expected, since the current GPU path requires tokens to compute request keys but storage events don't have tokens because they operate at the block hash level. The storage indexing path does not exist today, but that is why this PR is foundational to enable it.

kfirtoledo (approved):

Issue #573, opened 2026-05-06
Tool calls are dropped before tokenization which breaks KV cache block indexing


shashwatj07:
My scenario
https://gist.github.com/shashwatj07/55c00e0b1de3adc5d8f498b054a6f5a0

Script to reproduce
```
m = [
    {
      "role": "system",
      "content": "You are a helpful assistant." + "hi " * 1000,
    },
    {
      "role": "user",
      "content": "Issue Description.",
    },
    {
      "content": "Reflection.",
      "role": "assistant",
      "tool_calls": [
        {
          "function": {
            "arguments": "{\"command\": \"ls -la\"}",
            "name": "bash"
          },
          "id": "chatcmpl-tool-81320909e20aa185",
          "type": "function"
        }
      ]
    }
]

import json
from pyexpat.errors import messages
from litellm import completion
MODEL = "huggingface/Qwen/Qwen3-30B-A3B-Instruct-2507"
API_BASE = "http://localhost:32511/v1" # Use your LLM-d endpoint

response = completion(
    model=MODEL,
    api_base=API_BASE,
    messages=m[0:2],
    # max_tokens=1,
    temperature=0
)

print(response.get("usage", {}))
response = completion(
    model=MODEL,
    api_base=API_BASE,
    messages=m,
    # max_tokens=1,
    temperature=0
)
print(response.get("usage", {}))
```

Observe the anomaly:
```
grep -E 'prompt_token_ids'  modelserving_pods.log > vllm.txt
grep -E '\"tokens\"' epp_pods.log > epp.txt
```

```
from transformers import AutoTokenizer

model_name = "Qwen/Qwen3-30B-A3B-Instruct-2507"
tokenizer = AutoTokenizer.from_pretrained(
    model_name,
    trust_remote_code=True
)
print(tokenizer.decode(epp_tokens_int).replace('\n', '\\n'))
print(tokenizer.decode(vllm_tokens_int).replace('\n', '\\n'))
```

I have a sample here: https://www.diffchecker.com/xwBgEXXF/

PR #577, opened 2026-05-06
fix(tokenization): preserve assistant tool calls in UDS rendering

yankay:
## Summary
- Preserve assistant `tool_calls` across the UDS tokenizer render path.
- Restore `tool_calls_json` before vLLM chat template rendering.
- Add focused Go and Python regression coverage.

Addresses the assistant `tool_calls` path from https://github.com/llm-d/llm-d-kv-cache/issues/573.

## Test plan
- `git diff --check`
- `go test -mod=readonly ./pkg/tokenization`
- `services/uds_tokenizer/.venv/bin/python -m pytest services/uds_tokenizer/tests/test_renderer.py::TestRenderChatCompletion::test_render_restores_tool_calls_json -q`
- `make lint`

## Follow-up
- `role=tool` messages with `tool_call_id`/`name` still need separate coverage.


shashwatj07:
I checked after pulling the changes. Doesn't seem to fix the problem.

vMaroon:
> I checked after pulling the changes. Doesn't seem to fix the problem.

The problem is at the scheduler's level @yankay - this information won't be passed to the tokenizer in the first place, because it is not parsed. Can you handle there and complete the pipeline?

yankay:
HI @vMaroon 

Thanks — agreed. Scheduler-side counterpart: [llm-d/llm-d-inference-scheduler#1046](https://github.com/llm-d/llm-d-inference-scheduler/pull/1046), which parses `tool_calls` and forwards them into `RenderChatRequest`.

This needs to wait for this PR to merge and v0.9.0 to be released, then llm-d/llm-d-inference-scheduler#1046 can be merged — that will close out the issue. shashwatj07 confirmed on #1046 that the two together resolve the mismatch.

yankay:
Thanks @vMaroon. I addressed the Copilot feedback in the latest push.

This PR is the kv-cache side of the `tool_calls` fix. The scheduler/router follow-up https://github.com/llm-d/llm-d-router/pull/1046 depends on this change being merged and released, since it needs the `ToolCalls` field on `Conversation`.

Could you please take another look ? If this looks good, a `/lgtm` or `/approve` would help unblock the scheduler-side PR.

vMaroon:
/lgtm
/approve

Thanks @yankay - though the UDS tokenization service is getting deprecated in favor of the vLLM backend in the router's `token-producer` plugin. Can you make sure it works on that path?

yankay:
@vMaroon 
Thanks! Confirmed in https://github.com/llm-d/llm-d-router/pull/1046: the router `token-producer` vLLM path now preserves `tool_calls` and was verified end-to-end.

Issue #572, opened 2026-05-06
Wire fs_backend Python tests into CI


yankay:
Python tests under `kv_connectors/llmd_fs_backend/tests/` are currently not executed by any CI workflow. `make unit-test` only runs Go tests, and no workflow invokes `pytest` against this directory. As a result, tests added by PRs (e.g. `test_storage_events.py` in #571) pass locally but get no regression protection in CI.

Additionally, `tests/conftest.py` has a session-level `require_cuda` fixture that would skip the whole suite on GPU-less runners.

**Proposal**: start with the zero-cost path — split out the CPU-safe subset (e.g. `test_storage_events.py`), exempt it from `require_cuda`, and run it on `ubuntu-latest` via a new `unit-test-fs-backend-cpu` make target. GPU-required tests can come later via self-hosted runners or nightly e2e.

albertoperdomo2:
@yankay I have it ready but waiting for #571 to be merged. Then, I'll update the branch and create the PR for this 👍🏽

larainema:
I'd like to work on this as a first contribution.

I have a local patch that:
1. marks tests/test_storage_events.py as CPU-safe
2. avoids requiring CUDA/vLLM for CPU-safe tests
3. adds a unit-test-fs-backend-cpu make target
4. wires that target into GitHub Actions on ubuntu-latest with Python 3.12

Local validation passes:
- make unit-test-fs-backend-cpu with Python 3.12: 7 passed
- ruff check on the changed Python files: passed

I'll open a PR once I push the branch.

kfirtoledo:
llmd-fs-backend is now the FS tier of vLLM's multi-tier offloading connector (TieringOffloadingSpec). All new features and support continue there.

PR #578, opened 2026-05-07
ci: Wire fs_backend Python tests into CI

albertoperdomo2:
## Summary

Adds CI coverage for the CPU-safe subset of the llm-d fs backend Python tests. The root `unit-test` target now also runs a dedicated `unit-test-cpu` target, which collects all tests under `kv_connectors/llmd_fs_backend/tests/cpu/`, a directory reserved for tests that do not require CUDA, vLLM, NIXL, or compiled storage extensions, and `kv_connectors/pvc_evictor/tests`. New CPU-safe tests can be added by dropping files into any of these directories.

The fs backend pytest harness now supports a `no_cuda_required` marker so CPU-safe tests can bypass the existing autouse CUDA setup/teardown while preserving CUDA requirements for the rest of the suite. Python test dependencies for this lightweight path are kept in `kv_connectors/llmd_fs_backend/tests/requirements-cpu.txt`, and CI installs them before running `make unit-test`.

## Test

Validated with:

```
make unit-test-cpu
```

## Related Issues

- #572


albertoperdomo2:
cc: @kfirtoledo this PR might help both the CI tests and the issue you mentioned in the storage events emission test file.

albertoperdomo2:
cc: @yankay

yankay:
Nice work! 
Only One ask : `make unit-test` now implicitly needs Python deps — Go-only devs will hit `ModuleNotFoundError` locally. Could we wrap pytest in a venv like `uds-tokenizer-install-deps` (Makefile:115-135) does? Then it stays self-contained and the CI `pip install` step goes away.

Thanks @albertoperdomo2!

Makefile
```diff
@@ -18,6 +18,11 @@ BUILDER := $(shell command -v buildah >/dev/null 2>&1 && echo buildah || echo $(
 UDS_TOKENIZER_IMAGE ?= llm-d-uds-tokenizer:e2e-test
 FS_BACKEND_NAME ?= llmd-fs-backend
 FS_BACKEND_DEV_IMG ?= $(IMAGE_TAG_BASE)/$(FS_BACKEND_NAME):$(DEV_VERSION)
+FS_BACKEND_DIR := kv_connectors/llmd_fs_backend
+PVC_EVICTOR_DIR := kv_connectors/pvc_evictor
```
kfirtoledo, line 22: **Why both a `cpu/` directory *and* a `no_cuda_required` marker?**

This PR expresses "CPU-safe" two ways: tests are moved into `tests/cpu/` (path-based selection for this target) *and* tagged `pytestmark = pytest.mark.no_cuda_required`, which is what actually bypasses the autouse CUDA fixture in `conftest.py`.

Since the marker is required regardless, the directory looks redundant — and the name is a bit confusing. Selecting by marker instead would let you drop the `cpu/` folder entirely and avoid the `parents[1] -> parents[2]` churn from moving the file one level deeper:

```suggestion
FS_BACKEND_CPU_TESTS ?= $(FS_BACKEND_DIR)/tests
```

…paired with `pytest -q -m no_cuda_required $(FS_BACKEND_CPU_TESTS)` in the `unit-test-fs-backend-cpu` recipe. Keeps a single source of truth: a test is CPU-safe because it's *marked*, not because of where it lives. Not blocking — just flagging before this becomes a convention.

Makefile
```diff
@@ -65,13 +70,31 @@ clang:
...
+		echo "Creating virtual environment in $(CPU_TEST_VENV_DIR)..."; \
+		$(PYTHON_EXE) -m venv $(CPU_TEST_VENV_DIR); \
+		echo "Upgrading pip..."; \
+		$(CPU_TEST_VENV_BIN)/pip install --upgrade pip > /dev/null; \
+	else \
+		echo "Virtual environment already exists"; \
+	fi
+	@$(CPU_TEST_VENV_BIN)/pip install -q -r $(FS_BACKEND_DIR)/tests/requirements-cpu.txt
+
+.PHONY: unit-test-cpu
```
kfirtoledo, line 93: **Target name `unit-test-fs-backend-cpu` carries the same `cpu` ambiguity.**

It reads like a CPU-specific *variant* of the fs-backend tests rather than "the subset that doesn't need a GPU." If you move to marker-based selection, a clearer name like `unit-test-fs-backend` (it's the only fs-backend Python unit target in CI) or `unit-test-fs-backend-nogpu` would be less confusing.

Note a rename touches 3 spots: the `.PHONY` (L92), the target itself (here), and the `unit-test` dependency (L72).

kv_connectors/llmd_fs_backend/tests/requirements-cpu.txt
```diff
@@ -0,0 +1,3 @@
+pytest
+msgpack
+pyzmq
```
kfirtoledo, line 3: Since you're on `requires-python >=3.12`, consider replacing this file with a **PEP 735 dependency group** in `pyproject.toml`, so the CPU test deps live next to the runtime `dependencies` (one source of truth, no drift with `msgpack`/`pyzmq`):

```toml
[dependency-groups]
test-cpu = ["pytest", "msgpack", "pyzmq"]
```

Then install without building the package (so no `torch`/`nixl`):

```makefile
pip install -q --group $(FS_BACKEND_DIR)/pyproject.toml:test-cpu
```

Verified locally: `--group` installs only these three and never triggers the package build, even with `torch` in `[project].dependencies` / `[build-system].requires`. Needs pip >= 25.1, which the existing `pip install --upgrade pip` step already guarantees. Non-blocking, but a cleaner long-term home than a separate requirements file.

kfirtoledo (changes requested):
Thanks for wiring up CI coverage for the CPU-safe fs-backend tests — verified locally that the `no_cuda_required` marker correctly lets the suite run without CUDA (13 passed) and CI is green. The mechanism works.

My main concern is that "CPU-safe" is currently expressed three different ways — a `tests/cpu/` directory, a `no_cuda_required` marker, and a separate `requirements-cpu.txt` — which is redundant and a bit confusing as a convention to build on. Since the marker is required regardless, I'd suggest consolidating:

- select by marker (`pytest -m no_cuda_required`) and drop the `tests/cpu/` directory + the `parents[1] -> parents[2]` file-move churn,
- reconsider the `unit-test-fs-backend-cpu` target name (reads like a CPU *variant* rather than "the GPU-free subset"),
- move the test deps into a PEP 735 `[dependency-groups]` entry in `pyproject.toml` instead of a separate `requirements-cpu.txt`.

Details inline. None are correctness issues, but worth settling before this pattern spreads.

Makefile
albertoperdomo2: Already addressed,  it was renamed to `unit-test-cpu`, which is broader and clearer.

Makefile
albertoperdomo2: `pytest -m no_cuda_required` over the whole `tests/` dir won't work. On CI without `torch` installed, `pytest` crashes at collection when it tries to import `test_fs_backend.py` (which does import `torch` at module level). We could either do lazy imports in the tests that do require `torch` and so on, or keep the split.

kv_connectors/llmd_fs_backend/tests/requirements-cpu.txt
albertoperdomo2: This depends highly on how we approach the test directories/structure. Now, with two target dirs for CPU tests, pinning the deps in llmd_fs_backend's pyproject.toml would be misleading.

kfirtoledo (approved):

PR #580, opened 2026-05-08
correct make target check-buildah to check-builder

### LINKED ISSUE #579: `check-buildah` make target not found
`make check-tools` throws error:

```sh
/home/dhritishikhar/repo/github/llm-d-kv-cache [git::main] [dhritishikhar@fedora] [19:11]
> make check-tools              
make: *** No rule to make target 'check-buildah', needed by 'check-tools'.  Stop.
```

DhritiShikhar:
fixes #579

### DIFF
--- a/.github/workflows/ci-lint.yaml
+++ b/.github/workflows/ci-lint.yaml
@@ -58,3 +58,6 @@ jobs:
 
       - name: Run C/C++/CUDA formatting check
         run: make clang
+
+      - name: Check all tools are installed
+        run: make -n check-tools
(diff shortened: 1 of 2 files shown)


yankay:
Nice catch! Verified — `check-builder` is the only defined target (Makefile:341), `check-buildah` doesn't exist.

/lgtm

Non-blocking follow-up idea: this slipped through because `check-tools` isn't wired into any CI job. A cheap guard would be adding `make -n check-tools` (dry-run, just resolves the dep graph, no tool install needed) to `ci-pr-checks`. Happy to send a follow-up PR if it sounds useful.

DhritiShikhar:
@yankay @hexfusion I have updated the PR to include make target `make -n check-tools` in CI PR check.

Can you please review again?

vMaroon (approved):
Thanks!

PR #582, opened 2026-05-11
fs_backend: migrate kv_offload API to vllm 0.20.2

kfirtoledo:
## Summary

- Migrate `llmd_fs_backend` to the new vllm 0.20 KV-offload API: `BlockHash` → `OffloadKey`, `req_context` added to `lookup` / `prepare_load` / `prepare_store`, `PrepareStoreOutput` fields renamed (`keys_to_store` / `evicted_keys`), `lookup` returns `bool`, and `GPULoadStoreSpec` is now constructed with the required `block_indices` arg.
- Bump pins: `torch==2.10.0 → 2.11.0`, `vllm==0.19.1 → 0.20.2`, package version `0.19 → 0.20`. README and `vllm-storage.yaml` deployment example updated. New `0.20+cu130` wheel added under `wheels/`.
- Scope: single-group only (no HMA). Engine C++ ABI unchanged — only Python is touched.

## Test plan
- [x] `make test` passes
- [x] Throughput e2e test with Llama 3.1 8B and 70B

### DIFF
--- a/kv_connectors/llmd_fs_backend/llmd_fs_backend/file_mapper.py
+++ b/kv_connectors/llmd_fs_backend/llmd_fs_backend/file_mapper.py
@@ -76,9 +76,9 @@ class FileMapper:
         """
         Return the file path for a KV block.
         The path is built using hash-based subdirectories to limit directory
-        fan-out, with the KV group index as a filename suffix so different
-        groups never collide on the same block hash:
-            <base>/<hhh>/<hh>/<hash>_g<group_idx>.bin
+        fan-out, with the KV group index baked into the second-level folder
+        so different groups never collide on the same block hash:
+            <base>/<hhh>/<hh>_g<group_idx>/<hash>.bin
 
         Args:
             key: OffloadKey identifying the KV-cache block.
@@ -89,4 +89,4 @@ class FileMapper:
         hash_hex = get_offload_block_hash(key).hex()
         group_idx = get_offload_group_idx(key)
         subfolder1, subfolder2 = hash_hex[:3], hash_hex[3:5]
-        return f"{self.base_path}/{subfolder1}/{subfolder2}/{hash_hex}_g{group_idx}.bin"
+        return f"{self.base_path}/{subfolder1}/{subfolder2}_g{group_idx}/{hash_hex}.bin"
--- a/kv_connectors/llmd_fs_backend/llmd_fs_backend/file_mapper.py
+++ b/kv_connectors/llmd_fs_backend/llmd_fs_backend/file_mapper.py
@@ -66,22 +66,18 @@ class FileMapper:
             f"/{dtype}"
         )
 
-    def get_file_name(self, block_hash: int | bytes) -> str:
+    def get_file_name(self, key: bytes) -> str:
         """
         Return the file path for a KV block.
         The path is built using hash-based subdirectories:
         <base>/<hhh>/<hh>/<hash>.bin, to limit directory fan-out.
 
         Args:
-            block_hash: Hash identifying the KV-cache block (int or bytes).
+            key: OffloadKey identifying the KV-cache block.
 
         Returns:
             Full file path for the given block.
         """
-        if isinstance(block_hash, bytes):  # convert bytes to int
-            block_hash = int.from_bytes(block_hash, "big")
-        assert isinstance(block_hash, int)
-
-        block_hash_hex = f"{block_hash & ((1 << 64) - 1):016x}"
-        subfolder1, subfolder2 = block_hash_hex[:3], block_hash_hex[3:5]
-        return f"{self.base_path}/{subfolder1}/{subfolder2}/{block_hash_hex}.bin"
+        key_hex = key.hex()
+        subfolder1, subfolder2 = key_hex[:3], key_hex[3:5]
+        return f"{self.base_path}/{subfolder1}/{subfolder2}/{key_hex}.bin"
--- a/kv_connectors/llmd_fs_backend/llmd_fs_backend/file_mapper.py
+++ b/kv_connectors/llmd_fs_backend/llmd_fs_backend/file_mapper.py
@@ -12,6 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from vllm.v1.kv_offload.abstract import (
+    OffloadKey,
+    get_offload_block_hash,
+    get_offload_group_idx,
+)
+
 
 class FileMapper:
     """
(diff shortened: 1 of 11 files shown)


kv_connectors/llmd_fs_backend/llmd_fs_backend/manager.py
```diff
@@ -16,11 +16,12 @@
 from collections.abc import Iterable
 
 from vllm.logger import init_logger
-from vllm.v1.core.kv_cache_utils import BlockHash
 from vllm.v1.kv_offload.abstract import (
```
orozery, line 19:
```suggestion
from vllm.v1.kv_offload.base import (
```

kv_connectors/llmd_fs_backend/tests/test_fs_backend.py
```diff
@@ -24,6 +24,7 @@
 import pytest
 import torch
 from vllm.v1.core.kv_cache_utils import BlockHash
+from vllm.v1.kv_offload.abstract import OffloadKey, make_offload_key
 from vllm.v1.kv_offload.mediums import GPULoadStoreSpec
```
orozery, line 28:
```suggestion
from vllm.v1.kv_offload.base import GPULoadStoreSpec
```

kv_connectors/llmd_fs_backend/tests/test_fs_backend.py
```diff
@@ -24,6 +24,7 @@
...
 from vllm.v1.kv_offload.spec import (
```
orozery, line 29:
```suggestion
from vllm.v1.kv_offload.base import (
```

kv_connectors/llmd_fs_backend/llmd_fs_backend/manager.py
```diff
@@ -16,11 +16,12 @@
 from collections.abc import Iterable
```
orozery, line 16: APIs moved to using `Collection` instead of `Iterable`.

kv_connectors/llmd_fs_backend/llmd_fs_backend/file_mapper.py
```diff
@@ -66,22 +66,18 @@ def __init__(
             f"/{dtype}"
         )
 
-    def get_file_name(self, block_hash: int | bytes) -> str:
+    def get_file_name(self, key: bytes) -> str:
```
orozery:
```suggestion
    def get_file_name(self, key: OffloadKey) -> str:
```

kv_connectors/llmd_fs_backend/llmd_fs_backend/file_mapper.py
```diff
@@ -66,22 +66,18 @@ def __init__(
...
             Full file path for the given block.
         """
-        if isinstance(block_hash, bytes):  # convert bytes to int
-            block_hash = int.from_bytes(block_hash, "big")
-        assert isinstance(block_hash, int)
-
-        block_hash_hex = f"{block_hash & ((1 << 64) - 1):016x}"
-        subfolder1, subfolder2 = block_hash_hex[:3], block_hash_hex[3:5]
-        return f"{self.base_path}/{subfolder1}/{subfolder2}/{block_hash_hex}.bin"
+        key_hex = key.hex()
```
orozery: You should use `get_offload_block_hash` and `get_offload_group_idx`.

kfirtoledo: Done

kv_connectors/llmd_fs_backend/llmd_fs_backend/file_mapper.py
kfirtoledo: Done in 46552c6.

kv_connectors/llmd_fs_backend/llmd_fs_backend/manager.py
kfirtoledo: In vllm 0.20.2 the abstract methods still use `Iterable[OffloadKey]` — `Collection` came in after 0.20.2. Will tighten when we bump.

kv_connectors/llmd_fs_backend/llmd_fs_backend/manager.py
kfirtoledo: `vllm.v1.kv_offload.base` exists on vllm main but not in 0.20.2 (the version this PR pins) — only `abstract.py` is present. Will switch to `base` when we bump.

kv_connectors/llmd_fs_backend/tests/test_fs_backend.py
kfirtoledo: Same as above — `kv_offload.base` is not in vllm 0.20.2. Will switch when we bump.

kv_connectors/llmd_fs_backend/tests/test_fs_backend.py
kfirtoledo: Same as above — `kv_offload.base` is not in vllm 0.20.2. Will switch when we bump.

kfirtoledo:
@orozery PTAL

yankay:
Nice migration overall, thanks @kfirtoledo!

One quick question on the wheels: `wheels/llmd_fs_connector-0.20+1cu130-...whl` looks like an accidental rename of the old `0.1.0` wheel — the sha matches the previous `0.1.0`, and a separate correct `0.20+cu130-...whl` was added alongside. Should this one be `git rm`'d? And while we're at it, is dropping `0.1.0` from the index intentional, or should it be kept for back-compat?

kfirtoledo:
Thanks @yankay, I removed the old wheel 0.1.0. I'd rather not keep too many wheels in the repo to avoid bloating its size, especially since we use GitHub Pages to fetch and install the wheel anyway.

kfirtoledo:
@orozery , @dannyharnik PTAL

Example to Folder:
`<root>/meta-llama_Meta-Llama-3.1-8B_7126659d40c9/
`
config.json:

```
{
  "dtype": "torch.bfloat16",
  "gpu_block_size": 16,
  "gpu_blocks_per_file": 16,
  "inference_engine": "vllm",
  "kv_cache_groups": [
    {
      "block_size": 16,
      "num_layers": 32,
      "spec_type": "FullAttentionSpec"
    }
  ],
  "model_name": "meta-llama/Meta-Llama-3.1-8B",
  "pcp_size": 1,
  "pp_size": 1,
  "rank": 0,
  "tp_size": 1
```
}
Sample block file:
```
<root>/meta-llama_Meta-Llama-3.1-8B_7126659d40c9/012/34_g0/0123456789abcdef.bin
                                                  └┬─┘ └┬┘  └─────┬───────┘
                                                  hh[0:3] hh[3:5]+group  hash.bin
```

orozery:
Claude:
```
  decode_context_parallel_size (dcp) — DCP reuses GPUs within the TP group to shard KV across ranks during decode. If
   dcp > 1, each rank stores only a subset of the KV cache tokens. The PR does NOT include dcp_size in the hash. If two
  runs differ only in decode_context_parallel_size, the same rank would write different subsets of KV data into the same
   directory.
```

dannyharnik:
On second thought: why are we treating groups differently that TP Rank? We can use the same approach for both (have a config file per each one...)

orozery:
BTW @kfirtoledo if we want to merge fast we can defer the base path change to another PR (just add the group index part here + the api changes in 0.20.2)

kfirtoledo:
@orozery  you are right, I will split it to diffrent PR

kfirtoledo:
@orozery, @dannyharnik  PTAL

orozery (approved):

PR #585, opened 2026-05-14
fs_backend: collapse FileMapper path + add config.json audit

kfirtoledo:
## Summary

- Replace the deep `<model>/block_X.../tp_Y.../rank_Z/<dtype>/` layout with a flat `<model>_<sha256-12hex>/` folder fingerprinted by everything that affects the on-disk format: model, parallel sizes (tp/pp/pcp/dcp), dtype, block sizes, kv_cache_groups, inference_engine. Including `dcp_size` so two DCP runs don't collide.
- `rank` lives outside the hash. The base folder (with `config.json`) is shared across ranks; per-rank KV blocks land in a sibling `_r<rank>` folder that `get_file_name` appends at lookup time.
- New `FileMapper.write_run_config()` persists `self.fields` as JSON to `<base>/config.json` on demand (no-op if it exists), so each folder is self-describing.
- `FileMapper.from_vllm_config(...)` classmethod hides VllmConfig/KVCacheConfig extraction from `spec.py` while keeping the explicit `__init__` for tests.
- `parallel_agnostic=False` kwarg on `FileMapper.__init__` — when True, forces `tp/pp/pcp/dcp = 1` and `rank = 0` so multiple parallelism layouts share one cache folder (safe only if the offloaded format is parallelism-invariant).

## Test plan
- [x] `make test` passes (21 passed, 3 skipped)
- [x] Throughput e2e test with Llama 3.1 8B

### DIFF
--- a/kv_connectors/llmd_fs_backend/llmd_fs_backend/file_mapper.py
+++ b/kv_connectors/llmd_fs_backend/llmd_fs_backend/file_mapper.py
@@ -12,11 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import hashlib
+import json
+import os
+
+from vllm.config import VllmConfig
+from vllm.v1.kv_cache_interface import KVCacheConfig
 from vllm.v1.kv_offload.base import (
     OffloadKey,
     get_offload_block_hash,
     get_offload_group_idx,
 )
+
+_BASE_PATH_HASH_LEN = 12
+_CONFIG_FILENAME = "config.json"
 
 
 class FileMapper:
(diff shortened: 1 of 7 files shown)


kv_connectors/llmd_fs_backend/llmd_fs_backend/file_mapper.py
```diff
@@ -33,60 +42,103 @@ def __init__(
...
+        self.fields: dict = {
+            "model_name": model_name,
+            "gpu_block_size": gpu_block_size,
+            "gpu_blocks_per_file": gpu_blocks_per_file,
+            "tp_size": tp_size,
+            "pp_size": pp_size,
+            "pcp_size": pcp_size,
+            "dcp_size": dcp_size,
+            "dtype": str(dtype),
+            "kv_cache_groups": kv_cache_groups or [],
```
orozery, line 72: This is leaking vllm internal structs.
I think we should encode {[layer_names], block_size for group in groups].
Also, you need to change `gpu_block_size` to take it from vllm_config.cache_config.block_size.
It's actually `hash_block_size` (should be renamed).

natoscott:
```
Replace the deep <model>/block_X.../tp_Y.../rank_Z/<dtype>/ layout with a flat <model>_<sha256-12hex>/ folder [...]
```

@kfirtoledo  what's the reason for doing this OOC?  At first glance it would seem this may introduce contention on the single top-level directory when a clustered filesystem is being used across multiple hosts.

kv_connectors/llmd_fs_backend/llmd_fs_backend/file_mapper.py
kfirtoledo: @orozery check now if it is better

kfirtoledo:
@natoscott main motivation is to avoid the deep folder fan-out we had before (`<model>/block_size_X/tp_Y.../rank_R/<dtype>/`, ~5 nested levels mostly empty). Different workloads still land in different folders because the sha256 fingerprint covers every field that affects on-disk format — model, parallel sizes, dtype, block sizes, kv_cache_groups, inference_engine.

kv_connectors/llmd_fs_backend/llmd_fs_backend/file_mapper.py
```diff
@@ -28,65 +37,108 @@ def __init__(
...
+    ) -> "FileMapper":
+        """Build a FileMapper from a vllm VllmConfig + KVCacheConfig."""
+        parallel_config = vllm_config.parallel_config
+        dtype = str(vllm_config.cache_config.cache_dtype).replace("torch.", "")
+        kv_cache_groups = [
+            {
+                "num_layers": len(group.layer_names),
+                "block_size": group.kv_cache_spec.block_size,
+                "spec_type": type(group.kv_cache_spec).__name__,
+                "layer_names": list(group.layer_names),
```
orozery: Let's just have layer_names and block_size.
Layer types can be deduced from the model's config.json.

natoscott:
Excellent, thanks for the explanation @kfirtoledo

kfirtoledo:
@orozery, @dannyharnik  PTAL

orozery (approved):
LGTM.
@dannyharnik please have a look as well

dannyharnik (approved):
\LGTM

PR #587, opened 2026-05-15
Add SHA256-CBOR hashing algorithm for token processor with extra keys…

leipanhz:
Add KV-cache file prefetch plugin for inference requests (experimental feature)
Part 1: changes in KV-Cache (current PR)
Part 2: changes in llm-d-router

PR Description:
Introduces a new experimental feature that aims to proactively prefetch KV-cache blocks across different storage tiers before inference requests are processed by the GPU pod. The plugin extends the precise prefix cache scorer with engine key calculation to determine the storage location (file names) of KV-cache blocks that will be needed and arrange for them to be promoted to a closer storage tier to improve inference latency. The current implementation is intended for a shared file system that includes transparent access to a remote storage tier, such as IBM Storage Scale configured to off-load cold data to remote object storage. The prefetch plugin uses a concurrent worker thread pool architecture to efficiently prefetch multiple (configurable) files in parallel from remote storage to the shared file system. In a future version of the plugin this could be extended, for example, to prefetch KV-cache blocks from the file system to CPU memory on the worker node that the request is being routed to.

For this to work correctly, the plugin must be configured to use a hash algorithm for generating engine keys that matches the algorithm used by vLLM when offloading KV-cache blocks to storage. For this purpose, this work adds a configurable hashing algorithm SHA256-CBOR to the token processor as an alternative for vLLM compatibility. The SHA256-CBOR implementation supports extra keys (multimodal features) in block hash computation. In addition, this feature relies on logic derived from the llm-d-fs-connector to generate KV file names, so it currently only works with the llm-d-fs-connector.

Changes include:

New Prefetch Plugin (prefetch_prerequest_experimental.go):
- Implements PreRequest interface for pre-inference file prefetching
- Converts engine keys to filesystem paths using llm-d-fs-connector format
- Manages worker thread pool for concurrent file prefetching (configurable workers)
- Each worker reads configurable number of blocks (BlockSize x BlockCount bytes) from KV-cache files to trigger prefetch of the rest of the file from remote storage.
- Supports configurable prefetch parameters (block size, concurrency, queue size)

Precise Prefix Cache Scorer Enhancement (precise_prefix_cache.go):
- Add GetEngineKeysForRequest() method to extract engine keys from requests
- Support multimodal features in engine key computation

Add SHA256-CBOR hashing algorithm for token processor with extra keys support
- Add configuration to choose hashing function via the field name “hashAlgorithm”: FNV64a default, SHA256-CBOR for vLLM
- Implement SHA256-CBOR hashing matching vLLM engine-key computation
- Extend BlockExtraFeatures for multimodal content support

### DIFF
--- a/pkg/kvcache/indexer_test.go
+++ b/pkg/kvcache/indexer_test.go
@@ -46,6 +46,13 @@ func (m *mockTokenProcessor) TokensToKVBlockKeys(
 	return m.blockKeys, nil
 }
 
+func (m *mockTokenProcessor) TokensToKVBlockKeysWithDigests(
+	_ kvblock.BlockHash, tokens []uint32, _ string, _ []*kvblock.BlockExtraFeatures,
+) ([]kvblock.BlockHash, [][]byte, error) {
+	m.receivedTokens = tokens
+	return m.blockKeys, nil, nil
+}
+
 func (m *mockTokenProcessor) BlockSize() int {
 	return 16
 }
--- a/pkg/kvcache/kvblock/extra_keys.go
+++ b/pkg/kvcache/kvblock/extra_keys.go
@@ -29,8 +29,73 @@ type MMHash struct {
 
 // BlockExtraFeatures holds per-block extra data that taints the block hash.
 // A nil *BlockExtraFeatures means pure text (no taint).
+//
+// The fields below mirror the per-block extras vLLM concatenates inside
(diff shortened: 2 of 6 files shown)


leipanhz:
Related issue: https://github.com/llm-d/llm-d-router/issues/866

leipanhz:
keep-alive

PR #589, opened 2026-05-17
fix: remove MIN_STAGING_BUFFER_SIZE that inflates KV cache on disk

### LINKED ISSUE #454: Remove MIN_STAGING_BUFFER_SIZE, leads to way more data stored than required on smaller models
Why is a [MIN_STAGING_BUFFER_SIZE](https://github.com/llm-d/llm-d-kv-cache/blob/5fe9e4edbff5f5c7114584d544631c083466a02f/kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp#L35) needed? This seems to unnecessarily increase size on disk of the KV cache tokens, and therefore the time spent reading/writing the tokens as well. For a small model (such as llama 3.1 8b), this more than doubles the size (7mb -> 16mb). So instead of reading 4.38GB for 40k tokens, it ends up reading/writing 10GB to disk.

At a prefix (context) size of 40k tokens, this leads to significantly worse TTFT (6.5 seconds vs 2.8 seconds).

Used [here](https://github.com/llm-d/llm-d-kv-cache/blob/5fe9e4edbff5f5c7114584d544631c083466a02f/kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp#L230).

Jwrede:
## Summary
- Remove the 16 MB `MIN_STAGING_BUFFER_SIZE` floor in `allocate_staging_buffer`
- Staging buffers now use the exact size computed by `calc_staging_bytes` (block_size * blocks_per_file)
- For Llama 3.1 8B, this halves the per-file on-disk size (16 MB back to ~7 MB) and reduces TTFT from 6.5 s to 2.8 s at 40k-token prefix length (numbers from #454)

The floor was applied via `std::max(required_bytes, MIN_STAGING_BUFFER_SIZE)`, then `write_buffer_to_file` wrote `buf.size` bytes to disk, padding every file to at least 16 MB regardless of actual KV block size.

Fixes #454

## Test plan
- [x] Extension compiles and loads successfully on NVIDIA L4 (CUDA 12.4, torch 2.6.0)
- [ ] CI `test_fs_backend.py` (requires matching vllm version)

### DIFF
--- a/kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
+++ b/kv_connectors/llmd_fs_backend/csrc/storage/thread_pool.cpp
@@ -31,8 +31,6 @@
 #include "numa_utils.hpp"
 #include "logger.hpp"
 
-// Minimum staging buffer size: 16 MB
-const size_t MIN_STAGING_BUFFER_SIZE = 16 * 1024 * 1024;
 
 // Define static thread-local members
 thread_local StagingBufferInfo ThreadPool::m_staging_buffer{};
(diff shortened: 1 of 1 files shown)


Jwrede:
Friendly ping, this has been open for two weeks. Would appreciate a review when someone has a chance. Happy to address any feedback. @kfirtoledo @vMaroon

kfirtoledo (approved):

PR #590, opened 2026-05-18
Fix: Avoid OBJ backend runtime failures after vLLM 0.21 KV-offload migration

weizhoublue:
Trigger Scenario
  When llmd_fs_backend is configured with backend == "OBJ", SharedStorageOffloadingSpec.get_manager() returns NixlStorageOffloadingManager. vLLM 0.21 then calls lookup() and complete_store() using the new KV-offload API signatures.

  NixlStorageOffloadingManager still uses the old method signatures:
  lookup(block_hashes: Iterable[BlockHash]) -> int
  complete_store(block_hashes: Iterable[BlockHash], success: bool = True)

  This can cause runtime argument mismatch errors, incorrect lookup behavior, or failed tracking of stored keys in dict lookup mode.

 Root Cause
  PR #588 migrated SharedStorageOffloadingManager to the vLLM 0.21 KV-offload API, but missed the overridden methods in NixlStorageOffloadingManager. Because the NIXL subclass overrides the fixed parent methods, OBJ backend calls still hit the stale vLLM 0.20-style implementation.

### DIFF
--- a/kv_connectors/llmd_fs_backend/llmd_nixl/manager.py
+++ b/kv_connectors/llmd_fs_backend/llmd_nixl/manager.py
@@ -12,10 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from collections.abc import Iterable
+from collections.abc import Collection
 
 from vllm.logger import init_logger
-from vllm.v1.core.kv_cache_utils import BlockHash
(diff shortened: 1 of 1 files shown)


kfirtoledo:
looks good, @effi-ofer  can you review it also

kfirtoledo (approved):

PR #596, opened 2026-05-20
build(fs_connector): tag wheel manylinux_2_28 for UBI9/RHEL9 compatibility

### LINKED ISSUE #595: llmd_fs_connector wheel requires manylinux_2_35 (glibc 2.35) but llm-d-cuda runtime image uses UBI9 (glibc 2.34)
## Summary

The `llmd_fs_connector` wheel published in v0.8.0 is tagged `manylinux_2_35`, which requires glibc ≥ 2.35. The `llm-d-cuda` runtime image is based on RHEL UBI 9, which ships glibc **2.34**. This causes `pip install` of the wheel to fail at runtime with:

```
ERROR: llmd_fs_connector-0.19+cu130-cp312-cp312-manylinux_2_35_x86_64.whl is not a supported wheel on this platform.
```

## Reproduction

Confirmed on `ghcr.io/llm-d/llm-d-cuda:v0.7.0`:

```
$ ldd --version
ldd (GNU libc) 2.34

$ pip3.12 install llmd_fs_connector-0.19+cu130-cp312-cp312-manylinux_2_35_x86_64.whl
ERROR: llmd_fs_connector-0.19+cu130-cp312-cp312-manylinux_2_35_x86_64.whl is not a supported wheel on this platform.
```

## Root cause

`kv_connectors/llmd_fs_backend/Dockerfile.wheel` builds on `ubuntu22.04` (glibc 2.35) and stamps the wheel with `--plat manylinux_2_35_$(uname -m)` via `auditwheel repair`. However:

- The **runtime container** (`llm-d-cuda`) is UBI9-based (glibc 2.34), not Ubuntu 22.04
- `manylinux_2_35` requires glibc ≥ 2.35 — one minor version above what UBI9 provides
- vLLM itself is installed at **image build time** and thus bypasses the pip platform check; the fs_connector is installed at **runtime** via pip and hits this check

Note: vLLM 0.19.1's cu130 wheel is also tagged `manylinux_2_35`, but the Dockerfile.wheel comment "matches vLLM" refers to the wheel tag convention, not runtime compatibility.

## Immediate fix

Change `kv_connectors/llmd_fs_backend/Dockerfile.wheel` line 81:

```dockerfile
# Before
--plat manylinux_2_35_$(uname -m) \

# After
--plat manylinux_2_28_$(uname -m) \
```

`manylinux_2_28` requires glibc ≥ 2.28, which glibc 2.34 (UBI9) satisfies. If `auditwheel` rejects the tag due to actual glibc 2.35 symbol usage in the binary, the build base should also be changed from `ubuntu22.04` to a glibc 2.28-compatible image (e.g. `ubi8`-based CUDA image).

A PR with this fix follows.

## Design consideration: runtime pip install vs baking into the image

This bug highlights a deeper design question. The fs_connector is currently installed at pod startup via a `VLLM_PRE_CMD` (`pip install` before vLLM launches). This approach:

- **Requires network access** at every pod startup
- **Adds latency** to server restarts (pip install + dependency resolution on the critical path)
- **Exposes platform compatibility issues** at runtime — as seen here
- **Differs from vLLM** itself, which is baked into the `llm-d-cuda` image at build time and therefore never hits a runtime pip platform check

The alternative — including `llmd_fs_connector` in the `llm-d-cuda` image at build time — would eliminate all of these issues, since the build environment (Ubuntu 22.04) satisfies `manylinux_2_35` and the wheel installs cleanly there. The tradeoff is that connector updates require a new base image build rather than just staging a new wheel.

For production deployments on RHEL/UBI-based infrastructure, baking the connector into the image is likely the more robust path. The manylinux tag fix (this PR) unblocks the current runtime-install approach, but the design question is worth discussing.

natoscott:
Fixes #595.

The `llmd_fs_connector` wheel was tagged `manylinux_2_35` (requires glibc ≥ 2.35), but the `llm-d-cuda` runtime image is RHEL UBI 9-based (glibc 2.34). This caused pip to reject the wheel when installed at runtime:

```
ERROR: llmd_fs_connector-0.19+cu130-cp312-cp312-manylinux_2_35_x86_64.whl
       is not a supported wheel on this platform.
```

This blocks all filesystem-offload configurations that install the wheel via `VLLM_PRE_CMD` at pod startup.

`kv_connectors/llmd_fs_backend/Dockerfile.wheel`: change `auditwheel repair --plat` from `manylinux_2_35` to `manylinux_2_28`.

`manylinux_2_28` requires glibc ≥ 2.28, which UBI9's glibc 2.34 satisfies. The build base (Ubuntu 22.04) is unchanged — the `--only-plat` flag stamps the specified tag without auditwheel attempting its own glibc symbol detection.

See #595 for discussion of the broader design question around runtime pip install vs baking the connector into the `llm-d-cuda` image.

### DIFF
--- a/kv_connectors/llmd_fs_backend/Dockerfile.wheel
+++ b/kv_connectors/llmd_fs_backend/Dockerfile.wheel
@@ -1,8 +1,9 @@
 # syntax=docker/dockerfile:1
 
-# Ubuntu 22.04 (glibc 2.35) = manylinux_2_35 baseline; multi-arch; mirrors
-# vLLM's published wheel naming. No PPA / no api.launchpad.net dependency
-# because Python is installed via uv (hermetic standalone interpreter).
+# Ubuntu 22.04 (glibc 2.35) build base; wheel is tagged manylinux_2_28 for
+# compatibility with the llm-d-cuda runtime image (RHEL UBI 9, glibc 2.34).
+# No PPA / no api.launchpad.net dependency because Python is installed via
+# uv (hermetic standalone interpreter).
 ARG CUDA_BASE_TAG=12.8.1-devel-ubuntu22.04
 FROM nvcr.io/nvidia/cuda:${CUDA_BASE_TAG} AS build
 
@@ -73,12 +74,14 @@ RUN rm -rf build dist-raw dist *.egg-info && \
     mkdir -p dist-raw dist && \
     python setup.py bdist_wheel -d dist-raw
 
-# auditwheel: stamp manylinux_2_35_<arch> (matches vLLM) and bundle small
-# non-system deps. torch + CUDA libs excluded — they come from the user's
-# torch install; bundling would bloat the wheel and conflict at runtime.
+# auditwheel: stamp manylinux_2_28_<arch> so the wheel is pip-installable on
+# RHEL UBI 9 (glibc 2.34) which is the llm-d-cuda runtime base image.
+# manylinux_2_28 requires glibc >= 2.28 — satisfied by UBI9's glibc 2.34.
+# torch + CUDA libs excluded — they come from the user's torch install;
+# bundling would bloat the wheel and conflict at runtime.
 RUN auditwheel show dist-raw/*.whl && \
     auditwheel repair \
-      --plat manylinux_2_35_$(uname -m) \
+      --plat manylinux_2_28_$(uname -m) \
       --only-plat \
       --exclude 'libtorch*.so*' \
       --exclude 'libc10*.so*' \


natoscott:
Closing this — on reflection we were solving the wrong problem.

We discovered that `llmd_fs_connector` is already baked into `ghcr.io/llm-d/llm-d-cuda:v0.7.0` at image build time via `docker/scripts/cuda/runtime/install-offloading-connector.sh`, using the `linux_x86_64` wheels committed in `kv_connectors/llmd_fs_backend/wheels/`. Our VLLM_PRE_CMD runtime pip install was unnecessary and the wrong approach — the guides make this clear in retrospect.

The `manylinux_2_35` releases page wheels may be a separate concern for users building custom images, but we're not certain enough to propose a fix. Apologies for the noise.

PR #601, opened 2026-05-21
pvc_evictor: support updated fs_backend layout (FileMapper v0.20+)

guygir:
Updates the PVC evictor crawler for the flat llmd_fs_backend layout from #585 .

- Path only discovery (no FileMapper / vLLM import in the image)
- Small Dockerfile (evictor code only, no need for bundled llmd_fs_backend)
- Pytest coverage for traversal and sharding
- ci-pvc-evictor workflow runs tests on pvc_evictor/** changes

Tested:

- [x] make test
- [x] Cluster: discovery verified with filemapper-v0.21 image.


kfirtoledo:
cc @albertoperdomo2,

kv_connectors/pvc_evictor/processes/crawler.py
```diff
@@ -271,13 +172,18 @@ def crawler_process(
     max_queue_size = config_dict["file_queue_maxsize"]
     access_time_threshold_seconds = config_dict["file_access_time_threshold_minutes"] * MINUTES_TO_SECONDS
 
+    if not cache_path.exists():
+        logger.error(
+            f"Crawler P{process_num} cache_path does not exist: {cache_path}"
+        )
+        return
+
```
albertoperdomo2, line 180: In `stream_cache_files()` we already check `if not cache_path.exists():`, which logs a warning and returns. With this check here, in the `crawler_process()`, each crawler exits immediately when `cache_path` is missing, which I think is dangerous because the parent process does not restart dead crawlers. 

So if the evictor starts before the fs-backend cache directory exists, the pod stays up but all crawlers are gone permanently, and file discovery never begins even if the directory appears later, which is also misleading.

guygir:
I’m going to close this PR in favor of #611, which is the smaller focused implementation for the same FileMapper layout fix. Thanks for the review @albertoperdomo2, I agree. #611 keeps the missing path handling inside the stream loop, so crawlers keep retrying instead of exiting.

I’ll split the non-overlapping pieces from this PR (tests/CI/docs, dockerfile cleanup if still needed, etc) into a follow up after #611 lands.

PR #603, opened 2026-05-23
fix(sglang): decode token_ids as typed []uint32 with bigram support

ruixian94:
## Summary

`SGLangAdapter` currently fails to decode `BlockStored` events when the underlying SGLang engine has EAGLE-family speculative decoding enabled (`--speculative-algorithm EAGLE | EAGLE3 | FROZEN_KV_MTP`). In that mode SGLang emits `token_ids` as a **nested array of bigram pairs** rather than a flat int list, which the existing `[]uint32` msgpack target cannot satisfy — every BlockStored event from such pods is dropped with a decode error.

This PR teaches `sglang_adapter.go` to handle both wire shapes via a typed `msgpack.CustomDecoder`, with no `[]any` boxing on the hot path.

### The two shapes

SGLang's `mem_cache/events.py:_record_store_event` materialises `token_ids` differently depending on whether the request's `RadixKey` is in bigram mode:

| Mode | Wire payload | Trigger |
|---|---|---|
| Flat | `[t0, t1, t2, ...]` | normal inference |
| Bigram | `[[t0, t1], [t1, t2], [t2, t3], ...]` | `is_bigram = is_eagle == True`, i.e. EAGLE / EAGLE3 / FROZEN_KV_MTP |

`is_bigram` is set in `radix_cache.py`, `hiradix_cache.py`, `unified_radix_cache.py`, and `swa_radix_cache.py` whenever the cache was constructed with `is_eagle=True`. The tuple form has been the wire contract since `events.py` was introduced in [sgl-project/sglang#23678](https://github.com/sgl-project/sglang/pull/23678) (the in-code comment explicitly notes it preserves the historical EAGLE event payload after [#23106](https://github.com/sgl-project/sglang/pull/23106) made bigram an O(1) flag).

Pairs overlap on the boundary token, so collapsing each pair to its second element gives back the engine's `raw[1:]` token sequence — exactly the prefix the indexer needs.

### The fix

A new `sglangTokenIDs` (alias of `[]uint32`) implements `msgpack.CustomDecoder`:

1. `DecodeArrayLen` to read the outer count.
2. `PeekCode` on the first element — `FixArray | Array16 | Array32` ⇒ bigram, otherwise flat.
3. Bigram branch: per pair, `Skip()` the previous token and `DecodeUint32()` the current; defensive `Skip()` for any future trailing fields.
4. Flat branch: straight `DecodeUint32()` per element.

This eliminates the prior `[]any` + `flattenTokenIDs([]any)` + `toUint32(any)` boxing chain (three layers of reflect / type-switch per token) and replaces it with a single typed allocation per event. The dead helpers go away.

## Test plan

- [x] New `TestSGLangBlockStored_BigramTokenIDs` round-trips `[[10,11],[11,12],[12,13]]` and asserts the decoder yields `[11, 12, 13]` (= `raw[1:]`).
- [x] All existing `TestSGLang*` cases (FullFields / 7Fields / MinimalFields / TooFewFields / BlockRemoved / AllBlocksCleared / UnknownTag / ShardingKey / ParseMessage) remain green — flat-mode wire is unchanged.
- [x] `go test ./pkg/kvevents/engineadapter/ -count=1` passes.
- [x] `go vet ./pkg/kvevents/engineadapter/...` clean.

## Notes for reviewers

- No config flag was added. Shape detection is purely content-driven on the wire (the two msgpack codes are disjoint), so it works for mixed-pod pools where some pods run EAGLE and others don't — no operator-side knob to misconfigure.
- The `BlockStored.token_ids: list[int]` type annotation in SGLang's `kv_events.py` is misleading; the bigram payload violates it. That mismatch is what led the original adapter (PR #443) — modelled on the "SGLang ≈ vLLM wire format" assumption — to type the field as `[]uint32`. PR #484 fixed an analogous issue on the vLLM side; this is the SGLang counterpart it explicitly deferred.

### DIFF
--- a/pkg/kvevents/engineadapter/sglang_adapter.go
+++ b/pkg/kvevents/engineadapter/sglang_adapter.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	"github.com/vmihailenco/msgpack/v5"
+	"github.com/vmihailenco/msgpack/v5/msgpcode"
 
 	"github.com/llm-d/llm-d-kv-cache/pkg/kvevents"
 )
@@ -107,12 +108,79 @@ type msgpackSGLangBlockStoredEvent struct {
 	Tag             string
 	BlockHashes     []any
 	ParentBlockHash any
-	TokenIds        []uint32
+	TokenIds        sglangTokenIDs
 	BlockSize       int
 	LoraID          *int    `msgpack:",omitempty"`
 	Medium          *string `msgpack:",omitempty"`
 	LoraName        *string `msgpack:",omitempty"`
 	ExtraKeys       []any   `msgpack:",omitempty"`
+}
+
+// sglangTokenIDs decodes BlockStored.token_ids, which SGLang emits in two
+// shapes: flat [t0, t1, ...] in normal mode, or bigram [[t0,t1],[t1,t2],...]
+// when EAGLE-family speculative decoding is enabled. The wire codes for the
+// two shapes are disjoint (int vs array), so the decoder picks the branch
+// from the first element. The bigram branch keeps the second element of each
+// pair, which collapses overlapping pairs back to a unique token sequence
+// (equivalent to the engine's raw[1:]).
+//
+// See sglang/python/sglang/srt/mem_cache/events.py:_record_store_event and
+// the is_bigram flag set from is_eagle in radix_cache.py / hiradix_cache.py.
+type sglangTokenIDs []uint32
+
+func (t *sglangTokenIDs) DecodeMsgpack(dec *msgpack.Decoder) error {
+	n, err := dec.DecodeArrayLen()
+	if err != nil {
+		return err
+	}
+	if n <= 0 {
+		*t = nil
+		return nil
+	}
+
+	code, err := dec.PeekCode()
+	if err != nil {
+		return err
+	}
+	isBigram := msgpcode.IsFixedArray(code) ||
+		code == msgpcode.Array16 ||
+		code == msgpcode.Array32
+
+	out := make([]uint32, n)
+	if isBigram {
+		for i := 0; i < n; i++ {
+			inner, err := dec.DecodeArrayLen()
+			if err != nil {
+				return fmt.Errorf("token_ids bigram[%d]: %w", i, err)
+			}
+			if inner < 2 {
+				return fmt.Errorf("token_ids bigram[%d]: pair too short, len=%d", i, inner)
+			}
+			if err := dec.Skip(); err != nil { // prev token (overlaps with previous pair)
+				return fmt.Errorf("token_ids bigram[%d][0]: %w", i, err)
+			}
+			v, err := dec.DecodeUint32()
+			if err != nil {
+				return fmt.Errorf("token_ids bigram[%d][1]: %w", i, err)
+			}
+			for k := 2; k < inner; k++ {
+				if err := dec.Skip(); err != nil {
+					return fmt.Errorf("token_ids bigram[%d][%d]: %w", i, k, err)
+				}
+			}
+			out[i] = v
+		}
+	} else {
+		for i := 0; i < n; i++ {
+			v, err := dec.DecodeUint32()
+			if err != nil {
+				return fmt.Errorf("token_ids[%d]: %w", i, err)
+			}
+			out[i] = v
+		}
+	}
+	*t = out
+	return nil
 }
 
 type msgpackSGLangBlockRemovedEvent struct {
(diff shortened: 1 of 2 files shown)


vMaroon:
@bongwoobak can you give this a look?

pkg/kvevents/engineadapter/sglang_adapter.go
```diff
@@ -107,14 +108,99 @@ type msgpackSGLangBlockStoredEvent struct {
...
+		}
+	}
+	*t = out
+	return nil
+}
+
+// decodeBigramTokenIDs decodes the bigram-shaped token_ids array
+// ([[t0,t1],[t1,t2],...]) into out by keeping the second element of each pair,
+// which collapses overlapping pairs back to the underlying token sequence.
+func decodeBigramTokenIDs(dec *msgpack.Decoder, out []uint32) error {
```
bongwoobak, line 173: Decode succeeds, but EAGLE prefix matching still won't work.

llm-d's hashing side has no bigram awareness (token_processor.go:hash() is plain FNV/CBOR over the chunk, grep "EAGLE|bigram" pkg/ returns 0).

So with this PR, EAGLE pods store hash(raw[1:17]) while the router hashes incoming requests as hash(raw[0:16]). same prompt, different hashes between EAGLE and non-EAGLE pods.

llm-d's hashing side needs a paired path (shift-by-1 or pair-hash)

pkg/kvevents/engineadapter/sglang_adapter_test.go
```diff
@@ -183,6 +183,42 @@ func TestSGLangBlockStored_MinimalFields(t *testing.T) {
 	assert.Nil(t, blockStored.ExtraKeys)
 }
 
+// TestSGLangBlockStored_BigramTokenIDs verifies the decoder collapses the
+// nested [[prev, curr], ...] payload that SGLang emits when EAGLE-family
+// speculative decoding is enabled. Each pair's second element is kept, which
+// matches the engine's raw[1:] token sequence.
+func TestSGLangBlockStored_BigramTokenIDs(t *testing.T) {
```
bongwoobak, line 189: Could you add tests for the new error/skip branches? Specifically: inner < 2 (malformed pair), inner > 2 (the defensive Skip loop), and empty token_ids.

bongwoobak (changes requested):
Since SGLang stores KV cache differently when EAGLE-family speculative decoding is enabled, this fix is needed. That said, I've left a few concerns that I think should be addressed in this PR.

Bottom line: with this PR alone, SGLang + EAGLE/MTP pods won't produce any prefix cache hits unless the request-key side is also updated.

ryanx-sir:
Thanks for the careful read @bongwoobak — you were right that the first revision broke prefix matching for EAGLE pods. I pushed `2b9332f` which addresses both comments.

**On "shift-by-1 vs pair-hash":** I went with shift-by-1, but kept it adapter-local instead of touching `token_processor.go`. The reasoning is that SGLang's bigram payload of N pairs covers N+1 raw tokens (`raw[0:N+1]`), and the previous decoder was dropping `pair[0][0]`, emitting only `raw[1:N+1]`. The fix is to decode the page-head too — for `i == 0` we now keep both halves of the pair, for `i > 0` we keep only the tail (which is what the prior code did for every pair).

The trailing `pair[N-1][1]` is the overlap token shared with the next page; `kvblock.chunkTokens` already drops it as a partial block (`// no partial blocks`), so the recomputed canonical block hashes line up exactly with what the request side produces for the same flat prompt. No bigram awareness is needed in `token_processor.go`, EAGLE and non-EAGLE pods produce identical hashes for the same prefix, and they can coexist in the same `InferencePool`.

I considered the pair-hash route, but it would require teaching `token_processor.go` two hashing modes, plumbing per-pool engine identity to the request path, and either splitting EAGLE/non-EAGLE pods into separate pools or doing dual lookups per request — much wider blast radius than this one-engine quirk warrants. Happy to revisit if you'd prefer the hashing-side variant.

**On tests:** added three tests for the branches you flagged — `TestSGLangBlockStored_BigramMalformedPair` (`inner < 2`), `TestSGLangBlockStored_BigramOversizedPair` (`inner > 2` defensive skip), and `TestSGLangBlockStored_EmptyTokenIDs`. The existing bigram test was updated to assert the full `raw[0:N+1]` sequence.

bongwoobak (approved):

ryanx-sir:
Thanks @bongwoobak for the thorough review and approval on `2b9332f`. The PR is still showing `REVIEW_REQUIRED` because the changed files under `pkg/kvevents/engineadapter/` fall under the global CODEOWNERS rule, and a code-owner approval is needed to unblock merge. @vMaroon — since you triaged this originally, could you give it a code-owner review (or merge if it looks good)? All CI is green. Thanks!

PR #605, opened 2026-05-25
feat: Add PVC evictor BlockRemoved events

albertoperdomo2:
## Summary

This PR adds `BlockRemoved` event emission from the PVC evictor's deleter process, complementing the `BlockStored` events added in #571. When the evictor deletes KV cache files from shared storage, it now publishes `BlockRemoved` events so downstream consumers (e.g. the Go storage indexer) can remove stale checkpoints.

The implementation extends `StorageEventPublisher` with a `publish_blocks_removed()` method that supports per-call `model_name` overrides, since the evictor handles files from multiple models on the same PVC. The `model_name` constructor parameter is optional to just support this multi-model use case.

Block hashes are extracted by reversing the `FileMapper` filename convention (16-char hex basename), and the original model name is recovered from `config.json` at the FileMapper base directory. Model name lookups are cached per base directory so only one filesystem read occurs per model per process lifetime. Events are grouped by model and published after each successful batch deletion.

## Testing

All existing tests pass. This PR adds the new deleter tests and extends storage event tests.

Validated with:

```
pytest kv_connectors/pvc_evictor/tests/test_deleter.py -q
pytest kv_connectors/llmd_fs_backend/tests/test_storage_events.py -q
```

## Related Issues

- #520 

> [!NOTE]
> Test layout depends on wether #578 is merged before this one or not.

### DIFF
--- a/kv_connectors/pvc_evictor/processes/deleter.py
+++ b/kv_connectors/pvc_evictor/processes/deleter.py
@@ -73,17 +73,17 @@ def extract_model_name(file_path: str, cache_path: str) -> str | None:
     return model_name
 
 
-def delete_batch(file_paths: list[str], dry_run: bool, logger: logging.Logger) -> tuple[int, int]:
+def delete_batch(file_paths: list[str], dry_run: bool, logger: logging.Logger) -> tuple[int, int, list[str]]:
     """
     Delete a batch of files using xargs rm -f (batch deletion).
 
     If xargs fails, logs error and skips the batch (files will be retried in next cycle).
 
-    Returns: (files_deleted, bytes_freed)
+    Returns: (files_deleted, bytes_freed, deleted_paths)
     """
     if dry_run:
         logger.debug(f"[DRY RUN] Would delete {len(file_paths)} files")
-        return len(file_paths), 0
+        return len(file_paths), 0, []
 
     valid_paths = []
     total_bytes = 0
@@ -103,7 +103,7 @@ def delete_batch(file_paths: list[str], dry_run: bool, logger: logging.Logger) -
         # All files in batch don't exist (already deleted or invalid paths)
         # This is normal - files may have been deleted between queuing and processing
         # or the same files were queued multiple times
-        return 0, 0
+        return 0, 0, []
 
     # Use xargs rm -f for batch deletion
     # Use null-terminated input for xargs -0 (safe handling of file paths with special characters)
@@ -118,7 +118,7 @@ def delete_batch(file_paths: list[str], dry_run: bool, logger: logging.Logger) -
         )
 
         if result.returncode == 0:
-            return len(valid_paths), total_bytes
+            return len(valid_paths), total_bytes, valid_paths
         else:
             # Log error and skip batch - files will be retried in next cycle
             logger.error(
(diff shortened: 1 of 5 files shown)


albertoperdomo2:
cc: @kfirtoledo

kv_connectors/pvc_evictor/processes/deleter.py
```diff
@@ -99,6 +161,20 @@ def delete_file_batch(
     batch_start_time = time.time()
     deleted, freed = delete_batch(batch, dry_run, logger)
 
+    if deleted > 0 and event_publisher is not None and cache_path is not None:
```
guygir: When dry_run=True, delete_batch() still returns len(file_paths) as deleted, so this block can publish BlockRemoved even though no files were removed. could we skip event publishing when dry_run is true (and/or only publish after a real delete)?

kv_connectors/pvc_evictor/processes/deleter.py
```diff
@@ -99,6 +161,20 @@ def delete_file_batch(
...
+        model_hashes = {}
+        for b in batch:
```
guygir: delete_batch() only removes valid_paths (files that still exist) - if a batch mixes existing and already missing paths, we might publish removals for files that weren't deleted in this pass I believe? maybe should return the paths actually deleted (or share the filtered list) and publish from that?

kv_connectors/pvc_evictor/processes/deleter.py
```diff
@@ -165,6 +262,8 @@ def deleter_process(
                                 total_bytes_freed,
                                 prev_batch_time,
                                 result_queue,
+                                event_publisher,
```
guygir, line 265: On shutdown, the remaining current_batch goes through delete_batch() only (and not delete_file_batch() like normal batches), so those deletions won't emit events. Was that intentional? If not, routing shutdown through delete_file_batch() (or a shared publish helper) would keep behavior consistent.

guygir (changes requested):
Thanks @albertoperdomo2 , overall looks good - just a few small things that need addressing.

kv_connectors/pvc_evictor/processes/deleter.py
albertoperdomo2: I do not think we should publish `BlockRemoved` events when `dry_run=True`. That can cause the Go indexer to drop valid checkpoints (in the upcoming storage indexer), which is data corruption from the indexer's perspective. The event consumer shouldn't need to know about dry runs.

kv_connectors/pvc_evictor/processes/deleter.py
albertoperdomo2: Good catch! Even though this does not affect functionality itself (an event from a missing file is basically a no-op), I agree that for correctness it should be fixed.

kv_connectors/pvc_evictor/processes/deleter.py
albertoperdomo2: Great catch here too, thanks! I'll route it through `delete_file_batch`.

kv_connectors/pvc_evictor/processes/deleter.py
guygir: Agreed, and I saw the test you added that covers it aswell. Thanks!

guygir (commented):
Thanks @albertoperdomo2 , looks great.

\lgtm

kfirtoledo (approved):

guygir (approved):

PR #611, opened 2026-05-26
fix(pvc_evictor): walk current FileMapper layout in crawler

miroslavln:
After #585 collapsed the FileMapper layout to
`<root>/<safe_model_name>_<sha256-12>_r<rank>/<hhh>/<hh>_g<group_idx>/*.bin`, the crawler still walked the pre-#585 deep tree
(`block_size_*/tp_*_pp_size_*/rank_*/<dtype>/...`). Those paths no longer exist on disk, so the crawler discovered zero files and the evictor never freed any blocks.

This rewrites `stream_cache_files_with_mapper` to:

* find rank directories by their `_r<digits>` suffix anywhere under the cache root, instead of pattern-matching the deprecated deep tree;
* iterate the first-level hex bucket ({hhh}) and apply the existing hex_modulo_range sharding;
* yield .bin files from any second-level bucket underneath, leaving the `_g<group_idx>` encoding opaque so the walker doesn't need to understand kv-cache groups.

The new walker no longer instantiates `FileMapper` (it only inspects on-disk names), so the `FILEMAPPER_AVAILABLE` import guard and the matching early-exit in `crawler_process` are removed. This also unblocks running the evictor container without vllm installed, since the walker was the only consumer of the `from llmd_fs_backend.file_mapper import FileMapper` import — which transitively required vllm.

`parse_filemapper_params` is dropped along with its sole caller.

### DIFF
--- a/kv_connectors/pvc_evictor/processes/crawler.py
+++ b/kv_connectors/pvc_evictor/processes/crawler.py
@@ -11,17 +11,11 @@ from pathlib import Path
 from utils.logging_helpers import send_stats_to_queue
 from utils.system import setup_logging
 
-# FileMapper integration for canonical cache structure
-try:
-    from llmd_fs_backend.file_mapper import FileMapper
-
-    FILEMAPPER_AVAILABLE = True
-except ImportError:
-    FILEMAPPER_AVAILABLE = False
-    FileMapper = None
-
 # Module-level logger for functions
 logger = logging.getLogger(__name__)
+
+# Pattern for a rank-suffixed FileMapper base directory: <safe_model>_<sha>_r<rank>
+_RANK_DIR_RE = re.compile(r"_r\d+$")
 
 # Constants for hex modulo load balancing
 HEX_MODULO_BASE = 16  # Number of possible hex modulo values (0-15)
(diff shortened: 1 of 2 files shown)


guygir:
Thanks @miroslavln - this looks like the cleaner implementation for the post-#585 layout fix.

I’m closing my overlapping PR #601 in favor of this one. After this lands, I’ll open a small follow-up with the non-overlapping pieces from #601 (tests/CI/docs, dockerfile cleanup if still needed, etc).

/lgtm

kv_connectors/pvc_evictor/processes/crawler.py
```diff
@@ -116,134 +80,72 @@ def get_hex_modulo_ranges(num_processes: int = 8) -> list[tuple[int, int]]:
...
-        model_name = model_dir.name
+    modulo_range_min, modulo_range_max = hex_modulo_range if hex_modulo_range else (0, HEX_MODULO_BASE - 1)
 
-        # Iterate through block_size_*_blocks_per_file_* directories
-        for block_config_dir in Path(model_dir.path).glob("block_size_*_blocks_per_file_*"):
-            if not block_config_dir.is_dir():
+    for rank_dir in _iter_rank_dirs(cache_path):
+        # Iterate first-level hex buckets ({hhh}, three hex chars).
+        for hex3_dir in safe_scandir(rank_dir.path):
+            if not hex3_dir.is_dir() or len(hex3_dir.name) != 3:
```
kfirtoledo: Can you change 3 to a parameter?

kfirtoledo (commented):
Ca

kv_connectors/pvc_evictor/processes/crawler.py
miroslavln: Sure

PR #619, opened 2026-05-29
feat(evictor): implement background empty directory cleanup process

### LINKED ISSUE #614: pvc_evictor deletion leavs large amount of empty directories, slowing down file discovery.
**Problem Description**

Impact

miroslavln:
Description
After cache files are deleted, empty subdirectories (representing shard buckets and models) are left behind on the shared storage mount. Over time, this results in directory bloat.

This PR introduces an asynchronous background cleanup process to automatically purge empty cache directories as files are evicted, ensuring a clean storage footprint.

Key Changes
**Asynchronous Folder Cleaner (folder_cleaner.py)**: Added a dedicated background process folder_cleaner_process that consumes directory paths from a multiprocessing queue (folder_queue) and purges them using os.rmdir. Because os.rmdir naturally fails on non-empty directories, this is completely safe and skips directories containing remaining block files or active caches.
**Deleter Integration (deleter.py)**: Integrated the deleter_process to extract the parent directory of every successfully deleted cache .bin file and queue it for background cleanup.
**Orchestrator Support (evictor.py)**: Updated the main evictor launcher to spin up, monitor, and gracefully shut down the new cleaner process and handle its multiprocessing queues.
**Telemetry & Aggregated Logging (logging_helpers.py)**: Added live tracking and aggregated statistics output for the directory cleaner process (e.g., tracking active queue depth, total folders deleted, and status outputs).

Fixes #614

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,7 +7,7 @@ run:
   tests: true
 output:
   formats:
-    text:
+    - format: colored-line-number
       path: stdout
       print-linter-name: true
       print-issued-lines: false
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,7 +7,7 @@ run:
   tests: true
 output:
   formats:
-    - format: colored-line-number
+    text:
       path: stdout
       print-linter-name: true
       print-issued-lines: false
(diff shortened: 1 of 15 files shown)


guygir (dismissed):
Thank you for the follow-up @miroslavln, overall looks great. Bottom up cleanup in the folder cleaner and pulling in the #625 tests cover what we were missing - thanks @Jwrede.

Just need to fix lint CI and we're clear.

Nits:
- Startup log still says "N+2-Process Architecture" which may now be N+3. Only cosmetic though.
- Maybe worth more unit tests for queue_folder TTL / crawler empty-dir skips if we want coverage on race edge cases

Issue #549, opened 2026-04-27
[RFC][CI]:  CI/CD Improvements for `llm-d-kv-cache`


hickeyma:
## Summary

This RFC proposes a phased set of improvements to the `llm-d-kv-cache` CI/CD pipeline. The current pipeline has a solid foundation (lint, unit/e2e tests, nightly race detection, multi-arch image builds, Dependabot, Prow automation) but there are some gaps in supply-chain security, test coverage enforcement, pipeline reliability and release process quality.

## Motivation

An improved and reliable CI.

## Current State

| Workflow | Trigger | Covers |
|---|---|---|
| `ci-lint.yaml` | PR → main/dev | golangci-lint, pre-commit, clang-format |
| `ci-test.yaml` | PR → main/dev | Go unit tests + e2e (testcontainers) |
| `ci-examples.yaml` | PR → **main only** | Example binaries smoke test |
| `ci-uds-tokenizer.yaml` | PR (path-filtered) | Python integration tests |
| `ci-nightly-race.yaml` | Daily 06:00 UTC | Go race detector |
| `ci-dev-uds-tokenizer.yaml` | Push → main/release-* | Build & push UDS tokenizer image + Trivy |
| `ci-release.yaml` / `ci-release-uds-tokenizer.yaml` | Tag push / release published | Build & push main image + Trivy |
| `ci-wheels.yaml` | Tag push | CUDA wheel build (amd64 + arm64) |

## Gaps

- No code coverage tracking, reporting or enforcement threshold
- Supply-chain vulnerabilities: unsigned images, no SBOM, mutable `@main` action refs, Trivy findings with better visibility
- Race detector only runs nightly whereas races introduced in PRs merge undetected
- No protobuf/generated-code drift check
- Significant step duplication across six workflows (no shared composite action)
- No performance regression detection or benchmark tracking
- Missing SAST, secret scanning and source-level dependency auditing
- No job timeouts or concurrency controls which can cause runaway jobs and redundant queuing
- No automated changelog or release notes generation
- No notification for silent nightly failures

### Phase 1 — Immediate (Security & Correctness)

> [!NOTE]
>  No new infrastructure and no new secrets required. All changes below can land in a single PR and they help close the highest-risk gaps.

[...]

PR #623, opened 2026-05-31
ci: implement Phase 1 security and reliability hardening

Jwrede:
## Summary

Implements all items from the Phase 1 "Quick-Win" checklist in #549 across 10 files:

1. **Trivy SARIF upload**: Add SARIF output + `upload-sarif` step to `trivy-scan/action.yml` so CVE findings surface in the GitHub Security tab instead of only appearing in logs. Add `security-events: write` permission to the three calling workflows (`ci-release`, `ci-dev-uds-tokenizer`, `ci-release-uds-tokenizer`).

2. **Pin mutable action ref**: Pin `estroz/rerun-actions@main` to verified commit SHA `6660c16a`.

3. **Go module integrity check**: Add `go mod verify && go mod tidy && git diff --exit-code` step to `ci-test.yaml` so tampered or drifted modules fail CI.

4. **Fix ci-examples.yaml**: Add missing `libzmq3-dev` install block (matching the pattern in ci-lint and ci-test) so ZMQ-dependent examples no longer fail with cgo link errors.

5. **Concurrency groups**: Add `concurrency:` blocks to all 6 PR-triggered workflows (`ci-test`, `ci-lint`, `ci-examples`, `ci-uds-tokenizer`, `non-main-gatekeeper`, `prow-pr-remove-lgtm` already delegate to reusable workflows that handle their own concurrency). Rapid pushes now cancel stale runs.

6. **Job timeouts**: Add `timeout-minutes: 30` (15 for uds-tokenizer) to all jobs that lacked one. Only `ci-nightly-race` had a timeout before.

7. **Reduce automerge poll**: Change `prow-pr-automerge.yml` cron from `*/5` to `*/30` and add `pull_request_review` + `check_suite` event triggers for faster reaction without the polling cost.

8. **Extend ci-examples.yaml to dev**: Add `dev` to the branch filter so example breakage is caught before merge to main.

Ref: #549

## Test plan

- [ ] CI lint and test pass on this PR (no Go source changes, only workflow/action YAML)
- [ ] Trivy SARIF: verify findings appear in Security tab after a release build
- [ ] Concurrency: push two commits rapidly and verify the first run is cancelled
- [ ] Go mod check: submit a PR with a dirty go.sum and verify CI fails
- [ ] Examples: verify ci-examples no longer fails on ZMQ link errors

### DIFF
--- a/.github/actions/trivy-scan/action.yml
+++ b/.github/actions/trivy-scan/action.yml
@@ -1,14 +1,26 @@
 name: Trivy Scan
-description: Scan container image with Trivy
+description: Scan container image with Trivy and upload SARIF
 inputs:
   image:
     required: true
 runs:
   using: "composite"
   steps:
-    - name: Run Trivy vulnerability scanner
+    - name: Run Trivy vulnerability scanner (table)
       uses: aquasecurity/trivy-action@57a97c7e7821a5776cebc9bb87c984fa69cba8f1 # v0.35.0
       with:
         image-ref: ${{ inputs.image }}
         format: "table"
         severity: "HIGH,CRITICAL"
+    - name: Run Trivy vulnerability scanner (SARIF)
+      uses: aquasecurity/trivy-action@57a97c7e7821a5776cebc9bb87c984fa69cba8f1 # v0.35.0
+      with:
+        image-ref: ${{ inputs.image }}
+        format: "sarif"
+        output: "trivy-results.sarif"
+        severity: "HIGH,CRITICAL"
+    - name: Upload Trivy SARIF to GitHub Security tab
+      uses: github/codeql-action/upload-sarif@v3
+      if: always()
+      with:
+        sarif_file: "trivy-results.sarif"
--- a/.github/actions/trivy-scan/action.yml
+++ b/.github/actions/trivy-scan/action.yml
@@ -6,21 +6,17 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - name: Run Trivy vulnerability scanner (table)
-      uses: aquasecurity/trivy-action@57a97c7e7821a5776cebc9bb87c984fa69cba8f1 # v0.35.0
-      with:
-        image-ref: ${{ inputs.image }}
-        format: "table"
-        severity: "HIGH,CRITICAL"
-    - name: Run Trivy vulnerability scanner (SARIF)
+    - name: Run Trivy vulnerability scanner
       uses: aquasecurity/trivy-action@57a97c7e7821a5776cebc9bb87c984fa69cba8f1 # v0.35.0
       with:
         image-ref: ${{ inputs.image }}
         format: "sarif"
         output: "trivy-results.sarif"
         severity: "HIGH,CRITICAL"
+        # SARIF output reports every severity unless this is set.
+        limit-severities-for-sarif: true
     - name: Upload Trivy SARIF to GitHub Security tab
-      uses: github/codeql-action/upload-sarif@v3
+      uses: github/codeql-action/upload-sarif@02c5e83432fe5497fd85b873b6c9f16a8578e1d9 # v3.37.0
       if: always()
       with:
         sarif_file: "trivy-results.sarif"
--- a/.github/workflows/ci-examples.yaml
+++ b/.github/workflows/ci-examples.yaml
@@ -4,13 +4,39 @@ on:
   pull_request:
     branches:
       - main
+      - dev
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
 
 jobs:
   run-examples:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     steps:
       - name: Checkout source
         uses: actions/checkout@v6
+
+      - name: Cache system (apt) dependencies
+        uses: actions/cache@v4
+        with:
+          path: apt-archives
+          key: ${{ runner.os }}-apt-examples-${{ hashFiles('**/.github/workflows/ci-examples.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-apt-
+
+      - name: Install system dependencies
+        run: |
+          mkdir -p apt-archives
+          sudo cp -a apt-archives /var/cache/apt/archives
+          sudo apt-get update
+          sudo apt-get install -y libzmq3-dev pkg-config
+          cp -a /var/cache/apt/archives/*.deb apt-archives 2>/dev/null || true
+
       - name: Extract Go version from go.mod
         run: sed -En 's/^go (.*)$/GO_VERSION=\1/p' go.mod >> $GITHUB_ENV
 
--- a/.github/workflows/ci-examples.yaml
+++ b/.github/workflows/ci-examples.yaml
@@ -21,21 +21,10 @@ jobs:
       - name: Checkout source
         uses: actions/checkout@v6
 
-      - name: Cache system (apt) dependencies
-        uses: actions/cache@v4
-        with:
-          path: apt-archives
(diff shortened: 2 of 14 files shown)


.github/workflows/ci-examples.yaml
```diff
@@ -4,13 +4,39 @@ on:
...
+        with:
+          path: apt-archives
+          key: ${{ runner.os }}-apt-examples-${{ hashFiles('**/.github/workflows/ci-examples.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-apt-
+
+      - name: Install system dependencies
+        run: |
+          mkdir -p apt-archives
+          sudo cp -a apt-archives /var/cache/apt/archives
```
yankay: Thanks @Jwrede! Small suggestion on the apt cache in `ci-examples.yaml`:

  For just `libzmq3-dev` + `pkg-config` (~1-2 MB), the cache step isn't really worth it — `apt-get update` still runs every time, and there's also a
  path bug:

  ```yaml
  sudo cp -a apt-archives /var/cache/apt/archives
  ```

  Since `/var/cache/apt/archives/` already exists, `cp -a SRC DEST` nests the contents as a subdirectory — debs end up at
  `/var/cache/apt/archives/apt-archives/*.deb`. apt only scans the top level, so cache never hits and every run re-downloads. (Fix would be `cp -a
  apt-archives/. /var/cache/apt/archives/`.)

  Given the small payload, could we just drop the cache step entirely?

  ```yaml
  - name: Install system dependencies
    run: |
      sudo apt-get update
      sudo apt-get install -y libzmq3-dev pkg-config
  ```

  Matches the original intent and keeps the diff minimal.

Jwrede: Good catch on the cp -a nesting bug, dropped the cache step entirely in 46cbbbf. Agreed it's not worth the complexity for two small packages.

.github/actions/trivy-scan/action.yml
```diff
@@ -1,14 +1,26 @@
...
         image-ref: ${{ inputs.image }}
         format: "table"
         severity: "HIGH,CRITICAL"
+    - name: Run Trivy vulnerability scanner (SARIF)
+      uses: aquasecurity/trivy-action@57a97c7e7821a5776cebc9bb87c984fa69cba8f1 # v0.35.0
+      with:
+        image-ref: ${{ inputs.image }}
+        format: "sarif"
+        output: "trivy-results.sarif"
+        severity: "HIGH,CRITICAL"
```
hickeyma: Why do we need 2 steps for table and SARIF. Can this be done in 1 step?

hickeyma (changes requested):
Thanks @Jwrede for pushing the PR. Some comment below and inline:

- You added `@main` pin to `re-run-action.yml`. Do you mind adding it also to `prow-pr-automerge.yml`, `non-main-gatekeeper.yml`, and `prow-github.yml`
- You added `concurrency:` blocks to `ci-test`, `ci-lint`, `ci-examples` and `ci-uds-tokenizer`.  It needs to also be added to `non-main-gatekeeper.yml` and `prow-pr-remove-lgtm.yml` as llm-d/llm-d-infra` does not have `concurrency:` 
- Timeouts need to be added to the following also: 
   - `ci-release.yaml`, `ci-release-uds-tokenizer.yaml`, `ci-dev-uds-tokenizer.yaml` (docker build+push+Trivy scan, multi-arch)
   - `ci-wheels.yaml` (CUDA matrix build, amd64/arm64 × cu12/cu130 = 4 parallel jobs, Docker builds)
   - `prow-pr-automerge.yml`, `re-run-action.yml`
- Pin SHA to `github/codeql-action/upload-sarif@v3 ` in `.github/actions/trivy-scan/action.yml`

Jwrede:
Thanks @hickeyma, and no problem on the timing. Pushed `df4e349`, which rebases onto current main and works through the review. One item could not be done as asked, details below.

**SHA pins for the reusable workflows**

Pinned `prow-pr-automerge.yml`, `non-main-gatekeeper.yml` and `prow-github.yml` to llm-d-infra commit `991b736`. I also pinned `prow-pr-remove-lgtm.yml`, which was not on your list but carries the same `@main` ref and so had the same exposure. Dependabot already watches the `github-actions` ecosystem weekly in this repo, so these pins get bumped automatically instead of going stale.

**Concurrency**

Added to `non-main-gatekeeper.yml` and `prow-pr-remove-lgtm.yml`, using the same group expression as the other workflows.

**Timeouts (one of these is not possible here)**

Added `timeout-minutes` to `ci-release.yaml`, `ci-release-uds-tokenizer.yaml`, `ci-dev-uds-tokenizer.yaml`, `ci-wheels.yaml` (both jobs) and `re-run-action.yml`.

I could not add one to `prow-pr-automerge.yml`. That job calls a reusable workflow via `uses:`, and GitHub does not allow `timeout-minutes` on a reusable workflow caller. Adding it fails workflow validation:

```
when a reusable workflow is called with "uses", "timeout-minutes" is not available.
only following keys are allowed: "name", "uses", "with", "secrets", "needs", "if", and "permissions"
```

The same limitation applies to `non-main-gatekeeper.yml`, `prow-github.yml` and `prow-pr-remove-lgtm.yml`. The timeout has to sit on the job *inside* the reusable workflow, which lives in llm-d-infra. `reusable-prow-automerge.yaml` currently has no `timeout-minutes` on its `auto-merge` job. I am happy to open a PR against llm-d-infra adding timeouts to the reusable workflows, which would cover every consuming repo at once. Just say the word.

For the values, I used roughly twice the observed maximum successful run duration rather than picking numbers arbitrarily:

| workflow | observed max | timeout |
| --- | --- | --- |
| `ci-release.yaml` | 24 min | 45 |
| `ci-release-uds-tokenizer.yaml` | 21 min | 45 |
| `ci-dev-uds-tokenizer.yaml` | 20 min | 45 |
| `ci-wheels.yaml` (build-wheels) | 11 min | 30 |
| `ci-wheels.yaml` (attach-assets) | short | 10 |
| `re-run-action.yml` | 4 min | 10 |

**Trivy: collapsing the two steps**

Good catch, collapsed to one scan. `trivy-action` exposes a single `format` input, so table plus SARIF genuinely meant scanning the same image twice. The SARIF upload already gives a durable, filterable record in the Security tab, so I dropped the table scan and kept SARIF only. The table step was never gating anything: it sets no `exit-code`, so it was purely informational log output.

That change also surfaced a bug in my own earlier patch. SARIF output reports **every** severity unless `limit-severities-for-sarif` is set, so the Security tab would have been receiving LOW, MEDIUM and UNKNOWN findings despite `severity: "HIGH,CRITICAL"`. That is set now, so the SARIF report honours the same filter as the old table did.

If you would rather keep the human readable table in the build log as well, I can add it back as a second invocation with `cache: true` and `skip-setup-trivy: true`, which makes the second pass cheap. Your call.

Also pinned `github/codeql-action/upload-sarif` to `02c5e83` (v3.37.0).

All workflows are clean under `actionlint`.

.github/actions/trivy-scan/action.yml
Jwrede: Collapsed to a single step. trivy-action takes one `format` input per invocation, so table plus SARIF meant scanning the same image twice. The table scan set no `exit-code`, so it was never gating anything, just log output, and the SARIF upload already gives a durable record in the Security tab. Dropped the table and kept SARIF only.

This also surfaced a bug in my earlier patch: SARIF ignores the `severity` filter unless `limit-severities-for-sarif` is set, so the Security tab would have been getting LOW/MEDIUM/UNKNOWN findings despite `severity: "HIGH,CRITICAL"`. Set now.

If you want the readable table back in the log too, I can re-add it with `cache: true` and `skip-setup-trivy: true` so the second pass is cheap.

hickeyma: That is better, thanks

hickeyma (approved):
LGTM, thanks @Jwrede for the work on this.

PR #625, opened 2026-06-01
fix(pvc_evictor): clean up empty directories after batch file deletion

### LINKED ISSUE #614: pvc_evictor deletion leavs large amount of empty directories, slowing down file discovery.

Jwrede:
## Summary

- Add bottom-up empty directory cleanup after each successful batch file deletion in the deleter process
- After `xargs rm` completes, walk from each deleted file's parent directory upward, removing empty directories until reaching the cache root
- Uses `os.rmdir` which only succeeds on empty directories, so concurrent file creation by other processes is safe

Fixes #614

## Details

The pvc_evictor's deleter process removes `.bin` cache files via `xargs rm -f` but leaves behind the nested directory tree created by the FileMapper (`<model>_<sha>_r<rank>/<hhh>/<hh>_g<group_idx>/`). Over time, hundreds of thousands of empty directories accumulate and the crawler must traverse all of them, degrading file discovery performance.

The `cleanup_empty_dirs()` function collects unique parent directories from the deleted file paths, sorts them deepest first (longest path first), and walks upward attempting `os.rmdir` on each level. The walk stops at the cache root boundary or at the first non-empty directory. Directories that were already removed by a concurrent process or a previous iteration are skipped without error.

The `cache_path` parameter is optional to maintain backward compatibility. All existing call sites in `deleter_process` now pass `cache_path` through.

## Test plan

- [x] Unit tests for `cleanup_empty_dirs` covering: leaf + parent removal, full branch removal, cache root boundary, non-empty directories, already-removed directories
- [x] Unit tests for `delete_batch` integration: cleanup with cache_path, dry run preserves dirs, no cache_path skips cleanup
- [x] All 8 tests pass locally

### DIFF
--- a/kv_connectors/pvc_evictor/processes/deleter.py
+++ b/kv_connectors/pvc_evictor/processes/deleter.py
@@ -3,6 +3,7 @@
 import contextlib
 import logging
 import multiprocessing
+import os
 import subprocess
 import time
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


Jwrede:
Closing in favor of #619 by @miroslavln, which uses a dedicated background process that fits the evictor's multiprocess architecture better and also catches empty directories during crawling.

Two things from this PR that might be worth incorporating into #619:
- Bottom-up parent directory walk: this PR removes the entire empty chain (leaf up to cache root) in one pass, while #619 queues individual directories and relies on the next crawl sweep to discover newly-empty parents.
- Unit tests for the cleanup logic (8 tests here covering edge cases like already-removed dirs, non-empty dirs, cache root boundary).

PR #626, opened 2026-06-01
feat(kvblock): invalidate KV index on AllBlocksCleared via eager delete

vMaroon:
Wire the AllBlocksCleared KV-event to a new Index.Clear(podIdentifier) that eagerly removes the pod's entries across all device tiers. vLLM emits AllBlocksCleared only from reset_prefix_cache() — a pod-wide prefix-cache reset (with no tier) that, in RL rollouts, fires once per weight sync inside a drain/barrier (the engine is paused and the cache is cold afterward regardless). That cadence and slack make an O(N) clear off the hot path the right trade: it adds zero steady-state cost to the Lookup/Add paths (the index's hottest path), unlike a per-entry generation filter.

- Index.Clear(ctx, podIdentifier): pod-wide, all tiers. A string arg (not a PodEntry) so the signature cannot imply tier scoping the event never carries.
- InMemory: iterate keys, reuse evictPodsFromRequestKey for race-safe removal.
- CostAware: add a keyIndex set (ristretto has no iteration); prune on clear/evict.
- Redis/Valkey: SCAN + HDEL the pod's "<pod>@" fields + prune empty hashes; deletes from the shared store, so it is correct across replicas without cross-process coordination.
- pool.go: AllBlocksCleared now calls index.Clear (was a no-op log).
- Shared interface tests: basic clear, pod isolation, all-tiers, re-add.

### DIFF
--- a/pkg/kvcache/kvblock/cost_aware_memory.go
+++ b/pkg/kvcache/kvblock/cost_aware_memory.go
@@ -57,16 +57,7 @@ func NewCostAwareMemoryIndex(cfg *CostAwareMemoryIndexConfig) (*CostAwareMemoryI
 	}
 
 	// Parse the size string to get byte value using go-humanize
-
 	sizeBytes, err := humanize.ParseBytes(cfg.Size)
-	if err != nil {
-		return nil, fmt.Errorf("failed to initialize cost aware index: %w", err)
-	}
-	cache, err := ristretto.NewCache(&ristretto.Config[string, *CostPodCache]{
-		NumCounters: defaultNumCounters, // number of keys to track.
-		MaxCost:     int64(sizeBytes),   // #nosec G115 , maximum cost of cache
-		BufferItems: defaultBufferItems, // number of keys per Get buffer.
-	})
 	if err != nil {
 		return nil, fmt.Errorf("failed to initialize cost aware index: %w", err)
 	}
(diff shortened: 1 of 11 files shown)


vMaroon:
@yash9263 - I want to revisit the gen-counter direction. I think your original approach - eager eviction with the pod -> requestKeys reverse index - was closer to right, and I'd like to pivot back toward it.

What changed my mind is looking at when AllBlocksCleared actually fires. In vLLM it's only emitted from reset_prefix_cache(), which in an RL rollout is called once per weight sync - at a point where the engine is drained and the prefix cache is cold afterwards regardless. We also consume the event asynchronously, off the engine's critical path. So Clear runs with plenty of slack, and the gen-counter spends a permanent per-Lookup cost to optimize something that was never on the hot path. That's the same concern I raised with it originally, and the timing makes it concrete.

The one change I'd make to your original is to drop the reverse index and just scan on Clear. Maintaining it added 17-50% to Add, and Add runs on every event - whereas Clear is rare and off the critical path, so an O(N) scan there is an acceptable trade. The result is no steady-state overhead and a lot less moving parts.

One caveat worth calling out: on Redis I'm scanning the full keyspace, which is the one case where your reverse index was clearly better. If we expect Redis-backed multi-replica with a large keyspace, keeping a pod -> keys set there is worth it. I'd leave it out for now and add it if profiling shows we need it.

I pushed a minimal version here, with you as co-author. If you prefer taking this (and possibly modifying) in your open PR, that also works.

yankay:
HI @vMaroon 

One concern I had while reviewing is that the current Redis Clear path scans the whole keyspace and then removes fields by pod prefix. 

I put together a small follow-up PR to illustrate a possible no-Lua reverse-index approach: https://github.com/vMaroon/llm-d-kv-cache/pull/2

Please feel free to use, adapt, or ignore it if you prefer a different direction.

vMaroon:
> HI @vMaroon
> 
> One concern I had while reviewing is that the current Redis Clear path scans the whole keyspace and then removes fields by pod prefix.
> 
> I put together a small follow-up PR to illustrate a possible no-Lua reverse-index approach: [vMaroon#2](https://github.com/vMaroon/llm-d-kv-cache/pull/2)
> 
> Please feel free to use, adapt, or ignore it if you prefer a different direction.

I think we should have this work as a follow-up PR with numbers. What do you think?

yankay:
Thanks, that sounds reasonable to me. I’m okay with merging this PR as-is and handling the Redis reverse-index optimization as a follow-up with numbers.

/lgtm.

yankay (approved):

PR #628, opened 2026-06-03
fs_backend: migrate kv_offload API to vllm 0.22.0

kfirtoledo:
## Summary

Migrate `llmd_fs_backend` to vLLM **0.22.0**.

The vLLM 0.22 KV-offload Python API (`vllm.v1.kv_offload.base`, `worker.worker`, and the `OffloadPromMetrics` patch target) is **unchanged from 0.21**, so no import or signature changes are needed.

The one functional change is in `spec.py`: adopt the `OffloadingSpec` base class's `self.hash_block_size` (resolved via `resolve_kv_cache_block_sizes`) instead of `vllm_config.cache_config.block_size` when computing `gpu_blocks_per_file` and the handler's `gpu_block_size`. The two are equal for standard single-group models, but `cache_config.block_size` can be larger on hybrid models (e.g. DSv4), so the base-class value is the correct hash granularity.

Scope: **single-group only** (no HMA).

### Changes
- `spec.py`: use `self.hash_block_size` for `gpu_blocks_per_file` and handler `gpu_block_size`
- Bump pins: `vllm 0.21.0 → 0.22.0`, package `version 0.21 → 0.22`
- `README.md`, `Dockerfile.dev` base image, and `vllm-storage.yaml` deployment example updated to `0.22`

### DIFF
--- a/kv_connectors/llmd_fs_backend/Dockerfile.dev
+++ b/kv_connectors/llmd_fs_backend/Dockerfile.dev
@@ -17,10 +17,10 @@
 # For a cu130 dev image, override BOTH at build time:
 #
 #   docker build \
-#     --build-arg VLLM_IMAGE=vllm/vllm-openai:v0.21.0-cu130 \
+#     --build-arg VLLM_IMAGE=vllm/vllm-openai:v0.22.0-cu130 \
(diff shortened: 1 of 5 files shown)


dannyharnik (approved):
lgtm
\approve

PR #631, opened 2026-06-04
fix(make): use linux platform in Docker build targets

Alex-ai-future:
## Summary

On macOS, `TARGETOS` resolves to `darwin`, causing `docker build --platform darwin/arm64` to fail since base images only publish `linux/*` manifests.

CI was unaffected — it uses a separate buildx action with hardcoded `linux/*` platforms.

## Changes

- `--platform $(TARGETOS)/$(TARGETARCH)` → `--platform linux/$(TARGETARCH)` in all 3 Docker build targets
- Remove redundant `--build-arg TARGETOS/TARGETARCH` — Docker auto-injects these from `--platform` and Dockerfiles don't reference them

## Impact

| Environment | Before | After |
|------------|--------|-------|
| macOS | `darwin/*` → fails | `linux/*` → works |
| Linux | no change | no change |
| CI | no change | no change |

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,9 +107,7 @@ e2e-test: e2e-test-uds ## Run end-to-end tests
 image-build-uds: check-container-tool ## Build the UDS tokenizer container image
 	@printf "\033[33;1m==== Building UDS tokenizer image $(UDS_TOKENIZER_IMAGE) ====\033[0m\n"
 	$(CONTAINER_TOOL) build \
-		--platform $(TARGETOS)/$(TARGETARCH) \
-		--build-arg TARGETOS=$(TARGETOS) \
-		--build-arg TARGETARCH=$(TARGETARCH) \
+		--platform linux/$(TARGETARCH) \
 		-t $(UDS_TOKENIZER_IMAGE) services/uds_tokenizer
 
 .PHONY: e2e-test-uds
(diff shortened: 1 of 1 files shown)


liu-cong:
On Mac, why not just set the correct TARGETOS to `linux` when running the build? IMO Hard coding to `linux` is only acceptable when that's the only supported os, which I am uncertain.

Alex-ai-future:
Thank you @liu-cong! 

The fix is fundamentally about semantics: --platform in docker build specifies the target OS the container image will run on, not the host OS doing the build. The container always targets linux, regardless of whether you build from macOS or Linux.

TARGETOS from go env GOOS returns darwin on Mac, which is correct for Go compilation on the host, but passing --platform darwin/arm64 to docker build is semantically wrong — there's no such thing as a darwin container image for any of our base images.

So linux/$(TARGETARCH) is the correct semantic value. TARGETARCH stays variable because we do support both amd64 and arm64 targets, but the OS is always linux.

liu-cong:
> So linux/$(TARGETARCH) is the correct semantic value. TARGETARCH stays variable because we do support both amd64 and arm64 targets, but the OS is always linux.

Can we default to linux if target os is not set instead of hard coding it to linux? I think windows may be supported as well? though I haven't tested it.

Alex-ai-future:
Thank you @liu-cong!

A conditional approach ($(if $(filter darwin,$(TARGETOS)),linux,$(TARGETOS))) would keep the door open for Windows, with minimal added complexity.

However, --platform specifies the container target OS, not the host OS. These are semantically Linux images — linux/$(TARGETARCH) reflects that directly.

Even if this project ever needed Windows containers, it would require a completely separate Dockerfile and a Windows CI runner — no --platform flag alone could bridge that. So hardcoding linux/ doesn't block anything future-proof.

liu-cong (approved):

PR #638, opened 2026-06-05
fix(pvc-evictor): clean empty cache directories

### LINKED ISSUE #614: pvc_evictor deletion leavs large amount of empty directories, slowing down file discovery.

larainema:
## Summary

- Remove empty FileMapper parent directories after a successful PVC evictor deletion batch.
- Keep cleanup bounded to `cache_path` so the cache root and model metadata directories are preserved.
- Add tests for fully empty directory cleanup and non-empty parent preservation.

Fixes #614

## Testing

- `uv run --with pytest python -m pytest kv_connectors/pvc_evictor/tests/test_deleter.py`
- `uv run --with pytest python -m pytest kv_connectors/pvc_evictor/tests`
- `uv run --with ruff ruff check kv_connectors/pvc_evictor/processes/deleter.py kv_connectors/pvc_evictor/tests/test_deleter.py`
- `uv run --with ruff ruff format --check kv_connectors/pvc_evictor/processes/deleter.py kv_connectors/pvc_evictor/tests/test_deleter.py`

### DIFF
--- a/kv_connectors/pvc_evictor/processes/deleter.py
+++ b/kv_connectors/pvc_evictor/processes/deleter.py
@@ -141,6 +141,40 @@ def delete_batch(file_paths: list[str], dry_run: bool, logger: logging.Logger) -
         return 0, 0, []
 
 
+def cleanup_empty_parent_dirs(deleted_paths: list[str], cache_path: str | Path, logger: logging.Logger) -> int:
+    """Remove empty FileMapper parent directories under cache_path.
+
+    Deleted cache files live below:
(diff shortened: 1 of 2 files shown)


miroslavln:
We already have a PR outstanding for this issue https://github.com/llm-d/llm-d-kv-cache/pull/619

larainema:
Thanks for pointing this out. I missed #619, and it already covers the #614 empty-directory cleanup with a broader background cleaner approach plus tests. Closing this PR as a duplicate so review can stay focused there.

PR #215, opened 2025-12-15
PVC Evictor

guygir:
The PVC Evictor is a multi process kubernetes deployment designed to automatically manage disk space on PVCs used for vLLM KV cache offloading. It monitors disk usage and automatically deletes cold cache files when storage thresholds are exceeded, ensuring continuous operation of vLLM workloads without manual intervention.

- Follow up optimization plans can be seen at Issue #218 .
- Once complete benchmark results on llm-d will be available, I'll post them as a comment here.

PR #649, opened 2026-06-09
fix(pvc_evictor): configure activator log level via config_dict

Alex-ai-future:
The activator process (P(N+1)) has hardcoded its log level to "INFO" since the initial PVC Evictor PR (#215), ignoring the LOG_LEVEL environment variable that all other processes respect.

Root cause: the activator function signature did not receive config_dict, making log_level inaccessible. Other processes (Crawler/Deleter/FolderCleaner) were updated to use config_dict over time, but activator was left behind.

Changes:
- processes/activator.py: accept config_dict parameter; pass config_dict["log_level"] and config_dict.get("log_file_path") to setup_logging instead of hardcoded "INFO" and os.getenv
- evictor.py: forward self.config_dict to activator_process in both initial spawn and restart paths
- tests/test_activator.py: add 2 unit tests verifying setup_logging receives the correct log_level and log_file_path from config_dict
- .gitignore: ignore kv_connectors/pvc_evictor/.venv

Files changed: 4 (3 modified, 1 new)
Tests: 41 passed (2 new activator tests)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,5 +84,6 @@ cufile.json
 services/uds_tokenizer/tokenizers
 services/uds_tokenizer/.venv
 kv_connectors/llmd_fs_backend/.venv
+kv_connectors/pvc_evictor/.venv
 
 **/vllm_source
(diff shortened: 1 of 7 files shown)


guygir (approved):
Good catch - wiring config_dict on spawn/restart looks good.

PR #650, opened 2026-06-09
feat(kvcache): HMA-aware KV block scoring with window-aware SWA

sagearc:
## Summary

Adds HMA-aware prefix scoring, building on the group metadata parsed in #612 and indexed in #627.

The scorer mirrors vLLM's hybrid cache-hit convergence: route on the prefix that all attention groups actually share, rather than treating every group as plain full attention. Scope is homogeneous block sizes — all KV cache groups share one block size, equal to the router's hash block size (the common case). Differing-block-size hybrids are deferred (see Notes).

## What changed

- two-phase scoring in `LongestPrefixScorer`:
  1. main-attention (full / MLA / sink-full) contiguous prefix from block 0 — the binding constraint, since full attention needs the whole prefix
  2. sliding-window reduction — a right-to-left trailing-window scan per SWA group that can only shrink the prefix, mirroring vLLM's `SlidingWindowManager.find_longest_cache_hit`
- the SWA trailing-window length is `cdiv(window-1, blockSize)` using the group's own block size — mirroring vLLM's `_contiguous_blocks_for_hit`
- `GroupCatalog` learns group kind / block size / window from `BlockStored` events at runtime — no static per-model config; wired into the scorer via `Indexer.SetGroupCatalog`
- `group_idx == 0` fallback when a group's kind is not yet learned, matching the Dynamo consumer (ai-dynamo/dynamo#8751)

## HMA context

Third step of the HMA support tracked in #336, after #612 (parse metadata) and #627 (index group identity).

The algorithm is grounded in vLLM's `HybridKVCacheCoordinator.find_longest_cache_hit` convergence and the per-type managers, not the issue's original presence-set sketch.

## Notes

Scope is homogeneous block sizes (all groups share the router's hash block size). Differing-block-size hybrids (e.g. Gemma) are unsupported at the indexing layer — the single request-key granularity cannot match a differently-sized group's blocks — and are deferred (#336).

The phase-2 reduction is a single sequential per-group min — exact for one modeled SWA group (homogeneous windows). Multiple heterogeneous SWA groups would need vLLM's fixed-point re-check; also deferred (#336).

## Related

- #336
- #612
- #627
- https://github.com/vllm-project/vllm/pull/40984


examples/helper/events.go
```diff
@@ -37,7 +37,7 @@ func SimulateProduceEvent(ctx context.Context, publisher *Publisher) error {
 	// Use enough tokens to fill 4 blocks (matching PromptHashes count) at blockSize=16.
 	tokenIds := make([]uint32, 64)
 	for i := range tokenIds {
-		tokenIds[i] = uint32(i + 1) //nolint:gosec // i is bounded by len(tokenIds)=64, no overflow
+		tokenIds[i] = uint32(i + 1)
```
sagearc: unrelated lint error

pkg/kvevents/zmq_subscriber_bench_test.go
```diff
@@ -70,7 +70,7 @@ func BenchmarkZMQSubscriber_Throughput(b *testing.B) {
 	start := time.Now()
 
 	for i := 0; i < b.N; i++ {
-		binary.BigEndian.PutUint64(seqBytes, uint64(i)) //nolint:gosec // i is a non-negative loop counter, conversion is safe
+		binary.BigEndian.PutUint64(seqBytes, uint64(i))
```
sagearc: unrelated lint error

pkg/tokenization/pool_test.go
```diff
@@ -151,7 +151,7 @@ func TestPool_WorkerLoop(t *testing.T) {
 			setupMocks: func(mt *MockTokenizer) {
 				for i := range 5 {
 					prompt := "prompt " + string(rune('a'+i))
-					tokens := []uint32{uint32(i), uint32(i + 1)} //nolint:gosec // test code
+					tokens := []uint32{uint32(i), uint32(i + 1)}
```
sagearc: unrelated lint error

sagearc:
cc @vMaroon

vMaroon:
In its current state, this PR only helps pure SWA models - deferring to after release.


/hold

PR #661, opened 2026-06-11
fix(kvevents): support map-encoded vLLM KV events (vLLM #42892 compatibility)

Change72:
## What

vLLM dropped `array_like=True` from its KV cache event structs in [vllm-project/vllm#42892](https://github.com/vllm-project/vllm/pull/42892) (merged 2026-06-09). Newer vLLM versions therefore publish each KV event as a **field-name map** with the tag under the `"type"` key, instead of the positional array this adapter decodes. Result: every KV event from a new vLLM fails to parse (`failed to decode tagged union`), and the index receives nothing.

## How

`decodeVLLMEvent` now accepts both encodings: map-encoded events are normalized to the existing positional `[]any` layout before dispatch, so the converters — and their forward/backward-compatibility guards (length-guarded optional trailing fields) — stay encoding-agnostic and untouched. Absent map fields become `nil`, exactly like an omitted trailing array field. Positional arrays from older vLLM versions keep working unchanged; unknown tags keep flowing through the existing converter-lookup error path.

## Validation

- New unit tests: map-encoded `BlockStored` (full + omitted optional fields), `BlockRemoved`, `AllBlocksCleared`, a mixed array/map batch, and malformed map events (missing / non-string tag).
- `go test ./pkg/kvevents/...` green; `golangci-lint run` clean on the package.
- End-to-end against a **real captured stream** from a live vLLM serve run on a commit after #42892 (map encoding; multi-turn traffic with CPU offload stores and real LRU evictions): all 313 frames parse, and replaying them through the real `kvevents.Pool → kvblock.InMemoryIndex` path indexes and evicts every block correctly (71/71 live blocks CPU-matchable, 516/516 evicted blocks cleared).

AI assistance was used for implementation and tests; all changes reviewed and verified by the submitter.

### DIFF
--- a/pkg/kvevents/engineadapter/vllm_adapter.go
+++ b/pkg/kvevents/engineadapter/vllm_adapter.go
@@ -110,7 +110,7 @@ type msgpackVLLMEventBatch struct {
 func (v *VLLMAdapter) decodeVLLMEvent(rawEventBytes []byte) (kvevents.GenericEvent, error) {
 	var decoded any
 	if err := msgpack.Unmarshal(rawEventBytes, &decoded); err != nil {
-		return nil, fmt.Errorf("failed to decode tagged union: %w", err)
+		return nil, fmt.Errorf("failed to decode vLLM event: %w", err)
 	}
 
 	var fields []any
@@ -173,9 +173,13 @@ var (
 // omitted trailing field in the array encoding. Unknown tags are passed
 // through so converter lookup reports them uniformly.
 func mapEventToFields(ev map[string]any) ([]any, error) {
-	tag, ok := ev["type"].(string)
+	rawTag, exists := ev["type"]
+	if !exists {
+		return nil, fmt.Errorf("map-encoded event is missing the %q tag", "type")
+	}
+	tag, ok := rawTag.(string)
 	if !ok {
-		return nil, fmt.Errorf("map-encoded event tag (%q) is not a string: %T", "type", ev["type"])
+		return nil, fmt.Errorf("map-encoded event tag (%q) is not a string: %T", "type", rawTag)
 	}
 
 	var order []string
--- a/pkg/kvevents/engineadapter/vllm_adapter.go
+++ b/pkg/kvevents/engineadapter/vllm_adapter.go
@@ -27,14 +27,11 @@ import (
 // VLLMAdapter implements the kvevents.EngineAdapter interface for vLLM engines.
 // It parses raw transport messages (topic + msgpack payload) into domain events.
 //
-// Older vLLM serializes events using msgspec with array_like=True and
-// omit_defaults=True, producing positional msgpack arrays where trailing default
-// fields may be absent; newer vLLM (vllm-project/vllm#42892) dropped array_like
-// and emits tagged field-name maps instead. To maintain forward and backward
-// compatibility across vLLM versions (either encoding, new fields appended or
-// trailing fields omitted), we decode into []any -- normalizing maps to the
-// positional layout -- and extract fields positionally with length guards
-// instead of using fixed structs.
+// vLLM emits events either as positional msgpack arrays with trailing defaults
+// omitted (msgspec array_like=True) or, since vllm-project/vllm#42892, as
+// field-name maps tagged under "type". Both decode into the same positional
+// []any layout, extracted with length guards instead of fixed structs, so the
+// converters stay encoding-agnostic and tolerate appended or omitted fields.
 type VLLMAdapter struct {
 	eventConverters map[string]func([]any) (kvevents.GenericEvent, error)
 }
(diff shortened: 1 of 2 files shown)


pkg/kvevents/engineadapter/vllm_adapter.go
```diff
@@ -121,6 +156,48 @@ func (v *VLLMAdapter) decodeVLLMEvent(rawEventBytes []byte) (kvevents.GenericEve
...
+
+// mapEventToFields normalizes a map-encoded vLLM event into the positional
+// []any layout the converters consume. Absent fields become nil, matching an
+// omitted trailing field in the array encoding. Unknown tags are passed
+// through so converter lookup reports them uniformly.
+func mapEventToFields(ev map[string]any) ([]any, error) {
+	tag, ok := ev["type"].(string)
+	if !ok {
+		return nil, fmt.Errorf("map-encoded event tag (%q) is not a string: %T", "type", ev["type"])
+	}
```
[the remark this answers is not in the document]
Change72: Done in c8b1756.

pkg/kvevents/engineadapter/vllm_adapter.go
```diff
@@ -96,14 +99,46 @@ type msgpackVLLMEventBatch struct {
...
+//   - tagged maps: vLLM dropped array_like in vllm-project/vllm#42892, so newer
+//     versions emit field-name maps with the tag under the "type" key. These are
+//     normalized to the positional layout so the converters stay encoding-agnostic.
 func (v *VLLMAdapter) decodeVLLMEvent(rawEventBytes []byte) (kvevents.GenericEvent, error) {
-	var fields []any
-	if err := msgpack.Unmarshal(rawEventBytes, &fields); err != nil {
+	var decoded any
+	if err := msgpack.Unmarshal(rawEventBytes, &decoded); err != nil {
 		return nil, fmt.Errorf("failed to decode tagged union: %w", err)
 	}
```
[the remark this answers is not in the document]
Change72: Done in c8b1756.

pkg/kvevents/engineadapter/vllm_adapter_test.go
```diff
@@ -611,3 +611,103 @@ func TestVLLMEventBatch_NestedArrayEvents(t *testing.T) {
...
+		"non-string tag": map[string]any{"type": 7},
+	} {
+		payload, err := msgpack.Marshal([]any{0.0, []any{event}, nil})
+		require.NoError(t, err, name)
+		_, _, _, err = adapter.ParseMessage(&kvevents.RawMessage{
+			Topic:   "kv@pod-1@m",
+			Payload: payload,
+		})
+		require.Error(t, err, name)
+	}
```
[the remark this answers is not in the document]
Change72: Done in c8b1756.

pkg/kvevents/engineadapter/vllm_adapter.go
```diff
@@ -96,12 +99,44 @@ type msgpackVLLMEventBatch struct {
...
+//
+// Two encodings are supported:
+//   - positional arrays: vLLM < v0.x encoded events with msgspec array_like=True
+//   - tagged maps: vLLM dropped array_like in vllm-project/vllm#42892, so newer
+//     versions emit field-name maps with the tag under the "type" key. These are
+//     normalized to the positional layout so the converters stay encoding-agnostic.
 func (v *VLLMAdapter) decodeVLLMEvent(rawEventBytes []byte) (kvevents.GenericEvent, error) {
+	var decoded any
+	if err := msgpack.Unmarshal(rawEventBytes, &decoded); err != nil {
+		return nil, fmt.Errorf("failed to decode vLLM event: %w", err)
```
yankay: Nice Fix.

Small nit: this now wraps decode failures with `failed to decode vLLM event`, and `ParseMessage` wraps `decodeVLLMEvent` errors with the same prefix. A malformed event can log as `failed to decode vLLM event: failed to decode vLLM event: ...`. Could we make the inner message more specific, e.g. `unmarshal event payload: %w`?

pkg/kvevents/engineadapter/vllm_adapter.go
```diff
@@ -96,12 +99,44 @@ type msgpackVLLMEventBatch struct {
...
-		return nil, fmt.Errorf("failed to decode tagged union: %w", err)
+	switch ev := decoded.(type) {
+	case []any:
+		fields = ev
+	case map[string]any:
+		var err error
+		if fields, err = mapEventToFields(ev); err != nil {
+			return nil, err
+		}
+	case map[any]any:
```
yankay: Small question: is this `map[any]any` branch intended as defensive support for a custom/untyped msgpack decoder? With the current `msgpack.Unmarshal(..., &any)` path, real msgspec-encoded vLLM events seem to decode as `map[string]interface{}`, so this branch does not appear to be exercised today.

This is not blocking, but could we add a short comment explaining that intent if we keep it?

pkg/kvevents/engineadapter/vllm_adapter.go
```diff
@@ -96,12 +99,44 @@ type msgpackVLLMEventBatch struct {
...
+		strKeyed := make(map[string]any, len(ev))
+		for k, val := range ev {
+			s, ok := k.(string)
+			if !ok {
+				return nil, fmt.Errorf("map-encoded event key is not a string: %T", k)
```
yankay: Small question: if this defensive `map[any]any` path is kept, do we want to fail fast on non-string keys as it does now, or skip them?

pkg/kvevents/engineadapter/vllm_adapter.go
Change72: Done in 53a0e08 — the inner wrap is now `unmarshal event payload: %w` as you suggested, so it reads `failed to decode vLLM event: unmarshal event payload: ...`.

pkg/kvevents/engineadapter/vllm_adapter.go
Change72: Good catch — I checked vmihailenco/msgpack v5.4.1: untyped decode goes through `DecodeMap()`, which only ever produces `map[string]any` and rejects non-string keys inside `Unmarshal` itself (and we don't set a custom map decoder anywhere). So the branch was unreachable — removed in 53a0e08 rather than documented.

pkg/kvevents/engineadapter/vllm_adapter.go
Change72: Moot now that the branch is removed in 53a0e08. For the record, fail-fast was the intent: msgspec struct field names are always strings, so a non-string key would mean the payload isn't a vLLM event at all, and skipping it would just resurface as a more confusing converter error downstream.

sagearc (approved):
Thanks, looks good.
Keeping backward compatibility for array-like KV events makes sense.

@vMaroon we should include that in the upcoming release

/lgtm

vMaroon (approved):

PR #670, opened 2026-06-16
fix(kvcache): unify engine-to-request mapping and fix Redis scoring

Alex-ai-future:
## Summary

Extract duplicated engine→request key mapping logic from three Index implementations into a shared `engineToRequestMapping()` helper, and correct the Redis ZAdd score calculation to use per-mapping relative indices.

## Problem

`InMemoryIndex`, `CostAwareMemoryIndex`, and `RedisIndex` each contained identical 7-line proportional distribution logic for computing engine-to-request key mappings. Additionally, the Redis implementation used the global loop counter `i` as the ZAdd score instead of the relative index `j` within each engine's sorted set — semantically incorrect even though current mapping ratios produce equivalent results.

## Changes

- **pkg/kvcache/kvblock/index.go**: Add `engineToRequestMapping(engineKeys, requestKeys []BlockHash) map[BlockHash][]BlockHash` with defensive empty-slice guard
- **pkg/kvcache/kvblock/{in_memory,cost_aware_memory,redis}.go**: Replace inline mapping logic with calls to the shared helper
- **pkg/kvcache/kvblock/redis.go**: Use `j` (index within each mapping slice) instead of `i` (global loop counter) for ZAdd scores
- **pkg/kvcache/kvblock/index_internal_test.go**: Add unit tests covering 1:1, many:1, 1:many, empty slices, and score ordering

## Testing

- `go test ./pkg/kvcache/...` — all existing tests pass
- New `TestEngineToRequestMapping_*` tests cover all mapping scenarios and edge cases

### DIFF
--- a/pkg/kvcache/kvblock/cost_aware_memory.go
+++ b/pkg/kvcache/kvblock/cost_aware_memory.go
@@ -244,14 +244,8 @@ func (m *CostAwareMemoryIndex) Add(ctx context.Context, engineKeys, requestKeys 
 	//   many:1 (4 eng, 1 req) -> E0->R0, E1->R0, E2->R0, E3->R0
 	//   1:many (1 eng, 4 req) -> E0->[R0, R1, R2, R3]
 	if engineKeys != nil {
-		newMappings := make(map[BlockHash][]BlockHash)
-		n := max(len(engineKeys), len(requestKeys))
-		for i := 0; i < n; i++ {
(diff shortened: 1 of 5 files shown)


yankay:
/lgtm

Thanks for cleaning this up. I reviewed the shared mapping helper and the Redis score change, and the backend behavior looks consistent with the existing index contract.

vMaroon (approved):

PR #671, opened 2026-06-16
fix(kvevents): support narrow integer types in getHashAsUint64

### LINKED ISSUE #662: Support narrow integer block hashes in vLLM KV event decoding
## Summary

`getHashAsUint64` currently accepts `int64`, `uint64`, and `[]byte`, but msgpack decoding into `any` can produce narrower Go integer types such as `int8` or `uint8` for small Python `int` values.

I have not reproduced this against a full vLLM deployment, but I was able to reproduce it with a minimal `msgspec`-encoded payload that mirrors vLLM's KV event structs.

## Minimal reproduction

Using a `BlockStored` event with an integer block hash such as `100`, Python `msgspec` encodes it as a small integer. The Go adapter then decodes it as `int8`, and parsing fails because `getHashAsUint64` does not handle that type.

Observed error:

```text
failed to decode vLLM event: failed to parse block hash: unsupported hash type: int8
```

## Impact

If `VLLM_KV_EVENTS_USE_INT_BLOCK_HASHES` is enabled and vLLM emits an integer block hash that msgpack decodes into a narrow Go integer type, the KV event batch may fail to parse and the cache index would not be updated for that event.

This does not appear specific to map-encoded events; array-encoded events with small integer hashes could hit the same helper.

## Possible fix

Accept all non-negative Go integer widths in `getHashAsUint64`, for example `int8`, `int16`, `int32`, `int64`, `int`, `uint8`, `uint16`, `uint32`, `uint64`, and `uint`.

Alex-ai-future:
## Summary

`getHashAsUint64` only handled `uint64`, `int64`, and `[]byte`, but msgpack decoding into `any` produces narrower Go integer types (`int8`, `uint8`, etc.) for small Python `int` values. This causes:

```
failed to decode vLLM event: failed to parse block hash: unsupported hash type: int8
```

## Fix

Added type-switch cases for all Go integer widths (`int`, `int8`, `int16`, `int32`, `uint`, `uint8`, `uint16`, `uint32`), aligned with the existing `toInt()` helper in `vllm_adapter.go` that already handles these types.

## Impact

| Scenario | Before | After |
|----------|--------|-------|
| Small int hash (e.g. `100`) decoded as `int8` | parse error | `uint64(100)` |
| Medium int hash decoded as `int16`/`int32` | parse error | correct `uint64` |
| Existing `uint64`/`int64`/`[]byte` paths | works | unchanged |

Fixes #662

### DIFF
--- a/pkg/kvevents/engineadapter/common.go
+++ b/pkg/kvevents/engineadapter/common.go
@@ -54,19 +54,34 @@ func getHashAsUint64(raw any) (uint64, error) {
 	case uint64:
 		return val, nil
 	case int64:
-		//nolint:gosec // int64 to uint64 conversion is safe here
+		if val < 0 {
+			return 0, fmt.Errorf("negative hash value: %d", val)
+		}
+		//nolint:gosec // int64 to uint64 conversion is safe for non-negative values
 		return uint64(val), nil
 	case int:
-		//nolint:gosec // int to uint64 conversion is safe here
+		if val < 0 {
+			return 0, fmt.Errorf("negative hash value: %d", val)
+		}
+		//nolint:gosec // int to uint64 conversion is safe for non-negative values
 		return uint64(val), nil
 	case int32:
-		//nolint:gosec // int32 to uint64 conversion is safe here
+		if val < 0 {
+			return 0, fmt.Errorf("negative hash value: %d", val)
+		}
+		//nolint:gosec // int32 to uint64 conversion is safe for non-negative values
 		return uint64(val), nil
 	case int16:
-		//nolint:gosec // int16 to uint64 conversion is safe here
+		if val < 0 {
+			return 0, fmt.Errorf("negative hash value: %d", val)
+		}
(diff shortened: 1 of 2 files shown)


yankay:
HI @Alex-ai-future 
Thanks for the fix! The approach looks good overall.

Would you mind taking a look at Copilot's comments, especially rejecting negative signed hash values before converting to uint64 and adding a regression test for that case?

Alex-ai-future:
Thanks for the review! I've addressed the feedback:

**Negative value checks** — Added explicit negative checks for all signed integer cases (`int8`, `int16`, `int32`, `int64`, `int`). Negative values now return an error instead of silently wrapping to large uint64 values. Added a regression test `negative_signed_integers` covering all signed types.

**Code duplication** — The type switch pattern is consistent with the existing `toInt()` helper in `vllm_adapter.go`, which also uses explicit per-type cases. Go's type switch requires enumerating each type, and extracting a helper would add complexity without reducing the number of cases.

yankay:
Thanks, this looks good to me. 

/lgtm

PR #674, opened 2026-06-17
fix(examples): wire online kv-events ZMQ subscriber

Iceber:
## Summary

This PR fixes the online KV events example used by the Helm [kv-cache-manager deployment](https://github.com/llm-d/llm-d-kv-cache/blob/main/vllm-setup-helm/templates/kv-cache-manager.yaml#L73)

Previously, the online manager started the KV events pool but never created a ZMQ subscriber, so vLLM KV cache events published to `ZMQ_ENDPOINT` were not ingested. The manager could still serve HTTP scoring endpoints, but its KV block index would not be updated from runtime events.

The fix wires `SubscriberManager` into the online startup path:
- reads the KV events config once in `run`
- starts the events pool
- creates a static ZMQ subscriber for `ZMQ_ENDPOINT` / `ZMQ_TOPIC`
- shuts down the subscriber manager and pool during graceful shutdown

### DIFF
--- a/examples/kv_events/online/main.go
+++ b/examples/kv_events/online/main.go
@@ -45,6 +45,8 @@ const (
 	envZMQEndpoint = "ZMQ_ENDPOINT"
 	envZMQTopic    = "ZMQ_TOPIC"
 
+	staticSubscriberID = "static-zmq-subscriber"
+
 	envPoolConcurrency = "POOL_CONCURRENCY"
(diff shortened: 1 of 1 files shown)


yankay:
Thanks for fixing this. The subscriber wiring matches the Helm kv-cache-manager path, and CI is green.
/lgtm

Issue #644, opened 2026-06-07
[Observability] Extend OTel Tracing Coverage (kvevents + tokenization)


gyliu513:
Sub task of #617 

### Background
Index Add, Evict, and Lookup are already traced via the tracedIndex wrapper (pkg/kvcache/kvblock/traced_index.go, completed in #637), and score_tokens / scorer.compute spans exist. The remaining gaps are the ZMQ event-processing path and tokenization span propagation.

### Scope
Add spans for the ZMQ event-processing path in pkg/kvevents (message receive → decode → pool dispatch) and strengthen context propagation so tokenization spans link to their parent request span.

Instrument at the interface boundary following the existing tracedIndex wrapper pattern, do not inline spans into business logic. Reuse telemetry.InstrumentationName and the llm_d.kv_cache.* span-name prefix.

gyliu513:
There will be two sub task for this issue:

- kvevents span https://github.com/llm-d/llm-d-kv-cache/pull/678
- tokenization @satyamg1620  if you are busy, I can also fix this, otherwise, I will leave it to u, thanks!

satyamg1620:
@gyliu513 I have the changes ready in local for both. I will make changes for tokenization part only and will raise the PR, will itbe ok?

gyliu513:
@satyamg1620 sounds good, thanks!

gyliu513:
this will be fixed in llm-d-router

PR #677, opened 2026-06-21
feat(telemetry): correlate kv-cache logs with traces via span context

### LINKED ISSUE #665: [Observability] Logs are not correlated with traces: trace_id/span_id never injected into the logger, aligned with llm-d-router#1630
Sub task of #617

## Description

llm-d-kv-cache emits OpenTelemetry spans and structured logr (zap) logs, but the two are never correlated: no code path injects `trace_id`/`span_id` into the logger after a span is started, and there is no automatic correlation hook anywhere. When debugging a scoring request, there is no way to jump from a log line to the corresponding trace, or to find the log lines belonging to a given trace.

This aligns kv-cache with llm-d/llm-d-router#1630, which tracks the same gap in the EPP and the disaggregation sidecar (and llm-d/llm-d-inference-payload-processor#161 for IPP). The kv-cache scorer runs *inside* the EPP process and its spans (`llm_d.kv_cache.*`) appear as children of the EPP's `gateway.request` span in the same trace, so log–trace correlation should be consistent across the whole chain.

## Current behavior

The scoring path starts its own spans but never refreshes the logger stored in the context:

- In `Indexer.ScoreTokens` (`pkg/kvcache/indexer.go`), `tracer.Start` creates the `llm_d.kv_cache.score_tokens` span and the very next statement retrieves the logger via `log.FromContext(ctx)`. The span context is already in `ctx` at that point, but the logger is never enriched with `trace_id`/`span_id` and never stored back via `log.IntoContext`, so log lines such as `"found tokens"` (with the block keys for the request) carry no correlation fields.
- The same pattern applies to the other span sites: `llm_d.kv_cache.scorer.compute` in `pkg/kvcache/traced_scorer.go` and `llm_d.kv_cache.index.{lookup,add,evict}` in `pkg/kvcache/kvblock/traced_index.go` — the wrapped index/scorer implementations keep logging through the un-enriched context logger.
- Because kv-cache is a library, the logger comes from the host's context. Even once llm-d/llm-d-router#1630 lands and the EPP injects `trace_id`/`span_id` at the `gateway.request` root, kv-cache logs would carry the *parent EPP span's* `span_id`, not the kv-cache span that actually produced the log line — and standalone consumers (`examples/kv_cache_index_service`, offline/online kv-events examples) would still have no correlation at all.
- The kv-events ingestion path (`pkg/kvevents/pool.go`) currently has neither spans (#644 / #639) nor correlation fields; whatever mechanism is chosen here should be reused when those spans are added.
- `pkg/utils/logging/logger.go` only provides plain zap test loggers with no OTel hook, so there is no fallback either.

## Proposed change

Add a small helper in `pkg/telemetry` and call it right after each `tracer.Start`, e.g.:

```go
sc := span.SpanContext()
if sc.IsValid() {
    logger := log.FromContext(ctx).WithValues(
        "trace_id", sc.TraceID().String(),
        "span_id", sc.SpanID().String(),
    )
    ctx = log.IntoContext(ctx, logger)
}
```

Call sites: `pkg/kvcache/indexer.go` (`ScoreTokens`), `pkg/kvcache/traced_scorer.go`, `pkg/kvcache/kvblock/traced_index.go` — all downstream code already retrieves the logger from the context, so enriching at the span boundaries covers the index/scorer implementations without further changes. The kvevents spans from #644/#639 should use the same helper when they land.

Whatever field names and mechanism llm-d/llm-d-router#1630 settles on should be adopted here as-is, so that logs from EPP, IPP, the sidecar, and kv-cache can be queried uniformly by `trace_id`.

## Why is this needed

- Without correlation, debugging a slow lookup or a wrong cache-hit score means manually matching timestamps between Jaeger and the log backend; the most useful diagnostic lines (block keys, pod scores, eviction decisions) are logged at DEBUG/TRACE and are exactly the ones you want to scope to a single trace.
- Trace-correlated logging is being standardized across llm-d components (llm-d/llm-d-router#1630, llm-d/llm-d-inference-payload-processor#161); without this change the kv-cache spans inside the same EPP trace would be the only hop whose logs cannot be found by `trace_id`.
- The injection point is cheap: three call sites that already start spans, with no new dependencies and no change to existing metrics or span structure (#651/#652).

satyamg1620:
## What
Adds log–trace correlation to kv-cache. Introduces `telemetry.LoggerWithSpanContext(ctx, span)`, which injects the active span's `trace_id`/`span_id` into the context logger, and calls it at the three span boundaries:
- `Indexer.ScoreTokens` (`pkg/kvcache/indexer.go`)
- `tracedScorer.Score` (`pkg/kvcache/traced_scorer.go`)
- `tracedIndex.{Add,Evict,Lookup}` (`pkg/kvcache/kvblock/traced_index.go`)

Downstream code already reads the logger from context, so enriching at the span boundaries covers the index/scorer implementations with no further changes. The helper is a no-op when the span context is invalid (tracing off or span unsampled).

## Why
Without this, debugging a lookup or cache-hit score means manually matching timestamps between the trace backend and logs. This aligns kv-cache with the cross-component standardization in llm-d-router#1630.

Fixes #665

## Notes 
- Field names are `trace_id`/`span_id` per this issue's snippet. If llm-d-router#1630 settles on different keys, this should follow it.
- kvevents spans (#644/#639) don't exist yet; they should reuse this helper when they land.

### DIFF
--- a/pkg/telemetry/logging_test.go
+++ b/pkg/telemetry/logging_test.go
@@ -1,0 +1,100 @@
+/*
+Copyright 2025 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package telemetry_test
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/go-logr/logr"
+	"github.com/go-logr/logr/funcr"
+	"go.opentelemetry.io/otel"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/trace"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/llm-d/llm-d-kv-cache/pkg/telemetry"
+)
+
+// captureLogger returns a logr.Logger that appends every emitted record's
+// formatted key/values to *out, so tests can assert on what was logged.
+func captureLogger(out *[]string) logr.Logger {
+	return funcr.New(func(_, args string) {
+		*out = append(*out, args)
+	}, funcr.Options{})
+}
+
+// startSampledSpan starts a real, always-sampled span so its span context is
+// valid (a no-op span would report IsValid() == false).
+func startSampledSpan(t *testing.T, ctx context.Context) (context.Context, trace.Span) {
+	t.Helper()
+	tp := sdktrace.NewTracerProvider(sdktrace.WithSampler(sdktrace.AlwaysSample()))
+	otel.SetTracerProvider(tp)
+	return telemetry.Tracer().Start(ctx, "test")
+}
+
+func TestLoggerWithSpanContextInjectsFields(t *testing.T) {
+	var logged []string
+	ctx := log.IntoContext(context.Background(), captureLogger(&logged))
+
+	ctx, span := startSampledSpan(t, ctx)
+	defer span.End()
+
+	sc := span.SpanContext()
+	if !sc.IsValid() {
+		t.Fatal("expected a valid span context for the sampled span")
+	}
+
+	ctx = telemetry.LoggerWithSpanContext(ctx, span)
+	log.FromContext(ctx).Info("hello")
+
+	if len(logged) != 1 {
+		t.Fatalf("expected 1 log record, got %d", len(logged))
+	}
+	rec := logged[0]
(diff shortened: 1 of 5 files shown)


pkg/telemetry/logging_test.go
```diff
@@ -0,0 +1,100 @@
...
+		t.Errorf("log record missing span_id: %q", rec)
+	}
+}
+
+func TestLoggerWithSpanContextNoopForInvalidSpan(t *testing.T) {
+	var logged []string
+	ctx := log.IntoContext(context.Background(), captureLogger(&logged))
+
+	// A span from the default no-op tracer has an invalid span context.
+	otel.SetTracerProvider(trace.NewNoopTracerProvider())
```
gyliu513:
```suggestion
otel.SetTracerProvider(noop.NewTracerProvider())
```

satyamg1620:
@gyliu513 @liu-cong Can you please review this PR

gyliu513:
Thanks @satyamg1620 , looks good to me.

@liu-cong @vMaroon can you help review? Thanks

liu-cong:
Thanks, please open a PR in the router repo per https://github.com/llm-d/llm-d-kv-cache/issues/701

PR #683, opened 2026-06-23
refactor(tokenization): drop dead offsets return from Render

### LINKED ISSUE #462: Remove offsets from Tokenizer interface
`Render` and `RenderChat` in the `Tokenizer` interface return `([]uint32, []types.Offset, error)`, but since switching to the renderer service (#461) these methods always return `nil` offsets. The offset return value is now dead weight — all callers already discard it with `_`.

Should drop `[]types.Offset` from both signatures in the interface and all implementations (`UdsTokenizer`, `CachedTokenizer`, `CompositeTokenizer`, mocks).

mayur-tolexo:
## What

Removes the `[]types.Offset` return value from the `Tokenizer.Render(string)` interface method and its implementation.

## Why

Since the move to the renderer service (`RenderChatCompletion` RPC), `Render` has always returned a `nil` offset slice — the underlying `RenderCompletion` RPC doesn't produce character offsets. Every caller already throws it away with `_`, so the return value is dead weight that just makes the signature noisier and invites confusion about whether offsets are ever populated.

Closes #462.

## Changes

- Interface `Tokenizer.Render` is now `Render(string) ([]uint32, error)`.
- `UdsTokenizer.Render` drops the always-`nil` offset return; doc comment updated to note the renderer service doesn't produce offsets.
- Updated the single production caller (`pool.go`), the `MockTokenizer`, and the unit + e2e tests.

`Encode` is intentionally left alone — it still returns real offsets from the `Tokenize` RPC and the e2e tests assert on them.

## Testing

- `go build ./...` and `go vet ./...` clean
- `gofmt -l` clean
- `go test ./pkg/tokenization/...` passes

No behavior change — this is a pure signature cleanup.

### DIFF
--- a/pkg/tokenization/pool.go
+++ b/pkg/tokenization/pool.go
@@ -136,7 +136,7 @@ func (pool *Pool) processTask(task Task) error {
 	var features *MultiModalFeatures
 	var err error
 	if task.RenderReq == nil {
-		tokens, _, err = pool.tokenizer.Render(task.Prompt)
+		tokens, err = pool.tokenizer.Render(task.Prompt)
 		if err != nil {
 			log.Log.Error(err, "failed to render tokens", "prompt", task.Prompt)
 			return err
--- a/pkg/tokenization/pool_test.go
+++ b/pkg/tokenization/pool_test.go
@@ -47,23 +47,16 @@ func (m *MockTokenizer) RenderChat(renderReq *types.RenderChatRequest) ([]uint32
 	return tokens, features, args.Error(2)
 }
 
-func (m *MockTokenizer) Render(prompt string) ([]uint32, []types.Offset, error) {
+func (m *MockTokenizer) Render(prompt string) ([]uint32, error) {
 	args := m.Called(prompt)
 	if args.Get(0) == nil {
-		return nil, nil, args.Error(2)
+		return nil, args.Error(1)
 	}
 	tokens, ok := args.Get(0).([]uint32)
 	if !ok {
 		panic("MockTokenizer.Render: expected []uint32")
 	}
-	if args.Get(1) == nil {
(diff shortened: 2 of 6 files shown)


yankay:
HI @mayur-tolexo 

Thanks, this cleanup makes sense to me. Since this changes a public Go API signature, could we mention it in the release notes as a breaking API cleanup?

I checked llm-d-router: the default vLLM HTTP render path is not affected, but the deprecated UDS adapter still uses the old three-return-value UdsTokenizer.Render call and will need a small compile-time update when router bumps kv-cache:
https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/uds.go#L55

/ok-to-test

mayur-tolexo:
> HI @mayur-tolexo
> 
> Thanks, this cleanup makes sense to me. Since this changes a public Go API signature, could we mention it in the release notes as a breaking API cleanup?
> 
> I checked llm-d-router: the default vLLM HTTP render path is not affected, but the deprecated UDS adapter still uses the old three-return-value UdsTokenizer.Render call and will need a small compile-time update when router bumps kv-cache: https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/uds.go#L55
> 
> /ok-to-test

Thanks @yankay. Agreed it's a breaking API change - Render is exported, so importers need a compile-time update — though it's source-only, no behavior change. Could you flag it as a breaking cleanup in the release notes on your side?

sagearc (approved):
/lgtm

vMaroon:
lets hold with this, we're bringing them back:
- https://github.com/vllm-project/vllm/pull/44226

PR #685, opened 2026-06-26
fix(fs_backend): remove OffloadPromMetrics monkey-patch

### LINKED ISSUE #529: fs-connector: remove OffloadPromMetrics monkey-patch once we move to vLLM v0.20
### Background

PR #522 added a monkey-patch (`llmd_fs_backend/metrics.py`, `install_offload_metric_suffix_patch`) that works around `ValueError: Duplicated timeseries in CollectorRegistry` when `MultiConnector` wraps two `OffloadingConnector` children (CPU + FS). The patch is loaded as a `vllm.general_plugins` entry point and appends a `spec_name`-derived suffix (`_cpu`, `_sharedstorage`, ...) to offload metric names on collision.

We are currently pinned to `vllm==0.18.1`. vLLM v0.20 is expected to disambiguate the `OffloadPromMetrics` names natively, making this workaround unnecessary.

### Task

When `llmd_fs_backend` bumps its `vllm` pin to `>=0.20`:

- Delete `kv_connectors/llmd_fs_backend/llmd_fs_backend/metrics.py`.
- Remove the `install_offload_metric_suffix_patch()` call from `llmd_fs_backend/__init__.py`.
- Remove the `[project.entry-points."vllm.general_plugins"]` block from `pyproject.toml`.
- Rebuild the wheel; verify `MultiConnector(CPU+FS)` still starts cleanly and `/metrics` exposes distinct `vllm:kv_offload_*` series.

### Reference

- PR #522 — introduces the workaround
- TODO markers already present in `metrics.py`, `__init__.py`, and `pyproject.toml`

Alex-ai-future:
## Summary

vLLM PR [#40010](https://github.com/vllm-project/vllm/pull/40010) (merged as `4b7f5ea1a`) fixed the duplicate Prometheus timeseries issue in `MultiConnector.build_prom_metrics()` via a `seen_classes` dedup mechanism. Since llmd_fs_backend pins `vllm==0.22.0`, the monkey-patch is no longer needed.

## Changes

| File | Action |
|------|--------|
| `llmd_fs_backend/metrics.py` | **Deleted** — entire monkey-patch module (96 lines) |
| `llmd_fs_backend/__init__.py` | Removed `install_offload_metric_suffix_patch()` call and import |
| `pyproject.toml` | Removed `[project.entry-points."vllm.general_plugins"]` block |
| `tests/cpu/test_no_monkey_patch.py` | **Added** — verifies vLLM `seen_classes` dedup exists |

## Root Cause

Previously, `MultiConnector` with two `OffloadingConnector` children (CPU + FS) hit `ValueError: Duplicated timeseries` because both instances registered the same Prometheus metric names. The monkey-patch caught this error and retried with a `spec_name` suffix.

vLLM upstream now prevents duplicate registration at the source via `seen_classes` dedup in `MultiConnector.build_prom_metrics()` — same connector class only registers metrics once.

## Impact

- No functional change for single-connector deployments
- MultiConnector(CPU+FS) now relies on vLLM native dedup instead of try-catch retry
- Metrics use `transfer_type` label (`GPU_to_CPU` vs `GPU_to_FILE`) for per-connector visibility

Closes #529


kfirtoledo:
Hi, thanks @Alex-ai-future. Since the llmd-0.80 release is out, I prefer not to make any more changes in the llm-d fs connector unless it's a critical bug, so I think we can close this PR.

PR #692, opened 2026-07-06
fix(pvc_evictor): restart crawler processes on unexpected death

Alex-ai-future:
## Purpose

The main monitoring loop in `evictor.py` checked and restarted activator,
deleter, and folder_cleaner processes when they died unexpectedly, but did
not check crawler processes. If a crawler crashed (OOM, segfault), its
assigned hex range would permanently stop being scanned, causing cold files
in that range to accumulate and eventually fill the disk.

With the default 8 crawlers, a single crawler death means 12.5% of files
are no longer scanned. Multiple crawler deaths accelerate disk fill.

**Changes:**
- `kv_connectors/pvc_evictor/evictor.py`: Add crawler process alive check
  and restart logic to the `except Exception:` monitoring block, following
  the same pattern as the existing activator/deleter/folder_cleaner restart
  code. Logs crawler death with exit code and hex range for diagnostics.

## Test Plan

- [x] Run existing test suite: `cd kv_connectors/pvc_evictor && python3 -m pytest tests/ -v`
- [ ] Manual: kill a crawler process during runtime and verify it restarts via logs

## Test Result

```
41 passed in 0.12s
```

### DIFF
--- a/kv_connectors/pvc_evictor/evictor.py
+++ b/kv_connectors/pvc_evictor/evictor.py
@@ -346,6 +346,32 @@ class PVCEvictor:
                         )
                         folder_cleaner_process_obj.start()
 
+                    # Check and restart crawler processes
+                    for i, crawler_proc in enumerate(crawler_processes):
+                        if not crawler_proc.is_alive():
+                            hex_range = hex_ranges[i]
+                            self.logger.error(
+                                f"Crawler P{i + 1} died (exitcode={crawler_proc.exitcode}), "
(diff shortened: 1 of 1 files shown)


guygir (approved):
Looks great, thanks @Alex-ai-future .
I verified the restart path locally aswell, all good.

Alex-ai-future:
> Looks great, thanks @Alex-ai-future . I verified the restart path locally aswell, all good.

Thank you, can we merge and close this PR

PR #698, opened 2026-07-13
chore(pvc-evictor): standardize dev environment via pyproject.toml

### LINKED ISSUE #693: Standardize PVC Evictor development environment (pyproject.toml, Python version pin, venv support)
## Problem

PVC Evictor is the only Python subproject in this repo without `pyproject.toml`. The current dev environment is not reproducible:

| Dimension | Current State | Impact |
|-----------|--------------|--------|
| Package metadata | No `pyproject.toml` | Runtime deps (msgpack, pyzmq) only declared in Dockerfile |
| Python version | No pin | Local 3.14 vs CI 3.12 — behavior mismatch |
| venv | No Makefile support | `make test` uses system Python directly |
| Dev dependencies | `requirements-dev.txt` (only `pytest>=8.0`) | Runtime deps undeclared for local dev |

For comparison, `llmd_fs_backend` already has a complete `pyproject.toml` with `requires-python = ">=3.12"` and declared dependencies.

## Proposed Changes

1. **Add `pyproject.toml`** — project metadata, runtime deps (msgpack, pyzmq from Dockerfile), dev deps (pytest), `requires-python = ">=3.12"`
2. **Add `.python-version`** — pin to 3.12, matching CI's `actions/setup-python@v5`
3. **Update `Makefile`** — add `venv`/`dev` targets; `test` auto-detects venv (backward-compatible fallback to system Python)
4. **Remove `requirements-dev.txt`** — migrated to `pyproject.toml` `[project.optional-dependencies.dev]`
5. **Update CI** — `pip install -r requirements-dev.txt` to `pip install -e ".[dev]"`
6. **Fix `.gitignore`** — remove `*.python-version` pattern that would ignore the new `.python-version` file

## Scope

All changes are confined to `kv_connectors/pvc_evictor/` except:
- `.github/workflows/ci-pvc-evictor.yaml` (CI install command)
- `.gitignore` (remove `*.python-version` line — this pattern only matches `.python-version` files, which is exactly what we want to track)

## Verification

All 41 existing tests pass in three modes:
- `make test` with venv (`make dev` first)
- `make test` without venv (fallback to system Python)
- CI simulation (fresh venv + `pip install -e ".[dev]"` + pytest)

## Intent

I plan to implement this and submit a PR. Feedback on the approach (especially the `.gitignore` change and dependency placement) is welcome before I proceed.

gkneighb:
## Summary

Standardizes the PVC Evictor dev environment per #693, following the `llmd_fs_backend` pattern: `pyproject.toml` with a `python>=3.12` pin (matching CI and the runtime image), a `dev` extra replacing `requirements-dev.txt`, and a venv-backed `make test` mirroring the root Makefile's venv convention. CI installs the project editable with its dev extra. `.gitignore` already covers `kv_connectors/pvc_evictor/.venv`.

## One finding from the audit

The issue notes msgpack/pyzmq are "declared only in the Dockerfile" -- but nothing in the evictor imports them; the code is stdlib-only, so `dependencies` is deliberately empty rather than codifying phantom deps. The `pip install msgpack pyzmq` line in the Dockerfile appears to be dead weight -- flagging rather than removing it here in case it's anticipating planned work.

## Testing

- [x] `make test` from a clean checkout: creates the venv, installs `-e ".[dev]"`, 41/41 tests pass
- [x] CI workflow updated and exercises the same install path

Fixes #693


kv_connectors/pvc_evictor/pyproject.toml
```diff
@@ -0,0 +1,21 @@
...
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "pvc-evictor"
+version = "0.1.0"
+description = "PVC evictor for llm-d KV-cache offload storage"
+readme = "README.md"
+maintainers = [{ name = "llm-d community" }]
+requires-python = ">=3.12"
+# The evictor is stdlib-only; runtime third-party dependencies, if ever added,
```
guygir: Could we tweak this slightly? stdlib-only / if ever added is a little strong, the image still installs msgpack / pyzmq for the optional STORAGE_EVENTS_ENDPOINT path (llmd_fs_backend.event_publisher is copied into the image), so that Dockerfile install isn’t really unused
keeping dependencies = [] may still be fine since the package doesn’t import them directly - just want this comment to match that image-only optional path.

guygir (changes requested):
Thanks @gkneighb - looks great.

Two small requests before merge: one is inline, and could you also please update pvc_evictor/README.md so it no longer references the deleted requirements-dev.txt?

gkneighb:
Both done in c2fd89f -- and thanks for the catch on the msgpack/pyzmq comment. You're right, I misread that as unused; the Dockerfile's own comment (and `config.py`) make clear it's the optional `STORAGE_EVENTS_ENDPOINT` path via `event_publisher.py`. Comment now reflects that, `dependencies` stays `[]` since the package imports none directly. README no longer points at the removed `requirements-dev.txt`.

guygir (approved):
Thanks @gkneighb , looks great. /lgtm

PR #703, opened 2026-07-19
test(kvcache): add regression coverage for empty-BlockHashes MTP panic

### LINKED ISSUE #699: panic: index out of range in InMemoryIndex.Add when BlockStored event has tokens but empty block_hashes (MTP speculative decoding)
## What happened

  The kv-events worker panics and crashes the whole process when it receives a
  `BlockStored` event that carries `token_ids` but an empty `block_hashes` list.
  This reliably reproduces on a vLLM serving pod started with MTP speculative
  decoding:

  --speculative-config
  '{"method": "mtp", "num_speculative_tokens": 2}'

  Without `--speculative-config`, the crash does not occur.

  ## Stack trace

  panic: runtime error: index out of range [0] with length 0

  github.com/llm-d/llm-d-kv-cache/pkg/kvcache/kvblock.(*InMemoryIndex).Add
      pkg/kvcache/kvblock/in_memory.go:173
  github.com/llm-d/llm-d-kv-cache/pkg/kvcache/kvblock.(*instrumentedIndex).Add
      pkg/kvcache/kvblock/instrumented_index.go:36
  github.com/llm-d/llm-d-kv-cache/pkg/kvcache/kvblock.(*tracedIndex).Add
      pkg/kvcache/kvblock/traced_index.go:52
  github.com/llm-d/llm-d-kv-cache/pkg/kvevents.(*Pool).processEventBatch
      pkg/kvevents/pool.go:421
  github.com/llm-d/llm-d-kv-cache/pkg/kvevents.(*Pool).processRawMessage
      pkg/kvevents/pool.go:213
  github.com/llm-d/llm-d-kv-cache/pkg/kvevents.(*Pool).worker
      pkg/kvevents/pool.go:192

  ## Root cause

  In `processEventBatch` (`pkg/kvevents/pool.go:338`) engine keys are built as:

  ```go
  engineKeys := make([]kvblock.BlockHash, len(ev.BlockHashes))
```
  When ev.BlockHashes is empty this produces a non-nil, zero-length slice.
  ev.Tokens is non-empty, so TokensToKVBlockKeys returns a non-empty
  requestKeys, and index.Add(ctx, engineKeys /*len 0*/, requestKeys, ...)
  is called (pool.go:421).

  In InMemoryIndex.Add (pkg/kvcache/kvblock/in_memory.go:169) the guard is:

```go
  if engineKeys != nil {
      n := max(len(engineKeys), len(requestKeys)) // n = len(requestKeys) > 0
      for i := 0; i < n; i++ {
          ek := engineKeys[i*len(engineKeys)/n] // engineKeys[0] on a len-0 slice -> panic
          ...
      }
  }
```
  The guard checks != nil instead of length, so a non-nil empty slice enters
  the loop and indexes engineKeys[0] on a zero-length slice.

  ## Impact

  The panic runs in a kv-events worker goroutine started by Pool.Start
  (pool.go:141), so a single malformed event crashes the entire consuming
  process (e.g. the llm-d-router EPP). It is not a recoverable per-event error.

 ## Suggested fix

  Two independent guards, either of which prevents the panic:

  1. in_memory.go:169: change if engineKeys != nil to if len(engineKeys) > 0.
  The existing docstring already states that when engine keys are absent only
  requestKey -> PodEntry mappings should be created; a zero-length slice
  should be treated the same as nil.
  2. pool.go (~line 421): when len(engineKeys) == 0, either skip the event or
  pass nil explicitly so the speculative-entry path (request-key-only) is taken.

  Environment

  - llm-d-kv-cache: v0.9.0
  - Engine: vLLM with --speculative-config '{"method": "mtp", "num_speculative_tokens": 2}'
  - Consumer: llm-d-router EPP

elinacse:
The InMemoryIndex.Add index-out-of-range panic on empty (non-nil) engineKeys with non-empty requestKeys was already fixed by #670 via the shared engineToRequestMapping() length guard, but no test exercised Add() or Pool.processEventBatch() with the exact non-nil-empty-slice shape produced for BlockStored events without block hashes (e.g. vLLM MTP speculative decoding).

Add TestAddWithEmptyEngineKeys and TestBlockStoredEvent_MTPEmptyBlockHashes to lock in that behavior.

Co-Authored-By: Claude

Fixes #699

### DIFF
--- a/pkg/kvcache/kvblock/in_memory_test.go
+++ b/pkg/kvcache/kvblock/in_memory_test.go
@@ -222,3 +222,30 @@ func TestAddWithNilEngineKeys(t *testing.T) {
 	_, err = index.GetRequestKey(ctx, requestKey)
 	assert.Error(t, err, "GetRequestKey should fail since no engineKey mapping was created")
 }
+
+// TestAddWithEmptyEngineKeys tests that Add() with a non-nil but zero-length
+// engineKeys slice behaves the same as nil engineKeys (regression test: MTP
+// speculative decoding events carry tokens but no block hashes, and callers
+// build engineKeys via make([]BlockHash, len(ev.BlockHashes)), which yields a
+// non-nil empty slice rather than nil).
+func TestAddWithEmptyEngineKeys(t *testing.T) {
+	ctx := logging.NewTestLoggerIntoContext(t.Context())
+	index := createInMemoryIndexForTesting(t)
+
+	requestKey := BlockHash(66666666)
+	pod := PodEntry{PodIdentifier: "10.0.0.4:8080", Speculative: true}
+
+	// Add with a non-nil, empty engineKeys slice - must not panic.
+	err := index.Add(ctx, []BlockHash{}, []BlockHash{requestKey}, []PodEntry{pod})
+	require.NoError(t, err)
+
+	// Lookup by requestKey should work
+	podsPerKey, err := index.Lookup(ctx, []BlockHash{requestKey}, nil)
+	require.NoError(t, err)
+	assert.Len(t, podsPerKey[requestKey], 1)
+	assert.Contains(t, podsPerKey[requestKey], pod)
+
+	// GetRequestKey should NOT find a mapping (no engineKey was stored)
+	_, err = index.GetRequestKey(ctx, requestKey)
+	assert.Error(t, err, "GetRequestKey should fail since no engineKey mapping was created")
+}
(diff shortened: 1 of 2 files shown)


elinacse:
@vMaroon Can you review the PR . Thanks !

elinacse:
@yankay @dannyharnik Can anyone of you review this PR and post your comments ? Thanks !

vMaroon:
Hi, the codebase migrated to llm-d-router, pls reopen there

